PR #58, opened 2023-09-06
Use DataFrames to write to csv

### LINKED ISSUE #14: Use CSV and DataFrames to write the output
Currently, it is being done manually. Instead, we should try to create a DataFrame and write to csv.

abelsiqueira:
# Pull request details

Use the CSV and DataFrames packages to write the output file instead of manually writing each line.

Closes #14 

## Collaboration confirmation

As a contributor I confirm

- [x] Tests are passing
- [x] Lint is passing

### DIFF
--- a/src/io.jl
+++ b/src/io.jl
@@ -95,16 +95,20 @@ Saves the solution variable v_investment to a file "investments.csv" inside `out
 The format of each row is `a,v,p*v`, where `a` is the asset indexing `v_investment`, `v`
 is corresponding `v_investment` value, and `p` is the corresponding `p_unit_capacity` value.
 """
-function save_solution_to_file(output_folder, v_investment, p_unit_capacity)
+function save_solution_to_file(
+    output_folder,
+    s_assets_investment,
+    v_investment,
+    p_unit_capacity,
+)
     # Writing the investment results to a CSV file
-    output_file = open(joinpath(output_folder, "investments.csv"), "w")
-    write(output_file, "a,InstalUnits,InstalCap_MW\n")
-    for a in axes(v_investment, 1)
-        v = v_investment[a]
-        p = p_unit_capacity[a]
-        write(output_file, "$a,$v,$(p * v)\n")
-    end
-    close(output_file)
+    output_file = joinpath(output_folder, "investments.csv")
+    output_table = DataFrame(;
+        a = s_assets_investment,
+        InstalUnits = [v_investment[a] for a in s_assets_investment],
+        InstalCap_MW = [p_unit_capacity[a] * v_investment[a] for a in s_assets_investment],
+    )
+    CSV.write(output_file, output_table)
 
     return
 end
--- a/benchmark/benchmarks.jl
+++ b/benchmark/benchmarks.jl
@@ -23,6 +23,7 @@ solution = optimise_investments(parameters, sets)
 SUITE["io"]["output"] = @benchmarkable begin
     save_solution_to_file(
         $OUTPUT_FOLDER,
+        $(sets.s_assets_investment),
         $(solution.v_investment),
         $(parameters.p_unit_capacity),
     )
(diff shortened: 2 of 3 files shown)


abelsiqueira:
The benchmark is not being posted due to #59

src/io.jl
```diff
@@ -95,16 +95,20 @@ Saves the solution variable v_investment to a file "investments.csv" inside `out
...
-    output_file = open(joinpath(output_folder, "investments.csv"), "w")
-    write(output_file, "a,InstalUnits,InstalCap_MW\n")
-    for a in axes(v_investment, 1)
-        v = v_investment[a]
-        p = p_unit_capacity[a]
-        write(output_file, "$a,$v,$(p * v)\n")
-    end
-    close(output_file)
+    output_file = joinpath(output_folder, "investments.csv")
+    output_table = DataFrame(;
```
clizbe, line 106: What does the ";" in this line mean?

src/io.jl
```diff
@@ -95,16 +95,20 @@ Saves the solution variable v_investment to a file "investments.csv" inside `out
...
+        a = s_assets_investment,
+        InstalUnits = [v_investment[a] for a in s_assets_investment],
```
clizbe, line 108: If "a" is already defined as "s_assets_investment," and you're indexing "v_investment" by "a," isn't it redundant to specify "a in s_assets_investment?"

src/io.jl
```diff
@@ -95,16 +95,20 @@ Saves the solution variable v_investment to a file "investments.csv" inside `out
...
+        InstalCap_MW = [p_unit_capacity[a] * v_investment[a] for a in s_assets_investment],
```
clizbe, line 109: Is the comma at the end of this line correct?

clizbe (approved):
Thanks for the PR! I used the review as an opportunity to learn more about Julia. ;)

src/io.jl
abelsiqueira: `;` separates the normal arguments from the keywords arguments (arguments for which we must specify the full name). It is not necessary, but apparently, the formatter automatically added it.

src/io.jl
abelsiqueira: These two lines are independent. The constructor of `DataFrame` allows us to define the column names by passing keyword arguments. The first line says "create a column named 'a' with the content of `s_assets_investment`". The second creates a column named 'InstalUnits' with the values of `v_investment` for the assets that are in `s_assets_investment`. It uses something called vector comprehension.
So "a" in the first is the column name, and "a" in the second is just an index for the vector comprehension, and it isn't used anywhere internally after that.

src/io.jl
abelsiqueira: Yes, although it is optional and part of the formatting style: https://domluna.github.io/JuliaFormatter.jl/stable/#trailing_comma
The language understands that this last comma is part of style, and doesn't complain about it.

Trailing commas in multiline expressions (arrays, function arguments) are used with a bit of git in mind.
If you add an additional argument to that list, git will say you made two changes: 1. Add a comma at line N. 2. Add the new line.

Example: Let's say you have an array `v = [1, 2]` and changes it to `v = [1, 2, 3]`. Also assume that the actual array has long enough arguments that warrant breaking it into multiple lines.
If you don't have a comma:

```diff
 v = [
   1,
-  2
+  2,
+  3
]
```

If you have a comma:

```diff
 v = [
   1,
   2,
+  3,
]
```

The other argument for it is that if you reorder things, you don't have to fix that comma.

It is, of course, unnecessary and opinionated.

abelsiqueira:
Hi @clizbe, thanks for the review. Feel free to ask any questions, I :100:% agree with using the reviews as opportunity for learning.

abelsiqueira:
@suvayu see the fruits of your labour

src/io.jl
clizbe: This is super handy!

src/io.jl
clizbe: Thanks for explaining - now it makes much more sense how it's constructing a Dataframe.

src/io.jl
clizbe: Nice.

PR #76, opened 2023-09-15
Added run Julia Formatter on Save to README.dev

### LINKED ISSUE #29: Add "run on save" of Julia Formatter to README.dev
In VSCode (or similar) you can change your settings so that the Julia Formatter runs when you save. We should tell people to enable this in the README.dev.
This makes sure your code works in the Julia Formatter before running pre-commit, etc.

clizbe:
# Pull request details

In README.dev the instructions on how to run the JuliaFormatter when Saving (rather than manually) was added

Closes #29

## Collaboration confirmation

As a contributor I confirm

- [X] I read and followed the instructions in README.dev.md
- [X] The documentation is up to date with the changes introduced in this Pull Request (or NA)
- [X] Tests are passing
- [X] Lint is passing


datejada (commented):
Hi Lauren,

You haven't added the new files to your PR, and that is what was causing the error in your previous PR:

1. If you rename the file StaleWorkflow.yml to Stale.yml, then the old file is deleted (as it is in this PR), and a new file is created (the one that is missing in this PR)
2. The same happened to the .png file you were trying to add in the README.dev.md file. The test was not passing in the previous PR because the file was not in the commit. In this PR, the test passes because you don't have the reference to the file.

Please double-check the first point (the new file you need to add), and for the second point, if you choose to add the .png file, I recommend creating a new folder to store the file (to avoid having it in the root folder and use a more meaningful name).

.github/workflows/StaleWorkflow.yml
clizbe, line 1: @abelsiqueira  Help I can't figure out how to remove this part of the commit. :( I just want the part that adds the description to README.dev. This part deletes the StaleWorkflow, which I fixed in another commit.

suvayu: Are you still having an issue removing the faulty commit? You can revert any commit with something like this:
```
git revert <commit-hash>
```
In this case 249514e.

This will add a new commit with exactly the opposite changes as the mentioned commit, reverting the changes. Since we squash and merge PRs, the final version won't have the mistake.

Hope that helps

abelsiqueira: You probably have two diverging branches with some intersection between them. Like

```plaintext
o -- o -- o main
  \
    o -- o this_branch
```

(Note: assuming you already fetch and ff-only your main)

There are a few ways to try to fix, and can't tell which is easier right now.

Before: `git switch this_branch` and `git branch backup_this_branch`.

Approach 1: Try to rebase, see what happens, and squash online.

```plaintext
git rebase main this_branch
```

This should probably have some conflicts, which you will need to solve manually, and the `git add` the files. Using `git status` will help will instructions and seeing what files have issues.
At the end you will need to `git push myform this_branch --force`.

Approach 2: Squash first, then rebase

Check the squash info on the notes. After that, rebase using the idea above. There might still be conflicts.

Approach 3: Reset to the common root and restore unwanted removal

You can reset **two** commits with

```bash
git reset HEAD~2
```

Then it will be like you never committed those things, but your modification will be there.
From here you have to undo your changes.

Try `git checkout -- .github/workflows/*.yml` (or use the specific names). Check status before and after.

Try `git merge --ff-only main` to update your local branch to the main point.

Now you can commit only the change you want.

Approach 4: Copy your changes into a notepad, hard reset everything

Copy your changes to notepad.
Then

```
git reset --hard main
```

This will delete everything from the branch.

Copy back from notepad as if nothing ever happened.

---

Sorry for the info dump.

suvayu: @abelsiqueira is `git revert` not preferred for some reason?

abelsiqueira: No, just forgot about it. I don't normally use it.

suvayu: Since we squash merge when merging PRs, I think using revert is a much simpler workflow; no history rewriting required.

abelsiqueira (approved):
Looks great, thanks. :beach_umbrella:

Issue #59, opened 2023-09-06
Benchmark workflow does not work for PRs from forks


abelsiqueira:
The Benchmark workflow is supposed to run the benchmarks and comment the resources in the PR. However, it needs permissions that can't be granted if the PR is from a fork.

The "straightforward but not simple solution" would be to change our workflow entirely and use branches in the main repo instead of forks.

@suvayu also posted the idea below:
 
> IMO this should be a bot, and a cron job.  Bot to manually trigger it in a PR, and cron that checks performance regression, say, once a week.

_Originally posted by @suvayu in https://github.com/TNO-Tulipa/TulipaBulb.jl/issues/50#issuecomment-1706483823_

@suvayu, do you know how to set that up, do you know if it works without permissions?

suvayu:
The cron part is easy.

To trigger from comments, I have a possible implementation in mind using [`workflow_dispatch`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch), but it's a bit fidgety.

In the past, I've seen large open source projects use custom commands or tagging bots to trigger actions (like issue assignment and benchmark runs). I can figure it out.

We can start with cron. and add the manual trigger as we figure it out.

abelsiqueira:
I prefer to focus on the PR part. The regression after things are merged is not so useful unless you're comparing between releases, IMO.
I've seen some projects with the triggering comment implemented. Actually, I think one of the JSO ones. I just haven't been paying enough attention to remember what they did.
I think you can run on "comment on PR" instead of `workflow_dispatch`.

abelsiqueira:
The actions run on the main repo, so the permissions of the main apply. However, since we are admins of the main repo, when we issue `/benchmark`, it should use the right permissions. We still need to allow `write` in the `permissions` of `pull_request`, I believe (https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs).

suvayu:
Just to let you know, I solved it! PR coming soon!

PR #116, opened 2023-09-21
Benchmark workflow runnable as workflow_dispatch

suvayu:

The workflow_dispatch event accepts a fork and branch parameter,
and runs the benchmark workflow on it.  The results are shown in the
step summary

- #59

## Collaboration confirmation

As a contributor I confirm

- [x] Tests are passing
- [x] Lint is passing

### DIFF
--- a/.github/workflows/Benchmark.yml
+++ b/.github/workflows/Benchmark.yml
@@ -4,6 +4,10 @@ on:
         types: [created, edited]
     workflow_dispatch:
         inputs:
+            repo:
+                description: Fork to benchmark "user/repo", empty denotes current repo
+                type: string
+                default: ""
             ref:
                 description: Git branch/tag/reference to benchmark
                 type: string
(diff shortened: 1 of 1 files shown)


.github/workflows/Benchmark.yml
```diff
@@ -4,6 +4,10 @@ on:
         types: [created, edited]
     workflow_dispatch:
         inputs:
+            repo:
+                description: Fork to benchmark "user/repo", empty denotes current repo
+                type: string
+                default: ""
```
abelsiqueira, line 10: How does the empty string implies default to this repo?

.github/workflows/Benchmark.yml
```diff
@@ -19,23 +23,29 @@ concurrency:
...
+                            repo: context.repo.repo,
+                            pull_number: context.issue.number
+                          })
+                          core.setOutput('repo', pr.data.head.repo.full_name)
+                          core.setOutput('ref', pr.data.head.ref)
+                      }
+                      if (context.eventName == 'workflow_dispatch') {
+                          core.setOutput('repo', '${{ inputs.repo }}')
+                          core.setOutput('ref', '${{ inputs.ref }}')
+                      }
```
abelsiqueira, line 48: The benchmark runs against `TulipaEnergy/TulipaEnergyModel.jl#main`, so we might want to check that situation and raise an error instead of comparing it to itself. What do you think?

abelsiqueira (changes requested):
Thanks, seems right, although I don't understand exactly what happens so I left a comment about that. Also left a comment about a corner case asking your opinion.

.github/workflows/Benchmark.yml
suvayu: The checkout action (line 51) has `with.repository`.  When this is empty, it defaults to the current repo.

.github/workflows/Benchmark.yml
suvayu: Hmm, good point. I'll fix that

suvayu: I added an if block that checks for this, and fails explicitly

abelsiqueira (approved):
Thanks for the update and explanation. LGTM

PR #128, opened 2023-09-29
127-change-inputs-names

### LINKED ISSUE #127: Change InputTables.jl to match io.jl
### Summary

These changes need to be made so the files match. 
The types and column names should match.

### Basic Example

image
Also booleans

### Drawbacks

We may have future debates about naming.

### Reference Issues

_No response_

datejada:
# Pull request details

Changes in the Input files, InputTable definitions, and io code to make them match all together.

Closes #127

## Collaboration confirmation

As a contributor I confirm

- [NA] The documentation is up to date with the changes introduced in this Pull Request (or NA)
- [x] Tests are passing
- [x] Lint is passing


src/io.jl
```diff
@@ -12,31 +12,29 @@ function create_parameters_and_sets_from_file(input_folder::AbstractString)
...
+    A = s_assets = nodes_data_df[nodes_data_df.active.==true, :].name         #assets in the energy system that are active
+    Ap = s_assets_producer = nodes_data_df[nodes_data_df.type.=="producer", :].name  #producer assets in the energy system
+    Ac = s_assets_consumer = nodes_data_df[nodes_data_df.type.=="consumer", :].name  #consumer assets in the energy system
+    s_assets_investment = nodes_data_df[nodes_data_df.investable.==true, :].name #assets with investment method in the energy system
+    s_representative_periods = unique(nodes_profiles_df.rep_period_id)  #representative periods
+    s_time_steps = unique(nodes_profiles_df.time_step)   #time steps in the RP (e.g., hours)
 
     # Parameters for system
-    p_rp_weight = Dict((row.rp) => row.weight for row in eachrow(weights_df)) #representative period weight [h]
+    p_rp_weight = Dict((row.id) => row.weight for row in eachrow(rep_period_df)) #representative period weight [h]
```
clizbe, line 33: It's interesting to me that this shows as row.id instead of rep_period_df.id.
Does it make a difference? To me "id" is here a bit vague because it's only preceded by "row" and not "rep_period" and you have to read the whole line to see what it means.
Just a philosophical question. :P

src/io.jl
```diff
@@ -12,31 +12,29 @@ function create_parameters_and_sets_from_file(input_folder::AbstractString)
...
 
     # Parameters for assets
     p_profile = Dict(
-        (A[row.node_id], row.rp, row.k) => row.profile_value for
+        (A[row.id], row.rep_period_id, row.time_step) => row.value for
```
clizbe, line 37: Same goes for here. Not saying it's wrong or should be changed, just curious!

src/io.jl
```diff
@@ -46,19 +44,19 @@ function create_parameters_and_sets_from_file(input_folder::AbstractString)
     p_unit_capacity   = Dict{String,Float64}()
     p_init_capacity   = Dict{String,Float64}()
     for row in eachrow(nodes_data_df)
-        if row.node_name in Ap
-            p_variable_cost[row.node_name] = row.variable_cost
-            p_investment_cost[row.node_name] = row.investment_cost
-            p_unit_capacity[row.node_name] = row.asset_capacity
-            p_init_capacity[row.node_name] = row.initial_capacity
+        if row.name in Ap
```
clizbe, line 47: Is this the proper way of checking an item is in a set? Seems like the set should be by ID and not Name.

clizbe (approved):
Thanks for the PR! Looks good and consistent and quite readable.

src/io.jl
clizbe: @abelsiqueira
@datejada and I were wondering about the philosophy behind this. Is there a more efficient/proper way?

src/io.jl
clizbe: Nvm, this is addressed in another issue.

src/io.jl
abelsiqueira: This is pretty efficient - maybe the most efficient.
The dictionary is created with the correct elements. The alternative would be a more traditional for loop

```julia
p_rp_weight = Dict{Int,Float64}() # The types need to be specified because they can't be inferred from the element
for row in eachrow(rep_period_df)
    p_rp_weight[row.id] = row.weight
end
```

The comprehension is short and common, and the types should be inferred, so it looks nicer.
Since "row" is the variable name, it could be changed to anything, such as, "rep_period_row". I wouldn't, for personal preference, because I think row is enough in the context, and the context is short.

That said, the "(row.id)" should be changed to "row.id", since the parenthesis does nothing.

src/io.jl
abelsiqueira: Same answer.  I would change the for structure to the following:

```julia
dict = Dict(
    key => value
    for ...
)
```

To put the for in a separate line.

PR #132, opened 2023-09-30
Input validation

suvayu:
# Pull request details

- rename input tables schema file to snake_case
- validation function for input tables

- #122

## Collaboration confirmation

As a contributor I confirm

- [x] Tests are passing
- [x] Lint is passing


clizbe (approved):
Thanks for the PR! We're going to open some issues to change a couple things, but this looks good for now.

src/input_tables.jl
```diff
@@ -0,0 +1,96 @@
...
+            ),
+        ),
+    )
+
+    col_type_err = collect(
+        Iterators.filter(
+            ((_, expect, col_t),) -> if (supertype(expect) == supertype(col_t))
+                false
+            else
+                ((promote_type(expect, col_t) == Any) ? true : false)
```
suvayu, line 72: It would be good to get feedback on these two checks:
1. are the df column type and the expected type compatible types
2. if they are not, are the coercible without falling back to `Any`

abelsiqueira: I don't know if this is still relevant since I've been gone, but we should probably talk about these validations when you get back.

PR #144, opened 2023-10-17
Add functions to compute time resolution periods and matrices

### LINKED ISSUE #109: Implement a function that generates the resolution matrix
Create a function that receives (what?) and outputs the matrix.

We have to discuss  the details, which are not clear right now.

abelsiqueira:
# Pull request details

One assumption is made: the description of a time period is a range, i.e., an object `start:end`.
Creates a function `compute_rp_periods` that receives a list of time steps (such as the time steps of various flows), and computes the periods for that representative period.
Creates a function `resolution_matrix` that computes the matrix of a given flow/asset in a given representative period.

Closes #109 

## Collaboration confirmation

As a contributor I confirm

- [x] Tests are passing
- [x] Lint is passing

### DIFF
--- a/src/time-resolution.jl
+++ b/src/time-resolution.jl
@@ -40,19 +40,39 @@ function resolution_matrix(
     rp_periods::AbstractVector{<:UnitRange{<:Integer}},
     time_steps::AbstractVector{<:UnitRange{<:Integer}},
 )
-    M = [
+    matrix = [
         length(period ∩ time_step) / length(time_step) for period in rp_periods,
         time_step in time_steps
     ]
 
-    return M
+    return matrix
 end
 
 """
     rp_periods = compute_rp_periods(array_time_steps)
 
 Given the time steps of various flows/assets in the `array_time_steps` input, compute the representative period splits.
-Each element of `array_time_steps` is a an array of ranges.
+Each element of `array_time_steps` is an array of ranges with the following assumptions:
+
+  - An element is of the form `V = [r₁, r₂, …, rₘ]`, where each `rᵢ` is a range `a:b`.
+  - `r₁` starts at 1.
+  - `rᵢ₊₁` starts at the end of `rᵢ` plus 1.
+  - `rₘ` ends at some value `N`, that is the same for all elements of `array_time_steps`.
+
+Notice that this implies that they form a disjunct partition of `1:N`.
+
+The output will also be an array of ranges with the conditions above.
+The output is constructed greedly, i.e., it selects the next largest breakpoint following the algorithm below:
+
+ 0. Input: `Vᴵ₁, …, Vᴵₚ`, a list of time step ranges. Each element of `Vᴵⱼ` is a range `r = r.start:r.end`. Output: `V`.
+ 1. Compute the end of the representative period `N` (all `Vᴵⱼ` should have the same end)
+ 2. Start with an empty `V = []`
+ 3. Define the beginning of the range `s = 1`
+ 4. Define an array with all the next breakpoints `B` such that `Bⱼ` is the first `r.end` such that `r.end ≥ s` for each `r ∈ Vᴵⱼ`.
+ 5. The end of the range will be the `e = max Bⱼ`.
+ 6. Define `r = s:e` and add `r` to the end of `V`.
+ 7. If `e = N`, then END
+ 8. Otherwise, define `s = e + 1` and go to step 4.
 
 ## Examples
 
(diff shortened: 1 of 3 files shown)


abelsiqueira:
The coverage fails are from input_tables.jl, which I have not touched. We haven't discussed a coverage acceptance criteria, but either way, it is not because of this PR.

clizbe (commented):
Thanks looks great! I think Diego will have to check the second function.

src/time-resolution.jl
```diff
@@ -0,0 +1,86 @@
...
+ 1:4
+ 5:8
+ 9:12
+```
+"""
+function compute_rp_periods(
+    array_time_steps::AbstractVector{<:AbstractVector{<:UnitRange{<:Integer}}},
+)
+    rp_periods = UnitRange{Int}[]
+    period_start = 1
```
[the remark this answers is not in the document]
abelsiqueira: It should. I added a test for it.

src/time-resolution.jl
```diff
@@ -0,0 +1,86 @@
...
+
+# output
+
+3-element Vector{UnitRange{Int64}}:
+ 1:4
+ 5:8
+ 9:12
+```
+"""
+function compute_rp_periods(
```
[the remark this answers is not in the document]
abelsiqueira: I made major changes, can you check again? I think it is useful if I can make it more understandable.

src/time-resolution.jl
```diff
@@ -0,0 +1,117 @@
...
+
+  - An element is of the form `V = [r₁, r₂, …, rₘ]`, where each `rᵢ` is a range `a:b`.
+  - `r₁` starts at 1.
+  - `rᵢ₊₁` starts at the end of `rᵢ` plus 1.
+  - `rₘ` ends at some value `N`, that is the same for all elements of `array_time_steps`.
+
+Notice that this implies that they form a disjunct partition of `1:N`.
+
+The output will also be an array of ranges with the conditions above.
+The output is constructed greedly, i.e., it selects the next largest breakpoint following the algorithm below:
```
clizbe, line 65: "greedily"

src/time-resolution.jl
```diff
@@ -0,0 +1,117 @@
...
+time_steps2 = [1:3, 4:6, 7:9, 10:12]
+compute_rp_periods([time_steps1, time_steps2])
+
+# output
+
+3-element Vector{UnitRange{Int64}}:
+ 1:4
+ 5:8
+ 9:12
+```
```
clizbe, line 90: Is there a more complex example that would explain more?

src/time-resolution.jl
```diff
@@ -0,0 +1,117 @@
...
+
+    while period_start < representative_period_end
+        # The first range end larger than period_start for each range in each time_steps.
+        breakpoints = (
+            first(r[end] for r in time_steps if r[end] ≥ period_start) for
+            time_steps in array_time_steps
+        )
+        period_end = maximum(breakpoints)
+        @assert period_end ≥ period_start
+        push!(rp_periods, period_start:period_end)
```
clizbe, line 113: I just read what 'push!' does, but won't this link the values with period_start & period_end so they'll update as you change them?
So instead of [1:3, 4:6] you'd get [4:6, 4:6]?

clizbe (approved):
LGTM! Might still want Diego to check it over.

src/time-resolution.jl
abelsiqueira: No. That might happen if the `period_start` variable was an array-like (or pointer)-like object. In which case you would need to make a copy. But in this case that is not necessary.

For instance

```
L = []
v = ones(3)
push!(L, v) # Now L = [[1.0, 1.0, 1.0]]
v[1] = -1 # Now L = [[-1.0, 1.0, 1.0]]
push!(L, v) # Now L = [[-1.0, 1.0, 1.0], [-1.0, 1.0, 1.0]]
```

That happens because `v` is an array and pushing it into `L` pushes the "pointer" to it. Doing `push!(L, copy(v))` would fix it.

However, this would work:

```
L = []
v = [1, 1, 1]
push!(L, v)
v = [-1, 1, 1]
push!(L, v)
```

Because the second `v` is not the same "pointer" as the first. `L` has two elements, one is the pointer to the first `v`, and the second is the pointer to the second `v`.

In other words, pointers are complicated.

src/time-resolution.jl
abelsiqueira: I've added a more complex example, but I don't think it explains more. Let me know what would help.

clizbe (approved):
Looks good to me. Thanks for adding the complex example - I think it adds some explanation. :)

PR #148, opened 2023-10-17
Use id instead of name in the model

### LINKED ISSUE #125: Use ids in the model formulation instead of node_name
### Summary

The model currently uses the nodes' (assets) names as set. Instead, we should use the IDs of the nodes, since all data in the new CSV format is using the IDs.

Alternatively, we can make the IDs be node names and not the numbers.

### Basic Example

The edges are defined as `[(A[e.src], A[e.dst]) for e in edges(graph)]`, since we do a look up. Instead, we can possibly just use `edges(graph)` directly, since the `graph` is numerically indexed (and the IDs and the numerical index).

### Drawbacks

It might be harder to identify issues with specific variables and constraints. For instance, if a constraint is infeasible, we can't easily verify which variables are involved.

### Reference Issues

Should be done after #124

### LINKED ISSUE #146: [BUG] - The benchmarks have not been updated
### Description

They are broken at least since the graph additions.

### Reproduction steps

```bash
Run the benchmarks.
```

### Logs

_No response_

### OS

_No response_

abelsiqueira:
# Pull request details

Use id instead of name. Because of this, we can more easily use the graph structure.
@datejada, I hope you don't mind. I saw that you were assigned, but I wanted to check what could be done. There are some changes in the formulation because of that.

It would be best to fix #146 before this, so we can compare.

Closes #125 

## Collaboration confirmation

As a contributor I confirm

- [ ] I read and followed the instructions in README.dev.md
- [ ] Tests are passing
- [ ] Lint is passing


clizbe:
Decided against this.

PR #166, opened 2023-10-23
Validate CSV files schema while reading

### LINKED ISSUE #165: Read CSV passing type information to validate while reading
### Summary

It is possible to enforce both the existence of the column name and the column type while reading the CSV.
The `types` argument can be passed, and one of the options is that it is a Dictionary of all types.
All keys passed to `types` must exist in the CSV.
Furthermore, by also using the `strict=true` argument, we can enforce that all elements of the column can be converted to the desired type.
This would remove the need for the current validation (other validation might be desirable).

### Basic Example

The code

```julia
df = CSV.read(joinpath(dir, "bad-assets-data.csv"), DataFrame, header=2, types=Dict(:investable=>Bool), strict=true)
```

Fails with

```
ERROR: thread = 1 error parsing Bool around row = 1, col = 5: "t,", error=INVALID: DELIMITED | INVALID_DELIMITER
```

Which is not as clear as the validation error, but contains the desired error: Column 5 has a `"t"` which is not a `Bool`.

We can extend to

```julia
col_types = Dict(zip(fieldnames(schema), fieldtypes(schema)))
CSV.read(joinpath(dir, "bad-assets-data.csv"), DataFrame, header=2, types=col_types:, strict=true)
```

The `fieldnames` and `fieldtypes` have the names and types of the stsruct. The `zip` makes sure that they are connect one by one (like a zipper), and the `Dict` converts to a dictionary.

### Drawbacks

I don't see any.

### Reference Issues

_No response_

abelsiqueira:
# Pull request details

This will read the input_tables structs to get the column names and types.
Then, it requires that these columns exist and that their types are the ones from the structs.
Essentially, it validates while reading.

Closes #165

## Collaboration confirmation

As a contributor I confirm

- [x] Tests are passing
- [x] Lint is passing


suvayu (changes requested):
Much nicer! I don't think we need the `validate_df` code any more, should remove it.

clizbe:
Ooh fancy! :D

abelsiqueira:
Thank you for the reviews. I've removed the validate_df function, and since we don't require the explicit function name, I've changed the input to be a single file path. Also, I've renamed the function.

suvayu (approved):
LGTM

PR #240, opened 2023-11-08
Rename time symbols

### LINKED ISSUE #208: Better name for time resolution "stuff"
### What and Why

There are representative periods, and they have time steps.
There are Assets and Flows, and they have time steps, but different time steps because of the time resolution.
There are constraints and they have time steps, but different time steps from the other two because it depends on the time resolution of the others.

In addition to possibly selecting a better naming and using it in the formulation, it will also be useful for the time-resolution file.

Example:

- RPs have time steps, where each step has a duration; Keep `num_time_steps`; Change `rp_time_scale` to `rp_time_step_duration`;
- The assets and flows are defined on the RP time steps by aggregating these into **composite intervals**. Each asset and flow has its own definition of how many intervals and how they are composed. Each `flow` is being indexed in `K[rp]`, and instead should be indexed in a new set `intervals[f, rp]`.
- The balancing constraints are defined per group of assets and flows. The intervals of these define how the time steps of the RP are aggregated. Therefore, there are **periods** for the constraints. Each constraint is defined for a set `periods[a, rp]`.

### Possible Drawbacks

_No response_

### Related Issues

- #110 will be affected

Alfa-R:
# Pull request details

Rename symbols according to #208

Closes #208

## Collaboration confirmation

As a contributor I confirm

- [x] Tests are passing
- [x] Lint is passing

### DIFF
--- a/src/input-tables.jl
+++ b/src/input-tables.jl
@@ -49,9 +49,9 @@ struct RepPeriodData
     resolution::Float64         # Duration of each time steps (hours)
 end
 
-struct TimeIntervalsData
+struct PartitionData
     id::Int
     rep_period_id::Int
     specification::Symbol
-    time_intervals::String
+    partition::String
 end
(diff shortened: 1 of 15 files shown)


datejada (approved):
Thanks for the changes!

PR #265, opened 2023-11-17
Fix summation of flow profile over time block

### LINKED ISSUE #264: [BUG] - Flow profiles are not being used correctly
https://github.com/TulipaEnergy/TulipaEnergyModel.jl/blob/27924f99e5a2a3d0bdd8f391e583f0ac186cb680/src/model.jl#L211

The line above with the `get` is trying to find a `Bflow` index, which doesn't exist. It should have been changed to a sum, like the assets profile, but since the `get` checks existence, the code does not fail, and instead defaults to 1.0

This will lead to a change in the test as well, so this takes priority.

abelsiqueira:
# Pull request details

Creates a function for the sum of the flow profile over the time block.
Also changes the assets_profiles_sum to be a function instead of an expression, because it does not have any variables or expression, and therefore it can be immediately computed.

Closes #264 

## Collaboration confirmation

As a contributor I confirm

- [x] Tests are passing
- [x] Lint is passing

### DIFF
--- a/src/model.jl
+++ b/src/model.jl
@@ -70,6 +70,17 @@ function create_model(graph, params, sets; verbose = false, write_lp_file = fals
         return length(B1 ∩ B2) * params.rp_resolution[rp]
     end
 
+    # Sums the profile of asset a, representative period rp over the time block B
+    # Uses the default_value when that profile does not exist.
+    function assets_profile_sum(a, rp, B, default_value)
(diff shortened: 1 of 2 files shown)


datejada (approved):
Thanks for the fix and I like that we now have a function that has the default_value as an argument (that we can use with zero or one depending on the case).

PR #277, opened 2023-11-21
Store the solution inside the graph structure and the objective value inside the problem struct

### LINKED ISSUE #268: Store solution in the graph structures
Store the solution (variable values) inside the structure as well.

abelsiqueira:
# Pull request details

Edit: The old text is below the separator.

This PR stores the solution to the problem inside the graph structure.
It also saves the objective value inside the energy problem structure.

---

@datejada, this PR became too long, but splitting will take some time, and explaining in the parts will be hard. So I will explain it here, while I think of how to split this, ok?

This PR makes the structure EnergyProblem hold many things necessary to make it the center of attention. It still keeps the EnergyModel creation and solving separated with different functions. The benefits are not clear, since it depends if anyone will want to use these at all without the EnergyModel. Let me know if it is better to remove it.
Originally, it was meant to be #268, which involves changing the assets and flows structures, but they are inside the EnergyProblem, so it is harder to do it nicely.
More details below:

1. The solution values were moved to the asset and flow structures.
2. The `constraints_partitions` variable was a bit of lost, so moving it to the energy_problem makes sense. However, it is one of the few things that make energy_problem unavoidable.
3. Renamed `create_energy_model...` to `create_energy_problem...`.
4. Moved the `model` structure to EnergyProblem. Also created helper variables inside it: `solved`, `objective_value`, and `termination_status`.
5. Create `create_model!(energy_problem)` that calls `create_model` with the internal values of the energy problem and creates the internal model of energy_problem.
6. Create `solve_model!(energy_problem)` that calls `solve_model` with the interval model of the energy problem and solves it, and then updates the solution values.
7. Return a `solution` NamedTuple from `solve_model` with all variable values.
8. Change the parameters of the `save_solution_to_file` function to be only the `graph`, since the solution is in there. Also creates a version with the energy problem as input.

Closes #268

## Collaboration confirmation

As a contributor I confirm

- [x] Tests are passing
- [x] Lint is passing

### DIFF
--- a/benchmark/benchmarks.jl
+++ b/benchmark/benchmarks.jl
@@ -25,8 +25,8 @@ SUITE["model"]["solve_model"] = @benchmarkable begin
     solve_model!($energy_problem)
 end
 
-solution = solve_model!(energy_problem)
+solve_model!(energy_problem)
 
 SUITE["io"]["output"] = @benchmarkable begin
-    save_solution_to_file($OUTPUT_FOLDER_BM, $(energy_problem.graph), $solution)
+    save_solution_to_file($OUTPUT_FOLDER_BM, $energy_problem)
 end
--- a/src/io.jl
+++ b/src/io.jl
@@ -145,7 +145,19 @@ function read_csv_with_schema(file_path, schema; csvargs...)
 end
 
 """
-    save_solution_to_file(output_folder, graph, solution)
+    save_solution_to_file(output_folder, energy_problem)
+
+Saves the solution from `energy_problem` in CSV files inside `output_file`.
+"""
+function save_solution_to_file(output_folder, energy_problem::EnergyProblem)
+    if !energy_problem.solved
+        error("The energy_problem has not been solved yet.")
+    end
+    save_solution_to_file(output_folder, energy_problem.graph)
+end
+
+"""
+    save_solution_to_file(output_file, graph)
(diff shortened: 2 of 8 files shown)


datejada (approved):
EnergyProblem is holding many parts now. Is there a way to automatically create a diagram with all the elements that compose the structure for the documentation? In any case, we should treat it as a new issue. So, I approve this PR.

datejada:
@abelsiqueira:

1. Regarding the PR, let's keep all the changes in this PR; I don't see the benefit of splitting the PR, and it will make things harder for you without any particular gain.
2. I don't foresee any use case of the creation and solving separated with different functions from the EnergyProblem. So, I prefer to be practical and go for what is more effortless now and keep it like that. But, if you go for including them in the EnergyProblem, I'm ok with that too.
3. regarding the comment below, it is what it is, and I think we gain more than we loss, so go for it 😄 

> Originally, it was meant to be https://github.com/TulipaEnergy/TulipaEnergyModel.jl/issues/268, which involves changing the assets and flows structures, but they are inside the EnergyProblem, so it is harder to do it nicely.

abelsiqueira:
1. I will at least create some issues to separate the work, and maybe do some splitting. I'll let you know soon.
2. The creation and solving are separate from before already

PR #297, opened 2023-11-24
Include the investment potential

### LINKED ISSUE #197: Add a parameter and constraint in the model to include a max investment limit
### What and Why

We need to define two parameters to limit the maximum investment: one for the assets and another for the flows.

- For the assets, it should be in the file `assets-data.csv`, and for the flows, it should be in the file `flows-profiles.csv`.
- The name of the parameter could be `max_investment_limit`, the units should be in `MW`, and the type is a `Float64`
- Then add new constraints/bounds to limit the investment variables for assets and flows
- The constraint/bound should not be included in the model if there is no data (empty/null) 

### Possible Drawbacks

_No response_

### Related Issues

_No response_

datejada:
# Pull request details

Including an upper bound to the investment variables in the model.

Closes #197 

## Collaboration confirmation

As a contributor I confirm

- [x] Tests are passing
- [x] Lint is passing


datejada:
@abelsiqueira I have one questions before continuing with the changes:

- It would be nice to have an option that if there is no data in the column of the Max investment, then you don't impose a limit to the investment variable. My problem is that if I leave a blank, I get some validation errors. I would appreciate any ideas on how to achieve this. This might be useful for other current parameters.

If you have any other comments on the current implementation, please don't hesitate to add them.

abelsiqueira:
I think you have to change the type of the column in the `input-tables.jl` to `Union{Missing,Float64}`.
It would be useful to know all default/allow missing values, so we can see how to deal with it in the larger scale.

datejada:
Hi @abelsiqueira, thanks for the tip on the missing type. The PR is ready for your review now.

src/model.jl
```diff
@@ -279,6 +279,28 @@ function create_model(
...
+    )
+
+    for (u, v) ∈ Fi
+        if flow_max_capacity[(u, v)] > 0 && !ismissing(graph[u, v].investment_limit)
+            set_upper_bound(
+                flows_investment[(u, v)],
+                graph[u, v].investment_limit / flow_max_capacity[(u, v)],
+            )
+        end
+    end
```
abelsiqueira, line 302: Since the import and export capacities are only numbers, using expressions is overkill.
I just them explicitly below (the linter might complain). If you want to keep the variables for future use, then maybe it makes sense to create a different structure. If you just want to write a shorter equation, we can create the variable inside the for.
```suggestion
    for (u, v) ∈ Fi
        if flow_max_capacity[(u, v)] > 0 && !ismissing(graph[u, v].investment_limit)
            set_upper_bound(
                flows_investment[(u, v)],
                graph[u, v].investment_limit / max(graph[u, v].export_capacity, graph[u, v].import_capacity),
            )
        end
    end
```

abelsiqueira (changes requested):
I made a suggestion. Let me know if you want to discuss it more.

abelsiqueira:
> This is something I've been wondering about - what's the efficient way to specify: Variable with unlimited capacity vs. Variable with 0 capacity?

I think using missing is the right way, i.e., write nothing on the column for that value in the CSV.
Writing Inf would be the alternative, but that is Julia-specific, so not recommended in this case.

Notice that to use missing, we need to change the column type in the `input-tables.jl` to be `Union{Missing, <old type>}`, and treat the missing values with `ismissing`. This PR does it correctly.

abelsiqueira (approved):
I approved, but I think it should changed to dictionary. If you prefer, we can merge this and create a new issue to verify if the change impacts performance.

PR #301, opened 2023-11-28
Fix objective function calculation of the variable cost

### LINKED ISSUE #300: The variable cost of flow does not get multiplied by the partition range
### Description

Hi all,

I have built the case NL & BE. When I implement the uniform partition 2 on all BE-related assets and flows, I get 2/3 of the objective value, compared with when I have both in uniform partition 1.

I think this is because, in the objective, the variable cost of the flow does not get multiplied by the timesteps it covers. So when we have 1/2 of the flow variables in one of the two countries, we get 2/3 of the objective value. 

Is there any way to fix this? I would like to have an objective that is constant with time partition (counting every time step regardless of the partition).

[input_NLBE_1h2h_27Nov2023.zip](https://github.com/TulipaEnergy/TulipaEnergyModel.jl/files/13480699/input_NLBE_1h2h_27Nov2023.zip)
[input_NLBE_27Nov2023.zip](https://github.com/TulipaEnergy/TulipaEnergyModel.jl/files/13480700/input_NLBE_27Nov2023.zip)

### Reproduction steps

```bash
1. Run both cases "input_NLBE" and "input_NLBE_1h2h".
2. Notice that the objective value differs significantly, which should not happen.
```

### Logs

```bash
flows_variable_cost = @expression(
        model,
        sum(
            representative_periods[rp].weight *
            graph[u, v].variable_cost *
            flow[(u, v), rp, B_flow] for (u, v) ∈ F, rp ∈ RP,
            B_flow ∈ graph[u, v].partitions[rp]
        )
    )

Can we multiply here with the length of B_flow?
```

### OS

_No response_

datejada:
# Pull request details

Add the duration of the flow to the calculation of the variable cost in the objective function.

Closes #300 

## Collaboration confirmation

As a contributor I confirm

- [x] Tests are passing
- [x] Lint is passing

### DIFF
--- a/src/model.jl
+++ b/src/model.jl
@@ -74,7 +74,7 @@ function create_model(
         sum(
             representative_periods[rp].weight *
             graph[u, v].variable_cost *
-            flow[(u, v), rp, B_flow] for (u, v) ∈ F, rp ∈ RP,
+            flow[(u, v), rp, B_flow] * length(B_flow) for (u, v) ∈ F, rp ∈ RP,
             B_flow ∈ graph[u, v].partitions[rp]
         )
     )
--- a/src/model.jl
+++ b/src/model.jl
@@ -74,8 +74,8 @@ function create_model(
         sum(
             representative_periods[rp].weight *
             graph[u, v].variable_cost *
-            flow[(u, v), rp, B_flow] * length(B_flow) for (u, v) ∈ F, rp ∈ RP,
-            B_flow ∈ graph[u, v].partitions[rp]
+            flow[(u, v), rp, B_flow] *
+            length(B_flow) for (u, v) ∈ F, rp ∈ RP, B_flow ∈ graph[u, v].partitions[rp]
         )
     )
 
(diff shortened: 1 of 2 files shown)


src/model.jl
```diff
@@ -74,8 +74,8 @@ function create_model(
         sum(
             representative_periods[rp].weight *
             graph[u, v].variable_cost *
-            flow[(u, v), rp, B_flow] for (u, v) ∈ F, rp ∈ RP,
-            B_flow ∈ graph[u, v].partitions[rp]
+            flow[(u, v), rp, B_flow] *
+            length(B_flow) for (u, v) ∈ F, rp ∈ RP, B_flow ∈ graph[u, v].partitions[rp]
```
abelsiqueira, line 78: You need the resolution too, right?

datejada: Yes, I agree. Thank you for pointing that out

abelsiqueira (approved):
Thanks

Issue #294, opened 2023-11-22
Profile the problem solving


abelsiqueira:
The simplified european model by @gzclarence takes 5 minutes to build and solve. We need to profile it.

Some references:
https://www.youtube.com/watch?v=BFvpwC8hEWQ&ab_channel=TheJuliaProgrammingLanguage
https://www.youtube.com/watch?v=Okn_HKihWn8&ab_channel=TheJuliaProgrammingLanguage
https://docs.julialang.org/en/v1/manual/profile/

abelsiqueira:
@gzclarence

gzclarence:
> The simplified european model by @gzclarence takes 5 minutes to build and solve. We need to profile it.
> 
> Some references: https://www.youtube.com/watch?v=BFvpwC8hEWQ&ab_channel=TheJuliaProgrammingLanguage https://www.youtube.com/watch?v=Okn_HKihWn8&ab_channel=TheJuliaProgrammingLanguage https://docs.julialang.org/en/v1/manual/profile/

Thanks a lot for the information!

clizbe:
@abelsiqueira I don't know if I can wait till tomorrow to find out... How's it going? 😶

abelsiqueira:
There were various improvements (#315, #317, #319). The benchmark script is maybe not working, so I am benchmarking locally. The NLBE example from @gzclarence is not up-to-date with the current CSV requirements, so I haven't tested on it yet.

abelsiqueira:
This has been done up to a reasonable point. With the changes since ~1-2 weeks ago, we have:

Old version:
- Create graph and RPs: ~50s
- Create constraints partitions: ~40x
- Create model: ~500s

New version:
- Create graph and RPs: < 1s
- Create constraints partitinos: ~3.5s
- Create model: ~35s

I don't know if we want to close this and open a new one to continue improving the model, or just change the priority.

abelsiqueira:
With #407 and #408, we have the model creation down to 3-5 seconds. The dataframes version is more complicated but is faster. After #112 we can verify again how they scale, but I am guessing that the dataframes version will be better.

PR #311, opened 2023-12-04
Update benchmarks

abelsiqueira:
# Pull request details

Update the benchmarks with more functions.
Add a LargerNorse case to the benchmarks, which is just a copy of Norse with one more RP with 1000 time steps.
Comment out the benchmarks of solving the problem, since it isn't an useful benchmark for us at the moment.

Useful for #294. Does not close anything

## Collaboration confirmation

As a contributor I confirm

- [x] Tests are passing
- [x] Lint is passing


datejada (commented):
I agree that we shouldn't solve the model for the benchmark (that's a solver thing). We need to benchmark the creation of the model as proposed here.

Regarding the LargeNorse, my concern is that we are changing the input file columns from time to time (e.g., adding new information). So, before creating a pull request, we also need to run the benchmark to ensure it works properly (like with the local coverage), which might be the solution. Or if we don't want to run the benchmark each time for all the commits, it would be good to create it automatically from the Norse input data and add a large rp to make it bigger. So, I'm leaning toward doing a hacky thing to create it instead of having the files in the repo...

Those are my thoughts; let me know what you think.

abelsiqueira:
You're right, let me do a hacky thing, then.

datejada (approved):
Great! Thanks

PR #315, opened 2023-12-04
Don't sort profile to avoid memory allocation

abelsiqueira:
# Pull request details

Sorting creates intermediary data, which makes it slow. Assuming that profiles are created in order will avoid this and make it faster.

Part of #294

## Collaboration confirmation

As a contributor I confirm

- [x] Tests are passing
- [x] Lint is passing

**Please test on the NLBE dataset locally**


abelsiqueira:
Looks like the benchmark is too small. I'm increasing it. Also, the Input_NLBE is not in the correct format anymore.

datejada (commented):
The profiles should always be sorted. If it becomes a problem, we can always handle that in the user interface or throw a warning/error.

For the time being, please add the comment in the function documentation saying that the profile must be sorted. I tried to add it, but since it is not part of the changes, I couldn't.

abelsiqueira:
Updated the docstring

PR #350, opened 2023-12-11
Precompute summation indices for the model expressions

abelsiqueira:
# Pull request details

To avoid using conditionals inside of summations in the expressions, we precompute them.
We also compute the duration.
This speed up the code, but maybe is it readable enough?

Related to #294 

## Collaboration confirmation

As a contributor I confirm

- [x] Tests are passing
- [x] Lint is passing

### DIFF
--- a/src/create-model.jl
+++ b/src/create-model.jl
@@ -1,5 +1,8 @@
 export create_model!, create_model
 
+const TupleAssetRPTimeBlock = Tuple{String,Int,TimeBlock}
+const TupleDurationFlowTimeBlock = Tuple{Float64,Tuple{String,String},TimeBlock}
+
 """
     create_model!(energy_problem)
 
@@ -24,7 +27,7 @@ function create_model(graph, representative_periods, constraints_partitions; wri
     ## Helper functions
     # Computes the duration of the `block` that is within the `period`, and
     # multiply by the resolution of the representative period `rp`.
-    # It is equivalent to finding the indexes of these values in the matrix.
+    # It is equivalent to finding the indices of these values in the matrix.
     function duration(B1, B2, rp)
         return length(B1 ∩ B2) * representative_periods[rp].resolution
     end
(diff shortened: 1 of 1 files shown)


abelsiqueira:
PS, the profiling has to be run locally with the EU data.

src/model.jl
```diff
@@ -116,43 +119,112 @@ function create_model(
...
         model,
         outgoing_flow_lowest_resolution[a ∈ A, rp ∈ RP, B ∈ Pl[(a, rp)]],
         sum(
-            duration(B, B_flow, rp) * flow[(a, v), rp, B_flow] for
-            v in outneighbor_labels(graph, a), B_flow ∈ graph[a, v].partitions[rp] if
-            B_flow[end] ≥ B[1] && B[end] ≥ B_flow[1]
+            d * flow[f, rp, B_flow] for
+            (d, f, B_flow) ∈ flow_sum_indexes_outgoing_lowest[(a, rp, B)]
         )
     )
```
clizbe, line 211: If this is so similar for incoming and outgoing, couldn't we generalize?

src/model.jl
```diff
@@ -209,19 +281,17 @@ function create_model(
...
     @expression(
         model,
         outgoing_flow_highest_resolution[a ∈ A, rp ∈ RP, B ∈ Ph[(a, rp)]],
         sum(
-            duration(B, B_flow, rp) * flow[(a, v), rp, B_flow] for
-            v in outneighbor_labels(graph, a), B_flow ∈ graph[a, v].partitions[rp] if
-            B_flow[end] ≥ B[1] && B[end] ≥ B_flow[1]
+            d * flow[f, rp, B_flow] for
+            (d, f, B_flow) in flow_sum_indexes_outgoing_highest[a, rp, B]
         )
```
clizbe, line 295: Same for this - can we generalize?

clizbe (commented):
Have to admit I don't really understand it. Maybe good to have Diego or Ni review?

src/model.jl
```diff
@@ -116,43 +119,112 @@ function create_model(
     ## Objective function
     @objective(model, Min, assets_investment_cost + flows_investment_cost + flows_variable_cost)
 
+    function add_to_flow_sum_indexes!(flow_sum_indexes, a, rp, B, f)
```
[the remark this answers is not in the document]
abelsiqueira: I am not sure it is normal, but I've used both in the past. I guess I was feeling like indexes today, but I can change to indices :)

src/model.jl
```diff
@@ -116,43 +119,112 @@ function create_model(
...
+                add_to_flow_sum_indexes!(
+                    flow_sum_indexes_outgoing_highest[a, rp, B],
+                    a,
+                    rp,
+                    B,
+                    (a, v),
+                )
+            end
+        end
+    end
```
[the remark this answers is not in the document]
abelsiqueira: Please don't take my word for it! I would be great if you could run the profile code to check for the speed up - so it isn't unilaterally informed by me. I can help with how to run it, if you want to do it, or you leave to Diego or Ni.
This is, allegedly, faster, because the `sum(... if ...)` structure inside the JuMP `@expression` is slow. It is faster to create the sets outside the loop, and then just use them. Furthermore, the loops inside the `@expression` were a double loop with non-contiguous values. Now, the loop inside the `@expression` is simpler.
The nested loops above are slow, but apparently not as slow as the older code.

src/model.jl
abelsiqueira: There is a way, but I would need to deconstruct the `@expression` into something else much less readable.

src/model.jl
abelsiqueira: Same answer

src/model.jl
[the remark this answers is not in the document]
abelsiqueira: Done

datejada:
In the SpineOpt model (MOPO project), they follow this same strategy, meaning that they have a function that creates the indices for the constraints as a dictionary and then use it to create the constraints in the model. That's the basic idea, but they have a different structure for the temporal resolution, so their functions to find the indices are quite different.
We can learn how to run the profiler during tomorrow's meeting and make a decision. The readability is not looking as bad as I expected 😄

abelsiqueira:
> The readability is not looking as bad as I expected 😄

So far... ;)

abelsiqueira:
Results copied my fork (https://github.com/abelsiqueira/TulipaEnergyModel.jl/pull/6#issuecomment-1887168319):

### Benchmark Results


|                                      | 5c5e5b8247a011... | 0db252a5e64ba0... | t[5c5e5b8247a011...]/t[0db252a5e64ba0...] |
|:-------------------------------------|:-----------------:|:-----------------:|:-----------------------------------------:|
| energy_problem/create_model          | 22.8 s            | 5.07 ± 0.32 s     | 4.49                                      |
| energy_problem/input_and_constructor | 2.6 ± 0.0058 s    | 2.61 ± 0.0092 s   | 0.994                                     |
| time_to_load                         | 7.28 ± 0.024 s    | 7.29 ± 0.035 s    | 0.998                                     |

abelsiqueira:
Closing in favor of #408

PR #359, opened 2023-12-13
Read flow graph from json representation of esdl

suvayu:
## Changes

- [x] get the list of edges (asset-asset) from the JSON representation of an ESDL file
- [ ] read the data associated to each edge, the model expects data like the columns 4 & onwards in [this table](../blob/main/test/inputs/Norse/flows-data.csv)
  - identify which attributes in ESDL store this data
  - currently the graph is built from a `Vector{Pair{Tuple{String, String}, GraphFlowData}}` (see [io.jl](https://github.com/suvayu/TulipaEnergyModel.jl/blob/8298b111aa8a042eeb4f99489fb472edc50062dd/src/io.jl#L82-L85)), where as `flow_from_json` in this PR returns `Vector{Tuple{String, String}}`
  - the current implementation gives you easy access to both the from/to `asset` when building the graph (see the closure `edges` defined inside `flow_from_json`). so if the additional attributes are part of the `asset` objects, it should be possible to extract them by updating this function
- [ ] combine the flow profiles with the flow data, this is currently done here: https://github.com/TulipaEnergy/TulipaEnergyModel.jl/blob/b90a567198220b9f959db97fabf0f9f366a6ec2e/src/io.jl#L137-L148  It would be good if we can refactor `create_graph_and_representative_periods_from_csv_folder` such that the graph creation is contained in isolated blocks of code, then we can do something along the lines of (or maybe multiple dispatch):
  ```julia
  # pseudo code
  if esdl
    flows = flow_from_json(...)
  else
    flows = flow_from_csv(...)
  end
  ```
- [ ] Read the asset data; equivalent to: https://github.com/TulipaEnergy/TulipaEnergyModel.jl/blob/b90a567198220b9f959db97fabf0f9f366a6ec2e/src/io.jl#L77-L90

## Related issues

Closes TulipaEnergy/TulipaIO.jl#1

## Collaboration confirmation

As a contributor I confirm

- [X] I read and followed the instructions in README.dev.md
- [ ] Tests are passing
- [x] Lint is passing


suvayu:
### Some notes on Asset attributes

Some (more general) attributes are actually "references".  In the JSON
representation it might look like this:

```json
{
  "eClass": "http://www.tno.nl/esdl#//PVInstallation",
  "id": "...",
  "name": "...",
  "technicalLifetime": 3.14,
  "costInformation": {
    "$ref": "//@instance.0/@path/@to/@definition",
    "eClass": "http://www.tno.nl/esdl#//CostInformation"
  }
}
```

In the above example, `technicalLifetime` is an attribute, whereas
`costInformation` is a reference.  Following the reference would give
us a JSON object, something like:

```json
{
  "eClass": "..."
  "fixedOperationalAndMaintenanceCosts": {
    "eClass": "http://www.tno.nl/esdl#//SomeKindOfProfile",
	"": "..."
  }
}
```

We need a loop inside the `edges` closure to resolve all the
references (recursion much!).

wcoenraads:
Yep, that's correct! Ecore libraries usually resolve those references for you, but we'll have to do it manually. It's predefined which attributes are plain attributes and which are references, e.g. see the "Attributes" and "References" tabs for [Battery](https://energytransition.github.io/#router/doc-content/687474703a2f2f7777772e746e6f2e6e6c2f6573646c/Battery.html). But I don't think we even need to do anything with that, references are easy enough to detect on the fly in the JSON.

suvayu:
Transferred TulipaIO.jl

PR #363, opened 2023-12-14
Updates on the feature section

### LINKED ISSUE #346: Add a new section in the features in the documentation for the graph representation
### What and Why

Add an example and explanation of the flexible assets/connection configuration possible to the model's graph representation.

We can reuse the example from the flexible resolution representation to show for example that you can connect the storage directly to the wind without any intermediate node in the middle (as in the classic node representation).

### Possible Drawbacks

NA

### Related Issues

_No response_

datejada:
# Pull request details

Updates on the explanation of the feature section

Closes #346

## Collaboration confirmation

As a contributor I confirm

- [ ] I read and followed the instructions in README.dev.md
- [ ] Tests are passing
- [ ] Lint is passing


datejada:
@gnawin please double-check that all we have commented before is included in this new version. There are extra changes that German commented on with me. Also, would you mind using the "add a suggestion" feature in your review to incorporate your suggestions into the documentation easily?

docs/src/features.md
```diff
@@ -7,26 +7,80 @@
...
+
+Let's consider the modeling approach in *TulipaEnergyModel.jl*. As nodes are no longer needed to connect assets, we can connect them directly to each other as shown in the figure below:
+
+![Flexible connection](./figs/flexible-connection-2.png)
+
+By implementing this approach, we can reduce the number of variables and constraints involved in the process. For example, the balance constraint in the intermediate node is no longer needed, as well as the extra constraint to avoid the storage charging from the power grid.  Additionally, we can eliminate the variable that determines the flow between the intermediate node and the power grid.
```
gnawin, line 31:
```suggestion
By implementing this approach, we can reduce the number of variables and constraints involved in the process. For example, the balance constraint in the intermediate node is no longer needed, as well as the extra constraint to avoid the storage charging from the power grid.  Additionally, we can eliminate the variable that determines the flow between the intermediate node and the power grid because the flow from `phs` to `balance` can directly link to the external grid.
```

docs/src/features.md
```diff
@@ -7,26 +7,80 @@
...
+
+Let's explore the flexibility in the time resolution with the following examples. The following table shows the user input data for the asset time resolution definition. Please note that the values presented in this example are just for illustrative purposes and do not represent a realistic case.
+
+```@example print-partitions
+using DataFrames # hide
+using CSV # hide
+input_asset_file = "../../test/inputs/Variable Resolution/assets-partitions.csv" # hide
+assets = CSV.read(input_asset_file, DataFrame, header = 2) # hide
+assets = assets[assets.asset .!= "wind", :] # hide
+```
```
gnawin, line 62: I'm not familiar with `# hide`. So just pointing out that you have this in every line. But most probably that is intentional.

docs/src/features.md
```diff
@@ -176,17 +230,57 @@ For the connection from the hub to the demand there are associated transmission
 ```math
```
gnawin, line 230: Sorry, I cannot seem to add a comment/suggestion to the part that you did not change. Actually I have a suggestion to be put at line 216 in this commit (in #### Producers Capacity Constraints, after the equations for `wind`)

> Since the flow variables `$flow_{(wind,balance),1:2}$` `$flow_{(wind,balance),1:3}$` represent power, the first constraint sets the upper bound of the power for both time step 1 and 2, by assuming an average capacity across these two time steps. The same applies to the other two constraints.

docs/src/features.md
```diff
@@ -176,17 +230,57 @@ For the connection from the hub to the demand there are associated transmission
...
+
+| Modeling approach                          | Nº Variables | Nº Constraints | Objective Function |
+|:-------------------------------------------|:------------ |:---------------|:------------------ |
+| Classic approach with hourly resolution    | 48           | 84             | 28.4365            |
+| Flexible connection with hourly resolution | 42           | 72             | 28.4365            |
+| Flexible connection and time resolution    | 16           | 25             | 28.4587            |
+
+By comparing the classic approach with the other methods, we can analyze their differences:
+
+- The flexible connection with hourly resolution reduces 6 variables ($12.5\%$) and 12 constraints ($\approx 14\%$). The objective function is the same since in both cases we use an hourly time resolution.
```
gnawin, line 279: The percentage sign does not seem to show correctly in my browser, maybe you can see the correct behavior in your local version.

gnawin (changes requested):
Nice work! I think you added lots of useful information to it. I have made some comments and suggestions.

docs/src/features.md
```diff
@@ -7,26 +7,80 @@
...
+
+The representation of the energy system in *TulipaEnergyModel.jl* is based on [Graph Theory](https://www.britannica.com/topic/graph-theory), which deals with connection amongst vertices by edges. This representation provides a more flexible framework to model *energy assets* in the system as *vertices* and *flows* between energy assets as *edges*. In addition, it poses some advantages from the modeling perspective. For instance, connecting assets directly to each other without having a node between them allows for a reduction of the number of variables and constraints to represent different configurations (e.g., co-location of energy storage and renewable units to model a hybrid operation).
+
+Consider the following example to demonstrate the benefits of this approach. In the classic connection approach, the nodes play a crucial role in modeling. For example, every asset needs to be connected to a node that has balance constraints attached to it. Therefore, to include the co-location of a storage asset and a renewable asset, we must introduce a node between these assets and the external power grid, as shown in the following figure:
```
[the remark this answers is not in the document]
datejada: Regarding your questions:
1. co-location refers to location and operation (technically connected). Another example, apart from the one in the documentation, is a battery + solar configuration, where both are in the exact location, and the battery goal is to reduce the solar PV curtailment, so it only charges from the solar PV (they can't charge from the grid). I think there is no standard name for this configuration (hybrid operation is also sometimes used, but maybe co-location is not a good name; we can come up with a better name if you have any other suggestions).
2. If they are both connected to the balance node directly, then you can't model the fact that the storage can't charge from the external grid. So, from the optimization point of view, it would be the same as if they were individual assets (and you wouldn't get that the storage only charges from the solar PV).
3. Both are balancing nodes (the thing is that we don't have "nodes" in Tulipa); I'm not good at giving names. I want a short name in one word. I am also trying to change the rest of the written section as little as possible. Anyhow, maybe I need to change the names. Any ideas?

Another example of co-location (or hybrid operation) is when you want to model green-hydrogen production from an electrolyzer + renewable configuration. You can only guarantee that the hydrogen production is green if the electrolyzer produces hydrogen directly from the renewable asset. Since renewable assets might also produce energy for the power grid, you need (as in the previous case) to avoid the electrolyzer consuming (charging) electricity from the grid.

We may need to add more details to the section. But we need to find a balance because I also don't want to explain too many concepts that are not the focus of the section. I'm open to suggestions.

docs/src/features.md
datejada: The `#hide` is to avoid showing this code in the documentation. Here, I only want to call the file and print it in a table. If you clone the branch and create the documentation locally, you will see how it renders. This is a trick from Abel 😉

docs/src/features.md
datejada: It is the same as in the `#hide` thing; you need to create it locally to see how it renders. At least they render correctly for me locally, so I assume they will do the same in GitHub when the documentation is created.

docs/src/features.md
datejada: I agree that that's needed there. Thanks! I'll add it.

gnawin (approved):
Thanks! This looks good to me.

PR #379, opened 2024-01-02
default_parameters now dispatches on Val types

### LINKED ISSUE #376: Change how the default solver parameter function works
### What and Why

Currently the `default_parameters` function uses multiple dispatch to determine the values.
This means that unless we load a solver, we can't define the default parameters for that solver, which does not make sense.

A solution would be dispatching on `Val{::Symbol}`, where the symbol is the name of the solver. This means that we can define the `default_parameters` for whatever we want.

An alternative would be using package extensions, however, it would be more complicated, and harder to maintain. 

### Possible Drawbacks

_No response_

### Related Issues

_No response_

abelsiqueira:
# Pull request details

This still allows specifying the default parameters using a function, while allowing defining default parameters of solvers that were not loaded.

Now all of these work:

- `default_parameters(Val(:HiGHS))`
- `default_parameters(HiGHS.Optimizer)`
- `default_parameters(:HiGHS)`
- `default_parameters("HiGHS")`

But only the first is dispatching on the type.

Closes #376 

## Collaboration confirmation

As a contributor I confirm

- [x] Tests are passing
- [x] Lint is passing

### DIFF
--- a/docs/Project.toml
+++ b/docs/Project.toml
@@ -2,6 +2,7 @@
 CSV = "336ed68f-0bac-5ca0-87d4-7b16caf5d00b"
 DataFrames = "a93c6f00-e57d-5684-b7b6-d8193f3e46c0"
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
+HiGHS = "87dc4568-4c63-4d18-b0c0-bb2238e4078b"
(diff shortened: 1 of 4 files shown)


datejada (approved):
Thanks for the PR. I approve it to go forward with #380

PR #380, opened 2024-01-02
Test Norse with Cbc and GLPK and conditionally with SCIP

### LINKED ISSUE #378: [BUG] SCIP doesn't work on Windows so testing fails
### Description

#377 introduced SCIP in the tests, but apparently it doesn't work on Windows:

https://github.com/TulipaEnergy/TulipaEnergyModel.jl/actions/runs/7386730638/job/20093938798#step:6:39

Best solution would be finding another open source solver good enough for the tests. Maybe Cbc or GLPK will handle it.

### Reproduction steps

```bash
See actions: https://github.com/TulipaEnergy/TulipaEnergyModel.jl/actions/runs/7386730638/job/20093938798#step:6:39
```

### Logs

_No response_

### OS

Windows

abelsiqueira:
# Pull request details

Remove SCIP.
Test Norse with Cbc and GLPK as well.

Closes #378 

## Collaboration confirmation

As a contributor I confirm

- [x] Tests are passing
- [x] Lint is passing

### DIFF
--- a/src/solve-model.jl
+++ b/src/solve-model.jl
@@ -37,16 +37,16 @@ Dict{String, Any} with 1 entry:
 This also
 
 ```jldoctest
-default_parameters(Val(:SCIP))
+default_parameters(Val(:Cbc))
 
 # output
 
 Dict{String, Any} with 1 entry:
-  "display/verblevel" => 0
+  "logLevel" => 0
 ```
 
 ```jldoctest
-default_parameters(:SCIP) == default_parameters("SCIP") == default_parameters(Val(:SCIP))
+default_parameters(:Cbc) == default_parameters("Cbc") == default_parameters(Val(:Cbc))
 
 # output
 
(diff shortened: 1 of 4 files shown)


abelsiqueira:
Needs to be rebased after #379

datejada:
Hi @abelsiqueira,

Cbc and GLPK are good enough for the testing in the GitHub actions. It is a pity that SCIP can't be used for testing in Windows. Is it possible to set up the GitHub action to test SCIP only for Linux and Mac?

abelsiqueira:
It is possible, but it would be a bit ugly. Something like:

- Create a EXTRA_SOLVERS environment variable in the actions storing the package/solver pairs.
- Install the packages in the action
- In the test code, get the packages and `using` them
- In the test code, get the solvers and loop through them

It would change things in four places, so I would not recommend just to test SCIP.
However, it might be useful in the future to also run tests on the server/remote machine that has the proprietary solvers, and in that case we might want to run the tests with the proprietary solvers as well. Assuming it will be a Linux machine, we can also test SCIP.
If that is feasible, we can create an issue for that.

datejada:
Great! Thanks for the explanation. Having the option of testing on "server/remote machine that has the proprietary solvers" is good for the future since we will use Xpress, and then we also might want to test that using one of our machines. So, let's go forward without SCIP and use Cbc and GLPK instead.

datejada (approved):
LGTM

PR #384, opened 2024-01-03
Create function to read parameters from file

### LINKED ISSUE #373: Allow reading solver options from a file
### Description

Allow the user to the options to be passed to the solver via a file.

### Related Issues

_No response_

### LINKED ISSUE #374: Explain how to use different solvers and set different options in the documentation
### What and Why

With some examples. 

### Possible Drawbacks

_No response_

### Related Issues

- #373
- #246

abelsiqueira:
# Pull request details

Reads from a file to create the dictionary of parameters.
Does NOT check validity of the keys.
Does NOT allow reading the solver itself, only parameters.
I had to disable the docstring formatting because it was not being able to parse] the `\"` symbols - I think.
Also move the parameter-related functions to a new file.
Finally, updates the docstring of `solve_model` to mention the parameter functions and to fix the example.

Closes #373
Closes #374 

## Collaboration confirmation

As a contributor I confirm

- [x] Tests are passing
- [x] Lint is passing


datejada:
@abelsiqueira Thanks, It looks good 😃 
For the documentation, would you mind adding a small subsection at the same level as the "Run scenario" in the Tutorial section showing how to run the scenario with a different solver from the default and some parameters for the solver? We can add the link to the function to read the file with the parameters and mention that they can create a file if needed.

https://tulipaenergy.github.io/TulipaEnergyModel.jl/dev/tutorial/#Run-scenario

Then, the documentation will be updated with the changes here.

Thanks!

clizbe:
Hmm that's an awful lot of caveats. Should Suvayu/Sander/Diego review this to maybe clean it up a bit? Or should we roll with this first version and fix the issues later?

datejada:
@clizbe, I think we can't do that much about this *"Does NOT check validity of the keys."* since it will depend on the solver, and it is hard to keep up with changes in their parameters. We can add a warning in the documentation that makes the user aware that he/she should provide valid parameters for the solver.

Regarding this caveat *Does NOT allow reading the solver itself, only parameters.*, I would also suggest including it in the warning to the user since it is also hard to control all possibilities, and it is better to give an example in the documentation (including the warning or note)

abelsiqueira:
I have updated the PR with the documentation suggested by @datejada. There is some intersection with #374.

@clizbe, it is like @datejada commented. We can't do a lot about these two changes easily. I included the caveats because if these are desired features, they should be new issues IMO, so we can evaluate feasibility and urgency.

docs/src/tutorial.md
```diff
@@ -108,6 +108,60 @@ solution = solve_model(model)
...
+
+or
+
+```@example manual
+using Cbc
+
+solution = solve_model(model, Cbc.Optimizer)
+```
+
+Notice that, in any of these cases, we need explicitly add the Cbc package
```
clizbe, line 142:
```suggestion
Notice that, in any of these cases, we need to explicitly add the Cbc package
```

src/solver-parameters.jl
```diff
@@ -0,0 +1,118 @@
...
+using HiGHS
+default_parameters(HiGHS.Optimizer)
+
+# output
+
+Dict{String, Any} with 1 entry:
+  "output_flag" => false
+```
+
+This also
```
clizbe, line 37:
```suggestion
Another case
```

src/solver-parameters.jl
```diff
@@ -0,0 +1,118 @@
...
+
+The file should contain a list of lines of the following type:
+
+```toml
+key = value
+```
+
+The file is parsed as [TOML](https://toml.io), which is very intuitive. See the examples below.
+
+## Examples
```
clizbe, line 82:
```suggestion
The file is parsed as [TOML](https://toml.io), which is very intuitive. See the example below.

## Example
```

clizbe (changes requested):
LGTM! Just a couple changes while reading through. :)

abelsiqueira:
Thanks for the review. I've accepted the suggestions.

datejada:
> I have updated the PR with the documentation suggested by @datejada. There is some intersection with #374.
> 
> @clizbe, it is like @datejada commented. We can't do a lot about these two changes easily. I included the caveats because if these are desired features, they should be new issues IMO, so we can evaluate feasibility and urgency.

Great! Thanks @abelsiqueira, I would say that this PR also Closes #374, so two birds with one stone 😄

datejada (approved):

PR #394, opened 2024-01-08
Add option to have the initial storage level as free value

### LINKED ISSUE #296: <title> Setting the initial stored energy free
### Description

From the discussion with @datejada , I know that currently the storage is modelled in a way that the stored energy is, by the end of the time steps, returned, to the initial value, from which the first time step evolves. However, at the moment, we need to set an initial value manually.

I would like to propose a way that no initial value is required. This can be simply done by using the stored energy at the final time step (yes, it is a decision variable) as the base from which the first time step evolves. Basically, to connect the first time step with the final one, for the storage type.

Does this make sense? Shall we add an input type to the initial value like "none", such that this proposed way will be activated?

Thanks,
Zhi

### Related Issues

_No response_

datejada:
# Pull request details

Include the option to allow missing values in the initial storage level to consider having it as a free value only depending on the last time block.

Closes #296 

## Collaboration confirmation

As a contributor I confirm

- [x] Tests are passing
- [x] Lint is passing

### DIFF
--- a/docs/src/how-to-use.md
+++ b/docs/src/how-to-use.md
@@ -63,6 +63,11 @@ end
 Markdown.parse(out)
 ```
 
+The `Missing` data meaning depends on the parameter, for instance:
+
+- `investment_limit`: there is no investment limit
+- `initial_storage_level`: the initial storage level is free (between the storage level limits), meaning that the optimization problem decides the best starting point for the storage asset. In addition, the first time block `B[1]` is linked to the last time block `B[end]` to create continuity in the storage level.
+
 #### `flows-data.csv`
 
 Similar to `assets-data.csv`, but for flows. Each flow is defined as a pair of assets.
@@ -77,6 +82,10 @@ for (f, t) in zip(fieldnames(TulipaEnergyModel.FlowData), fieldtypes(TulipaEnerg
 end
 Markdown.parse(out)
 ```
+
+The `Missing` data meaning depends on the parameter, for instance:
+
+- `investment_limit`: there is no investment limit
 
 #### `assets-profiles.csv`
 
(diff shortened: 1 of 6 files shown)


docs/src/how-to-use.md
```diff
@@ -63,6 +63,11 @@ end
 Markdown.parse(out)
 ```
 
+The `Missing` data meaning depends on the parameter, for instance:
+
+- `investment_limit`: there is no investment limit
+- `initial_storage_level`: the initial storage level is free (between the storage level limits), meaning that the optimization problem decides the best starting point for the storage asset. In addition, the first time block `B[1]` is linked to the last time block `B[end]` to create continuity in the storage level.
```
abelsiqueira, line 69:
```suggestion
- `initial_storage_level`: the initial storage level is free (between the storage level limits), meaning that the optimization problem decides the best starting point for the storage asset. In addition, the first and last time blocks in a representative period are linked to create continuity in the storage level.
```

datejada:
@abelsiqueira I had to pass the option of MIP gap = 0 to all the solvers since, with the latest changes, they don't get the same optimal solution (subject to numerical tolerances) if you don't force the MIP gap to be zero. Testing takes a little bit more time, but It is better that way. In addition, to test the default options of the solvers, I had to add Cbc and GLPK to the test of the Tiny case study. This is cheap in terms of time since that case study solves fast. Anyhow, let me know if you have a better idea or suggestion 😉

abelsiqueira (approved):
Thanks. I made a very small suggestion, but I am approving. Does it make sense to set the mip gap to 0 as default? Either way, I am approving and we can change that later if we want.

datejada:
@abelsiqueira, thanks for the review. I also hesitate about putting the MIP gap = 0 as a default. The downside of including it is that if the problem is large-scale, it might take too much time, so solvers typically have a default of 0.01%, but then the solutions can be different for the tolerance we are testing the problems (i.e., 1e-5). We can discuss it in the next working session and change it if we all agree to include it as default.

PR #446, opened 2024-02-07
Add input data for seasonal storage (i.e., long-term storage) 

### LINKED ISSUE #255: Add linked-representative days formulation for seasonal (long-term) storage
### What and Why

This is one of the core features of the model. It is based on the following formulation:

https://ieeexplore.ieee.org/abstract/document/8334256

We need to translate this to the current formulation of Tulipa.

### Possible Drawbacks

We need more input data and possibly a new file.

Also, @abelsiqueira, I need your help to tackle these changes (maybe we can use the same logic of ranges to simply the proposed formulation in the paper)

### Related Issues

- [x] #455 
- [x] #456
- [x] #509
- [x] #512
- [x] #516
- [x] #522
- [x] #523
- [x] #514
- [x] #549
- [x] #458 
- [x] TulipaEnergy/TulipaPlots.jl/issues/15

### LINKED ISSUE #455: Update the input data and read it into the model for the linked-representative days
### Description

Include the new parameters and read them into the model

### Related Issues

Part of EPIC issue #255

datejada:
# Pull request details

For implementation, we need to do the following:

- [x] Update the input files to consider a new column with the new `storage_type` info and the moving window per asset
- [x] Update the structure and i/o to read the new column in the files
- [x] Define new subsets for the short- and long-term storage (subsets of storage assets)
- [x] New set with the base periods (the ones we want to represent with the rp)

Closes #455 

## Collaboration confirmation

As a contributor I confirm

- [x] Tests are passing
- [x] Lint is passing


datejada:
@abelsiqueira I'm halfway through these changes. So far, I have updated the files and the code to account for the data I need to formulate the new constraints. Since there are several changes, it is best that you give your feedback up to this point. I consider some things that can be improved, but I need to figure out how.
This will help to make the review process more straightforward. I tried to make the commits as atomic as possible so that it is easier for you to see the changes individually and the reason for them.
Thanks!

src/create-model.jl
```diff
@@ -205,7 +205,10 @@ function create_model(graph, representative_periods, dataframes; write_lp_file =
     ## Sets unpacking
     A = labels(graph) |> collect
     F = edge_labels(graph) |> collect
-    filter_assets(key, value) = filter(a -> getfield(graph[a], key) == value, A)
+    filter_assets(key, value) = filter(
+        a -> getfield(graph[a], key) === missing ? false : getfield(graph[a], key) == value,
+        A,
+    )
```
abelsiqueira, line 224: There are two slightly cleaner ways to do this:

```julia
a -> !ismissing(getfield(graph[a], key)) && getfield(graph[a], key) == value
```
So you check using `ismissing`, which simplifies `=== missing`, and then use `&&` which [short-circuits](https://en.wikipedia.org/wiki/Short-circuit_evaluation) to avoid the ternary operator.

Or
```julia
a -> coalesce(getfield(graph[a], key) == value, false)
```
Which will compute `getfield(graph[a], key) == value` and then `coalesce` will find the first non-missing argument. It is either `true` or `false` from the first argument, or `false` from the second.
Given the size of `A` and the number of times that we will be calling these operations, the cost is negligible, so I think it boils down to what looks better. 

PS. Code not tested

src/create-model.jl
```diff
@@ -217,6 +220,17 @@ function create_model(graph, representative_periods, dataframes; write_lp_file =
...
 
+    # Create subsets of storage assets by storage type
+    As_short = intersect(As, filter_assets(:storage_type, "short"))
+    As_long  = intersect(As, filter_assets(:storage_type, "long"))
+
+    # Check if all assets in As are in As_short and As_long, if not assign the missing ones to short by default
+    all_storage_assets_in_short_and_long = all(a -> a ∈ union(As_short, As_long), As)
+    if !all_storage_assets_in_short_and_long
+        As_short = union(As_short, setdiff(As, union(As_short, As_long)))
+    end
```
abelsiqueira, line 246: This means that some storage types won't be "short" nor "long". Is that allowed? If so, ignore this comment, otherwise, we should create a check preventing this instead of this code.

src/create-model.jl
```diff
@@ -217,6 +220,17 @@ function create_model(graph, representative_periods, dataframes; write_lp_file =
     Fi = filter_flows(:investable, true)
     Ft = filter_flows(:is_transport, true)
 
+    # Create subsets of storage assets by storage type
+    As_short = intersect(As, filter_assets(:storage_type, "short"))
+    As_long  = intersect(As, filter_assets(:storage_type, "long"))
+
+    # Check if all assets in As are in As_short and As_long, if not assign the missing ones to short by default
+    all_storage_assets_in_short_and_long = all(a -> a ∈ union(As_short, As_long), As)
```
abelsiqueira, line 243: This is very small here, but it might become more common in the future. Instead of `a in union(A, B)`, use `a in A || a in B`, otherwise the memory for `union(A, B)` is created.

src/create-model.jl
abelsiqueira: Also, it looks like we could define

```julia
As_long = ... # as it is
As_short = setdiff(As, As_long)
```
i.e., the "long" ones are in `As_long` and the others ("short" or anything else) will be in `As_short`.

src/create-model.jl
```diff
@@ -211,14 +211,16 @@ function create_model(graph, representative_periods, dataframes; write_lp_file =
...
+    Ap  = filter_assets(:type, "producer")
+    As  = filter_assets(:type, "storage")
+    Ah  = filter_assets(:type, "hub")
     Acv = filter_assets(:type, "conversion")
-    Fi = filter_flows(:investable, true)
-    Ft = filter_flows(:is_transport, true)
+    Ft  = filter_flows(:is_transport, true)
+
+    # Create subsets of assets by investable
+    Ai = intersect(union(Ac, Ap, As, Ah, Acv), filter_assets(:investable, true))
```
abelsiqueira, line 235: This also reads as something that could be enforced in validation. Should "conversion" assets be allowed to be "investable"?

src/create-model.jl
```diff
@@ -211,14 +211,16 @@ function create_model(graph, representative_periods, dataframes; write_lp_file =
...
+    Fi = intersect(Ft, filter_flows(:investable, true))
```
abelsiqueira, line 236: Maybe a similar comment applies here, although only the `Ft` is defined, so maybe not.

src/io.jl
```diff
@@ -46,6 +47,9 @@ The returned structures are:
 
   - `representative_periods`: An array of
     [`TulipaEnergyModel.RepresentativePeriod`](@ref) ordered by their IDs.
+
+    - `base_periods`: Information of
```
abelsiqueira, line 51: Indentation.

src/structures.jl
```diff
@@ -1,7 +1,14 @@
-export GraphAssetData, GraphFlowData, RepresentativePeriod, TimeBlock
+export GraphAssetData, GraphFlowData, RepresentativePeriod, BasePeriod, TimeBlock
 
 const TimeBlock = UnitRange{Int}
 
+"""
+Structure to hold the data of the base periods.
+"""
+struct BasePeriod
+    num_base_periods::TimeBlock
```
abelsiqueira, line 9: Since I don't know what is coming in the next commit, this might not be applicable.
`num_base_periods` implies that this is a number. `TimeBlock` is a set. I don't know what `base_periods` will be used for, so for now I am leaving it like this.

abelsiqueira (commented):
Left some comments. Not sure what comes next, but I guess we can discuss the details on Monday.

src/create-model.jl
datejada: Great! I have implemented the first option 👍

src/create-model.jl
datejada: This is much simpler and easy to read. Thanks! I've changed it.

src/create-model.jl
datejada: It is no longer in the code, but it is good to keep it in mind.

src/create-model.jl
datejada: Let's discuss how to start doing the data validation in the model. So, depending on that, we might change it so the validation is done in another part of the code.

src/create-model.jl
datejada: It is still relevant if we create a validation step.

src/io.jl
datejada: solved! thanks

src/structures.jl
datejada: You are right; it is easy to handle it as a number. Thanks! I've changed it in the code.

src/create-model.jl
datejada: I added the validation to #396

abelsiqueira (approved):
LGTM, good luck 🍰

PR #482, opened 2024-02-20
Display timer for energy problem

### LINKED ISSUE #474: Add to the EnergyProblem structure the time to create the problem and the time to solve
### What and Why

The current function below is quite useful, and it will be even more if we add the time to create the problem and the time to solve 😄 

```julia
function Base.show(io::IO, ep::EnergyProblem)
    println(io, "EnergyProblem:")
    println(io, "  - Model created: ", !isnothing(ep.model))
    println(io, "  - Solved: ", ep.solved)
    println(io, "  - Termination status: ", ep.termination_status)
    println(io, "  - Objective value: ", ep.objective_value)
end
```

For the solving time we can use this JuMP function:

```julia
solve_time(energy_problem.model)
```

For the time to create the model, there are several options so that we might ask @abelsiqueira for a suggestion. I like the function `@timed`, but maybe there is a better option, hopefully in Julia Base (to avoid adding more dependencies to the packages for these simple things).

### Possible Drawbacks

Measuring the create problem might include a small overhead to the time. Hopefully, it's not significant.

Also, please double-check that when the problem is infeasible what the solar returns in the time to solve (I think, it is okay, but just to double-check it)

### Related Issues

_No response_

gnawin:
# Pull request details

Add timer for `EnergyProblem`

- Time for creating the model
- Time for solving the model

Closes #474 

## Collaboration confirmation

As a contributor I confirm

- [x] Tests are passing
- [x] Lint is passing

### DIFF
--- a/src/run-scenario.jl
+++ b/src/run-scenario.jl
@@ -35,7 +35,8 @@ function run_scenario(
         @timeit to "solve_model!" solve_model!(energy_problem, optimizer; parameters = parameters)
     end
 
-    energy_problem.time_create_model = elapsed_time_read_data + elapsed_time_create_model
+    energy_problem.time_read_data    = elapsed_time_read_data
+    energy_problem.time_create_model = elapsed_time_create_model
     energy_problem.time_solve_model = elapsed_time_solve_model
 
     if output_folder != ""
--- a/src/run-scenario.jl
+++ b/src/run-scenario.jl
@@ -39,9 +39,7 @@ function run_scenario(
     energy_problem.time_solve_model = elapsed_time_solve_model
 
     if output_folder != ""
-        elapsed_time = @elapsed begin
-            @timeit to "save_solution_to_file" save_solution_to_file(output_folder, energy_problem)
-        end
+        @timeit to "save_solution_to_file" save_solution_to_file(output_folder, energy_problem)
     end
 
     show_log && show(to)
--- a/src/run-scenario.jl
+++ b/src/run-scenario.jl
@@ -37,7 +37,7 @@ function run_scenario(
 
     energy_problem.time_read_data    = elapsed_time_read_data
     energy_problem.time_create_model = elapsed_time_create_model
-    energy_problem.time_solve_model = elapsed_time_solve_model
+    energy_problem.time_solve_model  = elapsed_time_solve_model
 
     if output_folder != ""
         @timeit to "save_solution_to_file" save_solution_to_file(output_folder, energy_problem)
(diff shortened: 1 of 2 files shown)


src/run-scenario.jl
```diff
@@ -20,12 +20,23 @@ function run_scenario(
...
+
+    elapsed_time_create_model = @elapsed begin
+        @timeit to "create_model!" create_model!(energy_problem; write_lp_file = write_lp_file)
+    end
+
+    elapsed_time_solve_model = @elapsed begin
+        @timeit to "solve_model!" solve_model!(energy_problem, optimizer; parameters = parameters)
+    end
+
+    energy_problem.time_create_model = elapsed_time_read_data + elapsed_time_create_model
```
datejada, line 38:
```suggestion
    energy_problem.time_read_data    = elapsed_time_read_data
    energy_problem.time_create_model = elapsed_time_create_model
```

src/structures.jl
```diff
@@ -156,6 +156,8 @@ mutable struct EnergyProblem
     solved::Bool
     objective_value::Float64
     termination_status::JuMP.TerminationStatusCode
+    time_create_model::Float64
```
datejada, line 159:
```suggestion
    time_read_data::Float64
    time_create_model::Float64
```

src/structures.jl
```diff
@@ -176,6 +178,8 @@ mutable struct EnergyProblem
             false,
             NaN,
             JuMP.OPTIMIZE_NOT_CALLED,
+            NaN,
```
datejada, line 181:
```suggestion
            NaN,
            NaN,
```

src/structures.jl
```diff
@@ -186,4 +190,6 @@ function Base.show(io::IO, ep::EnergyProblem)
     println(io, "  - Solved: ", ep.solved)
     println(io, "  - Termination status: ", ep.termination_status)
     println(io, "  - Objective value: ", ep.objective_value)
+    println(io, "  - Time for creating the model (in seconds): ", ep.time_create_model)
```
datejada, line 193:
```suggestion
    println(io, "  - Time for reading the data (in seconds): ", ep.time_read_data)
    println(io, "  - Time for creating the model (in seconds): ", ep.time_create_model)
```

datejada (commented):
Thanks for the PR. Seeing your changes, I believe it is better to have the time to read the data apart from creating the model. So, please check my suggestions, and if you agree, please update the PR.

datejada (approved):
It is better to have these times in different parameters, since they are timing different parts of the code. So, this is PR looks good to me now 😉

PR #519, opened 2024-03-04
Add optimization problem size info to the display of `EnergyProblem`

### LINKED ISSUE #510: Show more solver info in `run_scenario`
### What and Why

`run_scenario` does not include solver log such as the number of constraints, variables and solving status, which can be inconvenient. Or to add this info to the display of `EnergyProblem`.

JuMP has functions to display the number of variables and constraints (https://jump.dev/JuMP.jl/stable/api/JuMP/#num_constraints). The easiest solution is to add this info to the display.

### Possible Drawbacks

_No response_

### Related Issues

_No response_

gnawin:
# Pull request details

Add more outputs to the display of `EnergyProblem`

Closes #510 

## Collaboration confirmation

As a contributor I confirm

- [x] Tests are passing
- [x] Lint is passing

### DIFF
--- a/src/structures.jl
+++ b/src/structures.jl
@@ -232,6 +232,25 @@ end
 function Base.show(io::IO, ep::EnergyProblem)
     println(io, "EnergyProblem:")
     println(io, "  - Model created: ", !isnothing(ep.model))
+    println(io, "  - Number of variables: ", !isnothing(ep.model) ? num_variables(ep.model) : NaN)
+    println(
+        io,
+        "  - Number of constraints (excluding bounds): ",
+        if !isnothing(ep.model)
+            num_constraints(ep.model; count_variable_in_set_constraints = false)
+        else
+            NaN
+        end,
+    )
+    println(
+        io,
+        "  - Number of constraints (including bounds): ",
+        if !isnothing(ep.model)
+            num_constraints(ep.model; count_variable_in_set_constraints = true)
+        else
+            NaN
+        end,
+    )
     println(io, "  - Solved: ", ep.solved)
     println(io, "  - Termination status: ", ep.termination_status)
     println(io, "  - Objective value: ", ep.objective_value)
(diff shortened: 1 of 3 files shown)


datejada (changes requested):
Thanks @gnawin. There are two comments:

1. Since there are several `!isnothing(ep.model)` it is better to have an if statement at the start of the function and only print the relevant information for each case. For example, if it is not solved, only show the "Model created". If you solve the model, then you show all the numbers. This will be it; you don't need to check for each line if the model is nothing.
2. instead of having the number of constraints including the bounds, please print the number of bounds, which is the difference between the values you currently have.

Thanks!

gnawin:
Thank you for the comments @datejada. Before I make further changes, I have one doubt though. Adding if statements to check the cases (e.g., model created, model solved) will make the display neater, but the code will be more repetitive, i.e., you put largely similar codes under each if-else block. 

In summary, either you print all the info, the display will be less neat but the code will be more compact, or you print for each case, the display will be neater but the code will be less compact. Do we want to opt for the latter? Or is there a 3rd way of implementation?

datejada:
Hi @gnawin, @abelsiqueira and I commented that we want to avoid printing NaN since the model has yet to be solved. So, that is the idea behind the first point. You should only print your information and avoid NaN, empty, etc. Does that make sense?

gnawin:
Hi @datejada @abelsiqueira, thanks for the clarification. 

Yes, that makes total sense. I was hesitating because I felt that as a result of not printing NaN, the code would be less compact. But after the latest implementation, I think it looks okay.

datejada:
Great! Thanks, we are almost there. I agree with all the changes, but the codecov is still complaining because it was not triggered with your last commit. Something similar happened to Abel on Monday, he manage to trigger it again by rebasing the branch and pushing again. Could you try that? if we make the codecov happy in te PR, then we can merge 😃

gnawin:
@datejada Great! I did that and now codecov is happy 😄 .

datejada (approved):
Awesome! Thanks

PR #535, opened 2024-03-12
Make the number of samples explicit in the benchmark

### LINKED ISSUE #534: Update benchmark options
### What and Why

The `@benchmarkable` command can have the `samples`, `evals`, and `seconds` specified as such:

```
SUITE[...] = @benchmarkable <cmd> samples = X evals = Y seconds = Z
```

We should set to something like `X = 3`, `Y = 1` and `Z = Inf`. More info below:

The benchmark will run `X` times (i.e., `X` measurements are made), each one runs the code `Y` times and computes the average, except that if the timing between samples exceeds `Z`, it will be stopped.
Since the code is slow enough to have significant timing, setting `Y = 1` is enough, but we should set `Z = Inf` to allow a larger number of samples to be run.
Setting `X = 3` should be enough to have very basic variance information. Note that this also make the benchmarking `X` times slower, so probably we don't want to increase it a lot.

We should also remove the `--tune` option from the .yml, since it will not be required anymore.

### Possible Drawbacks

_No response_

### Related Issues

_No response_

abelsiqueira:
# Pull request details

Use `samples = 3 evals =1 seconds = Inf` in the benchmarks, and don't tune.

Closes #534 

## Collaboration confirmation

As a contributor I confirm

- [x] Tests are passing
- [x] Lint is passing

### DIFF
--- a/.github/workflows/Benchmark.yml
+++ b/.github/workflows/Benchmark.yml
@@ -70,8 +70,7 @@ jobs:
                     --rev="$BASE,$HEAD" \
                     --url=${{ github.event.repository.clone_url }} \
                     --bench-on="$BASE" \
-                    --output-dir=results/ \
-                    --tune
+                    --output-dir=results/
             - name: Create plots from benchmarks
               run: |
                   mkdir -p plots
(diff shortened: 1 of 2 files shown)


clizbe (approved):
LGTM!

PR #545, opened 2024-03-15
Use Symbols instead of Strings for many cases

### LINKED ISSUE #540: Consider using Symbol instead of String for columns and other identifier
### What and Why

For instance, an `asset` would be a `Symbol` instead of a `String`.
Some keys of dictionaries are also Strings, such as `rep` and `base`.

More info on symbol vs string: https://bkamins.github.io/julialang/2022/08/05/symbol.html

### Possible Drawbacks

_No response_

### Related Issues

_No response_

abelsiqueira:
# Pull request details

Most notable cases are the names and types of assets and profiles. Some loose keys were also included in this. Note that partition definitions are still Strings, since they will be manipulated as Strings.
The only file change was the name of the max and min profiles, since they had dashes in their name. To avoid changing the file name, I sanitize it to convert - to _.

Closes #540 

## Collaboration confirmation

As a contributor I confirm

- [x] Tests are passing
- [x] Lint is passing

### DIFF
--- a/docs/src/tutorial.md
+++ b/docs/src/tutorial.md
@@ -200,14 +200,14 @@ labels(graph) |> collect
 To access the asset data, we can index the graph with an asset label:
 
 ```@example manual
-graph["ocgt"]
+graph[:ocgt]
 ```
 
 This is a Julia struct, or composite type, named [GraphAssetData](@ref).
 We can access its fields with `.`:
 
 ```@example manual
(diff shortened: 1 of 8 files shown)


datejada (approved):
Thanks! That was fast 😄 and it is fine to have the partition definition still as Strings. Later, we can see if it makes sense to refactor a bit to have them also as Symbols.

PR #564, opened 2024-03-25
Interpolate storage levels

### LINKED ISSUE #553: Export the storage level solution as an interpolation
### What and Why

The current implementation (see code below) in the `save_solution_to_file` function replicates the storage level value at the end of the block according to the duration of the block (either the time period block or the base periods block). We should interpolate between the storage level value at the begining of the block (i.e., $s_{t-1}$ or the end level of the previous block) and the storage level value at the end od the block (i.e., $s_{t}$).

The tricky case is the initial storage level, since it could be a value from the input data or could it be from the cycling constraint (last period equals is the starting point of the first one). See the storage balance constraints to have a better idea for this part.

```julia
    output_table.value = solution.storage_level_intra_rp
    output_table = DataFrames.flatten(
        DataFrames.transform(
            output_table,
            [:time_step, :value] =>
                DataFrames.ByRow(
                    (time_block, value) -> begin
                        n = length(time_block)
                        (time_block, Iterators.repeated(value, n))
                    end,
                ) => [:time_step, :value],
        ),
        [:time_step, :value],
    )
    output_table |> CSV.write(output_file)

    output_file = joinpath(output_folder, "storage-level-inter-rp.csv")
    output_table = DataFrames.select(
        dataframes[:storage_level_inter_rp],
        :asset,
        :base_period_block => :time_step,
    )
    output_table.value = solution.storage_level_inter_rp
    output_table = DataFrames.flatten(
        DataFrames.transform(
            output_table,
            [:time_step, :value] =>
                DataFrames.ByRow(
                    (base_period_block, value) -> begin
                        n = length(base_period_block)
                        (base_period_block, Iterators.repeated(value, n))
                    end,
                ) => [:time_step, :value],
        ),
        [:time_step, :value],
    )
```

### Possible Drawbacks

N.A.

### Related Issues

_No response_

gnawin:
# Pull request details

Add functions to interpolate storage levels

Closes #553 

## Collaboration confirmation

As a contributor I confirm

-   [x] Tests are passing
-   [x] Lint is passing

### DIFF
--- a/src/io.jl
+++ b/src/io.jl
@@ -450,7 +450,8 @@ function _interpolate_storage_level!(df, time_column::Symbol)
                 DataFrames.ByRow(
                     (period, value, start_value) -> begin
                         n = length(period)
-                        interpolated_values = range(start_value, value, n + 1)
+                        interpolated_values =
+                            range(; start = start_value, stop = value, length = n + 1)
                         (period, value, interpolated_values[2:end])
                     end,
                 ) => [time_column, :value, :processed_value],
--- a/src/io.jl
+++ b/src/io.jl
@@ -450,8 +450,7 @@ function _interpolate_storage_level!(df, time_column::Symbol)
                 DataFrames.ByRow(
                     (period, value, start_value) -> begin
                         n = length(period)
-                        interpolated_values =
-                            range(; start = start_value, stop = value, length = n + 1)
+                        interpolated_values = range(start_value; stop = value, length = n + 1)
                         (period, value, interpolated_values[2:end])
                     end,
                 ) => [time_column, :value, :processed_value],
--- a/src/io.jl
+++ b/src/io.jl
@@ -442,7 +442,7 @@ Tranform the storage level dataframe from grouped timesteps or periods to increm
 The starting value is the value of the previous grouped timesteps or periods or the initial value.
 The ending value is the value for the grouped timesteps or periods.
 """
-function _interpolate_storage_level!(df, time_column::Symbol)
+function _interpolate_storage_level!(df, time_column)
     DataFrames.flatten(
         DataFrames.transform(
             df,
--- a/src/io.jl
+++ b/src/io.jl
@@ -397,9 +397,11 @@ function save_solution_to_file(output_folder, graph, dataframes, solution)
         :timesteps_block => :timestep,
     )
     output_table.value = solution.storage_level_intra_rp
-    output_table = DataFrames.combine(DataFrames.groupby(output_table, :asset)) do subgroup
-        _check_initial_storage_level!(subgroup, graph)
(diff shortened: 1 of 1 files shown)


gnawin:
Observations when doing this PR:

Since functions are added for the storages, maybe the code can be simplified further by refactoring the flow part also into functions (see below code where the flatten and transform are use, which are largely also in a function now). 

```
    output_file = joinpath(output_folder, "flows.csv")
    output_table =
        DataFrames.select(dataframes[:flows], :from, :to, :rp, :timesteps_block => :timestep)
    output_table.value = solution.flow
    output_table = DataFrames.flatten(
        DataFrames.transform(
            output_table,
            [:timestep, :value] =>
                DataFrames.ByRow(
                    (timesteps_block, value) -> begin # transform each row using these two columns
                        n = length(timesteps_block)
                        (timesteps_block, Iterators.repeated(value, n)) # e.g., (3:5, [30, 30, 30])
                    end,
                ) => [:timestep, :value],
        ),
        [:timestep, :value], # flatten, e.g., [(3, 30), (4, 30), (5, 30)]
    )
    output_table |> CSV.write(output_file)
```

It will make the code cleaner, but that does not save many codes (since some conditions would need to be added to the functions) and may lose more readability (and maybe performance). 

If we want to do that, I propose to have a separate issue.

datejada:
> Observations when doing this PR:
> 
> Since functions are added for the storages, maybe the code can be simplified further by refactoring the flow part also into functions (see below code where the flatten and transform are use, which are largely also in a function now).
> 
> ```
>     output_file = joinpath(output_folder, "flows.csv")
>     output_table =
>         DataFrames.select(dataframes[:flows], :from, :to, :rp, :timesteps_block => :timestep)
>     output_table.value = solution.flow
>     output_table = DataFrames.flatten(
>         DataFrames.transform(
>             output_table,
>             [:timestep, :value] =>
>                 DataFrames.ByRow(
>                     (timesteps_block, value) -> begin # transform each row using these two columns
>                         n = length(timesteps_block)
>                         (timesteps_block, Iterators.repeated(value, n)) # e.g., (3:5, [30, 30, 30])
>                     end,
>                 ) => [:timestep, :value],
>         ),
>         [:timestep, :value], # flatten, e.g., [(3, 30), (4, 30), (5, 30)]
>     )
>     output_table |> CSV.write(output_file)
> ```
> 
> It will make the code cleaner, but that does not save many codes (since some conditions would need to be added to the functions) and may lose more readability (and maybe performance).
> 
> If we want to do that, I propose to have a separate issue.

I like the proposal; please create the issue and explain as much as possible so we can easily pick it up later. Thanks!

datejada (approved):
Thanks! Having functions with meaningful names improves a lot the readability; I like it 😄

PR #582, opened 2024-04-08
Modify investment limit to integer when necessary

### LINKED ISSUE #577: Improve investment limit constraint
### What and Why

For the maximum investment limit for assets and flows, here:

https://github.com/TulipaEnergy/TulipaEnergyModel.jl/blob/main/src/constraints/investment.jl

We need to add a condition for the right-hand side (RHS) of the investment limit constraints, such that if the parameter `investment_integer` is `true`, then we limit it to the lowest integer value of the fraction using a ceiling function, to guarantee that the limit is an integer. Otherwise, it can stay as it is today.

### Possible Drawbacks

- Be careful of the codecov! Since we are creating a new condition. 
- Run the benchmark to double-check any potential performance issues.

### Related Issues

_No response_

gnawin:
# Pull request details

Add floor function to investment limit when investment is integer, otherwise it stays the same.

Closes #577 

## Collaboration confirmation

As a contributor I confirm

-   [ ] Tests are passing
-   [ ] Lint is passing


datejada:
It seems that the benchmark runs correctly TulipaEnergyModel, but then, processing the results, there is an error (a problem with an `Inf` in JSON); see the screenshot below. I'm not sure how to solve it. So, I suggest waiting to @abelsiqueira return from holiday to help us with this:

image

gnawin:
The benchmark results show that the extra conditions slightly increase the time for `energy_problem/create_model` and `energy_problem/input_and_constructor`, and decreases it for `time_to_load`. I don't really know the difference between these, so I'll leave it for someone who knows to check and review.

src/constraints/investment.jl
```diff
@@ -13,7 +13,11 @@ function add_investment_constraints!(graph, Ai, Fi, assets_investment, flows_inv
         if graph[a].capacity > 0 && !ismissing(graph[a].investment_limit)
             JuMP.set_upper_bound(
                 assets_investment[a],
-                graph[a].investment_limit / graph[a].capacity,
+                if graph[a].investment_integer
+                    floor(graph[a].investment_limit / graph[a].capacity)
+                else
+                    graph[a].investment_limit / graph[a].capacity
+                end,
```
abelsiqueira, line 20: Add something like
```julia
value = graph[a].investment_limit / graph[a].capacity
if graph[a].investment_integer
    value = floor(value)
end
```
outside the function call and just use `value` instead. `bound_value` might be a better name, but since it is inside the `if`, it doesn't matter a lot.

src/constraints/investment.jl
```diff
@@ -23,7 +27,11 @@ function add_investment_constraints!(graph, Ai, Fi, assets_investment, flows_inv
         if graph[u, v].capacity > 0 && !ismissing(graph[u, v].investment_limit)
             JuMP.set_upper_bound(
                 flows_investment[(u, v)],
-                graph[u, v].investment_limit / graph[u, v].capacity,
+                if graph[u, v].investment_integer
+                    floor(graph[u, v].investment_limit / graph[u, v].capacity)
+                else
+                    graph[u, v].investment_limit / graph[u, v].capacity
+                end,
```
abelsiqueira, line 34: Same as above.

abelsiqueira (changes requested):
Thanks for the PR, here is a short comment.

gnawin:
@abelsiqueira thanks for the review! Based on your review, I added a function to calculate the bounds, how does it look now? 

For me to learn 😄, an extra question is that, instead of using if-else, you suggest to set an default value first and then check for "outliers", is this a better way of programming in general, or is it due to performance concerns or some other reasons?

abelsiqueira:
> For me to learn 😄, an extra question is that, instead of using if-else, you suggest to set an default value first and then check for "outliers", is this a better way of programming in general, or is it due to performance concerns or some other reasons?

Nothing general in this case. Here I was trying to avoid a long expression from appearing twice, so it looks easier to read.
No performance implications happen either. Also, this was only recommended because the type of the value did not change. If, instead, we had `value = round(Int, value)`, then the type would change to `Int`, which is bad in Julia.

If the expression was shorter, or involved different enough terms, I would have suggested a single-line solution using the ternary operator: `value = cond ? a : b`.

The change also reads fine in both situations. Originally, it was "if the investment is not integer, then its bound is A over B, otherwise, then the bound is the rounded value of A over B". Now it reads as "The bound is A over B. If the investment is integer, the round the bound".

abelsiqueira (approved):
Thanks!

PR #590, opened 2024-04-16
Change Inf to 86400 in the benchmarks.jl script and benchmark on the HEAD script

### LINKED ISSUE #588: [BUG] Benchmarks are failing with Inf not supported by JSON
### Description

As seen in #582.
I suspect there are some interactions going on with new versions of packages and the changes we made adding "seconds = Inf" to `Benchmark.yml` in #534/#535. Simplest solution will be changing "Inf" to a large enough value, such as 86400 seconds (1 day).

### Reproduction steps

See #582

### Logs

_No response_

### OS

_No response_

abelsiqueira:
# Pull request details

Inf is not allowed in the JSON file used internally by AirspeedVelocity.jl. The BenchmarkTools package seems to handle it properly when saving, but AirspeedVelocity bypass that and uses JSON directly.
We explicitly use Inf in the benchmarks.jl script, to the simplest tentative solution is to change that to another value.

This PR changes Inf to 86400, which should be enough for any reasonable benchmark that we do.
This PR also changes the bench-on parameter to $HEAD to indicate that we run the script given in the PR (otherwise we can't test this change before merging).

Closes #588 

## Collaboration confirmation

As a contributor I confirm

-   [x] Tests are passing
-   [x] Lint is passing

### DIFF
--- a/.github/workflows/Benchmark.yml
+++ b/.github/workflows/Benchmark.yml
@@ -69,7 +69,7 @@ jobs:
                   benchpkg ${{ steps.extract-package-name.outputs.package_name }} \
                     --rev="$BASE,$HEAD" \
                     --url=${{ github.event.repository.clone_url }} \
-                    --bench-on="$BASE" \
+                    --bench-on="$HEAD" \
                     --output-dir=results/
             - name: Create plots from benchmarks
               run: |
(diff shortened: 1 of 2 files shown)


datejada (approved):
LGTM

PR #591, opened 2024-04-16
Add balance sense option to the consumer assets

### LINKED ISSUE #585: Add an input parameter for `consumer` assets to choose the balance sense
### What and Why

In some cases, we want to impose a balance $\geq$ instead of $==$, e.g., heat demand or water to the ground (spillage to ground).

The initial idea to tackle this issue:
1. Create a new parameter in the assets-data.csv
2. Change the code to read the new parameter
3. Modify the function `add_consumer_constraints!` to change the sense of the constraint according to this new parameter (review JuMP documentation to see the best way to do this)

### Possible Drawbacks

More abstract code

### Related Issues

_No response_

datejada:
# Pull request details

Add the option to either have "==" or ">=" for the consumer balance

Closes #585 

## Collaboration confirmation

As a contributor I confirm

-   [ ] Tests are passing
-   [ ] Lint is passing


datejada:
@abelsiqueira the commits in the PR try to be as atomic as possible. I came up with this solution to get what we want to have in the user balance constraint, but it might be a better way. So, any comment or suggestion is more than welcomed 😄 

In addition, Julia v1.6 is failing, and I can't identify why. It might be that the functions that I'm using are not available for that version of Julia. But how can I know which function has backwards compatibility? Also, any hint is welcomed 

Thanks!

abelsiqueira (changes requested):
Thanks for the PR.
I've made a few suggestion to avoid changing the type of the consumer_balance_sence field. The check for nothing and ">=" is not at construction time, then.
I'm not really a fan of having the MathOptInterface objects like that, but it might be best. We'll see how it behaves when we use DataFrames throughout. A "pure math" alternative would be changing the constraint to 0 <= con <= U, and then U is 0 or Inf. There is no need to do that change now, but I'm writing it down to remember in the future.
I haven't checked the 1.6 failure yet.

datejada:
@abelsiqueira thanks for the review. I knew there was a better way 😄

abelsiqueira:
At the moment, the 1.6 failure has been identified as a GLPK failure. GLPK becomes infeasible for the norse data.
I imagine it is related to the parameters, but I can't figure out which combination to make it work.
Not sure how to proceed yet.

datejada:
> At the moment, the 1.6 failure has been identified as a GLPK failure. GLPK becomes infeasible for the norse data. I imagine it is related to the parameters, but I can't figure out which combination to make it work. Not sure how to proceed yet.

The weird thing is that with Julia, v1.10.2 is feasible, and v1.6.7 is infeasible for GLPK. So, it seems more like a bug in GLPK. I'm not sure how to proceed either. I'll try to get all the details and maybe post it as an issue in the GLPK.jl repo.

abelsiqueira:
I imagine that changing `a - b == c` to `a - b - c in K` was enough to break GLPK due to numerical issues.

datejada:
> I imagine that changing `a - b == c` to `a - b - c in K` was enough to break GLPK due to numerical issues.

I fixed it by adding the presolve option to GLPK; then it deals with the numerical issue and the test pass in Julia v1.6.7 🤷 

In addition:

1. I'll create a new issue and PR to change the test in the Norse. We really don't care about the optimal value, so we can change it to only check optimality in that case, which is what we actually care about. 
2. I had to update a link to arxiv in the docs that was broken, so the Lint tests pass

abelsiqueira (approved):
Thanks!

Issue #611, opened 2024-04-24
Change construct_dataframes to use tables (a.k.a. DuckDB) or input data frames


abelsiqueira:

abelsiqueira:
Closed by #928

PR #618, opened 2024-04-30
Update filter usage to be more efficient

abelsiqueira:
# Pull request details

This improves the filter function based on investigations on performance and online comments.
A summary seems to be:

- Use `filter` in the DataFrames syntax (`filter(:column => action, df)`) instead of `subset` (might not be true always).
- When possible, use the "implicit" functions  (don't know actual name), i.e., instead of `:column => x -> x == value`, use `:column => ==(value)`. The `==(value)` generates a function that checks for equality to `value`.
- Use `view=true`, when possible. This we knew.

Closes #ISSUE_NUMBER

## Collaboration confirmation

As a contributor I confirm

-   [x] Tests are passing
-   [x] Lint is passing

### DIFF
--- a/src/constraints/consumer.jl
+++ b/src/constraints/consumer.jl
@@ -22,7 +22,7 @@ function add_consumer_constraints!(
 )
 
     # - Balance constraint (using the lowest temporal resolution)
-    df = filter(row -> row.asset ∈ Ac, dataframes[:highest_in_out]; view = true)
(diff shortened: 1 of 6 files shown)


abelsiqueira:
Small improvement in time, but decent improvement in memory. Noticed as part of #611, so had to do it separately.

datejada (approved):
The benchmark looks really good. Thanks for the changes and improvement in the code!

PR #666, opened 2024-06-26
Standardize the names of representative periods to rep_period

### LINKED ISSUE #661: Rename `rep_periods_data.id` to `rep_periods_data.rep_period` for consistency
### What and Why

We don't have `id` for anything else, instead we use the value of the thing as identifier, so it makes sense to change this to `rep_period`. It is also used as `rep_period` or `rp` in other places.

### Possible Drawbacks

_No response_

### Related Issues

_No response_

datejada:
# Pull request details

Unify the name of representative periods to rep_period for consistency

Closes #661 

## Collaboration confirmation

As a contributor I confirm

-   [x] Tests are passing
-   [x] Lint is passing

### DIFF
--- a/benchmark/EU/rep-periods-data.csv
+++ b/benchmark/EU/rep-periods-data.csv
@@ -1,3 +1,3 @@
 ,hours,,hours
-id,weight,num_timesteps,resolution
+rep_period,weight,num_timesteps,resolution
 1,1,8760,1.0
--- a/docs/src/concepts.md
+++ b/docs/src/concepts.md
@@ -360,8 +360,8 @@ The `is_seasonal` parameter determines whether or not the storage asset uses the
 The `rep-periods-data` file has information on the representative periods in the example. We have three representative periods, each with 24 timesteps and hourly resolution, representing a day. The figure below shows the availability profile of the renewable energy sources in the example.
 
 ```@example seasonal-storage
-rp_file = "../../test/inputs/Storage/rep-periods-data.csv" # hide
-rp = CSV.read(rp_file, DataFrame, header = 2) # hide
+rep_periods_file = "../../test/inputs/Storage/rep-periods-data.csv" # hide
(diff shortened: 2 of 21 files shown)


abelsiqueira:
Oh, I think I was not clear in #661. I think `rp` is a fine shortcut for `rep_period`, when it makes sense, `id` is not.
I think in many cases is fine to leave `rp`, like we have `a` for `asset` and `(u, v)` for the flow.
I think in some cases it does make sense to change `rp` to `rep_period` (or `rep_periods`), but I don't know if I would do it for all.
For instance, I think all official names should be `rep_period`(s). Column name, field name, etc.
But variable iterating over `1:number_of_rep_periods` could be `rp`.
Maybe iterating over the `Vector` of `RepPeriods` should be `rep_period`, to disambiguate?

What do you think? Do you want to change all `rp` to `rep_period`? Are you thinking of something similar for `a` to `asset`? I'm not opposed, just wondering.

datejada:
> Oh, I think I was not clear in #661. I think `rp` is a fine shortcut for `rep_period`, when it makes sense, `id` is not. I think in many cases is fine to leave `rp`, like we have `a` for `asset` and `(u, v)` for the flow. I think in some cases it does make sense to change `rp` to `rep_period` (or `rep_periods`), but I don't know if I would do it for all. For instance, I think all official names should be `rep_period`(s). Column name, field name, etc. But variable iterating over `1:number_of_rep_periods` could be `rp`. Maybe iterating over the `Vector` of `RepPeriods` should be `rep_period`, to disambiguate?
> 
> What do you think? Do you want to change all `rp` to `rep_period`? Are you thinking of something similar for `a` to `asset`? I'm not opposed, just wondering.

Thanks for the comment. I misunderstood the final part of the issue #661. But it is clear now. So, I've updated the PR to change to rep_period for the columns, but use rp when we are iterating. Similar to assets in the code.

abelsiqueira (approved):
Looks great, thanks

PR #670, opened 2024-06-26
Move timing from run scenario

### LINKED ISSUE #667: Move timing from run_scenario to EnergyProblem function
### What and Why

The timing done in run_scenario is kept inside an EnergyProblem.
And the timings are of EnergyProblem functions.
So it would make sense to move the timing to the functions themselves, and make run_scenario shorter. This would also allows direct usage of the functions to have the timings.

### Possible Drawbacks

_No response_

### Related Issues

_No response_

datejada:
# Pull request details

Reallocate timing out of the scenario function

Closes #667 

## Collaboration confirmation

As a contributor I confirm

-   [x] Tests are passing
-   [x] Lint is passing

### DIFF
--- a/src/create-model.jl
+++ b/src/create-model.jl
@@ -349,25 +349,30 @@ end
 Create the internal model of an [`TulipaEnergyModel.EnergyProblem`](@ref).
 """
 function create_model!(energy_problem; kwargs...)
-    graph = energy_problem.graph
-    representative_periods = energy_problem.representative_periods
-    constraints_partitions = energy_problem.constraints_partitions
-    timeframe = energy_problem.timeframe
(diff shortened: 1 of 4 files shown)


abelsiqueira (approved):
Nice. I was gonna say that maybe we should try to move `to` to inside `EnergyProblem`, but it is used independently in other places as well, so maybe that won't be easy or good.
One thing that I noticed is that we can use the timer from `to` itself directly, like this:

```Julia
time_in_seconds = TimerOutputs.time(to["time to create blah"]) / 1e9
```

So maybe we could use only `to` and no `@elapsed`, but keep `to` as a global variable so that it is used in other places. At this point it doesn't seem really relevant what is the best, so let's keep it like this.

PR #675, opened 2024-06-27
Allow timeframe and partitions tables to be missing

### LINKED ISSUE #674: Accept default of missing timeframe and partition tables
### What and Why

As discussed in #664, when the timeframe and partitions tables are not passed, we should use the defaults of "no timeframe profiles" and "hourly resolution".

### Possible Drawbacks

_No response_

### Related Issues

_No response_

abelsiqueira:
These tables are already  permitted to be empty. Not they
are permitted to not exist at all.

Closes #674


datejada (approved):
LGTM

PR #677, opened 2024-06-27
Release v0.9.0

abelsiqueira:
Release will be necessary to have a standalone pipeline.
We might need another release for official M2 closure, though, so I don't know what you might find best.


datejada (approved):
Let's do it now. And depending on the changes up to Monday, we can create a new minor release version. Thanks!

PR #679, opened 2024-06-27
Revert "Release v0.9.0"

datejada:
Reverts TulipaEnergy/TulipaEnergyModel.jl#677

To create the release after the latest fixes.


abelsiqueira:
We can keep it like that, as long as we remember that we don't need to update from 0.9.0 to 0.10.0. I hope to release it today or tomorrow, if you agree. I'll give you a heads up when I'm ready.

PR #681, opened 2024-06-27
Create timing dict for EnergyProblem and time constructor

### LINKED ISSUE #680: Time creating dataframes et al. separately from reading CSV
### What and Why

The `create_energy_problem_from_csv_folder` function times two calls:

- `create_connection...` which reads the CSV files into the connection
- `EnergyProblem(connection)` which reads the dataframes input, creates internal structures and compute constraints partitions.

Most timing should happen inside `EnergyProblem(connection)`, then it also shows when called directly.

### Possible Drawbacks

_No response_

### Related Issues

_No response_

abelsiqueira:
To avoid changing the EnergyProblem structure too much because of timing, we create a timings dictionary.
We time the constructor of EnergyProblem instead of the function that reads the CSV folder. This way we time the functions inside the constructor, which are the expensive functions in the input.
This also allows EnergyProblem constructed directly to have times.

Closes #680

### DIFF
--- a/src/create-model.jl
+++ b/src/create-model.jl
@@ -368,7 +368,7 @@ function create_model!(energy_problem; kwargs...)
         energy_problem.objective_value = NaN
     end
 
-    energy_problem.time_create_model = elapsed_time_create_model
+    energy_problem.timings["creating the model"] = elapsed_time_create_model
 
     return energy_problem
 end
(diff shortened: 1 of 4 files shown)


abelsiqueira:
Hum, maybe this won't work as I wanted because of nested timing. Will think more about it.
In essence, if I call a `@timeit` inside another `@timeit`, then they will have a nested structure. So it is different if I call `run_scenario` or `EnergyProblem` directly. Might drop this.

datejada (approved):
LGTM

Issue #526, opened 2024-03-06
Create new variables and constraints to have a max/min energy outgoing flow for assets, using the base periods


datejada:
### Description

**This epic issue is part of the different levels of detail in the operation module for M3.**

The base periods allow for inter-representative period constraints since they recover chronological information, e.g., having maximum/minimum total energy flow for other types of assets (e.g., producers).

**Example: A gas contract with a max/min monthly limit**: if we have base periods 1...365, i.e., for each day of the year, then with a partition that represents a monthly time block (i.e., 1x31+1x28+1x31+1x30...), then we can set maximum/minimum limits for the monthly gas flow.

For this issue you can use as a reference the existing `storage_level_inter_rp` variable, constraints, and dataframes, since their temporal structure should be similar.

### Sub issues for assets with outgoing flows (i.e., producers, conversion, and storage)

- [x] #655
- [x] #657
  - Create the dataframe from the partition definitions for the variables and constraints. If there is no defined profile, there should be no constraints. We should apply this type of constraint if we have the data; otherwise, they are superfluous.
  - ~Create the energy variables e.g,. `asset_outgoing_energy_flow_inter_rp` (or a better name, and also maybe for the incoming energy flow, depending on the asset)~ No need of extra variables here
  - Create the constraints for the maximum and minimum energy limit
  - Update the outputs of the model
- [x] #659 
- [x] #662
  - Example
  - How-to
  - Mathematical documentation
- [x] #652

Add anything else that you can think of during the development of the epic...happy coding! 😄

clizbe:
For my own clarification, this is for operational considerations in the investment decision.
For instance, setting a min production of green hydrogen for the total year.

PR #691, opened 2024-07-08
Update formulation and how-to-use with the max/min outgoing energy limit

### LINKED ISSUE #662: Update documentation with max/min energy outgoing flow for assets
With the new feature, we need a general update on the documentation. Please double-check:
- Example
- How-to
- Mathematical documentation
- Concepts/Features

datejada:
# Pull request details

Update documentation with the latest changes in epic #526 

Closes #662 

## Collaboration confirmation

As a contributor I confirm

-   [x] Tests are passing
-   [x] Lint is passing


docs/src/formulation.md
```diff
@@ -424,6 +408,28 @@ v^{\text{inv}}_{f} \leq \frac{p^{\text{inv limit}}_{f}}{p^{\text{capacity}}_{f}}
 
 If the parameter `investment_integer` in the [`flows-data.csv`](@ref flows-data) file is set to true, then the right-hand side of this constraint uses a least integer function (floor function) to guarantee that the limit is integer.
```
clizbe, line 409: I'm a bit worried this also crosses the line between Explanation and Reference/How-To.

docs/src/formulation.md
```diff
@@ -18,12 +18,14 @@ The complete mathematical formulation, including variable temporal resolutions,
...
+| $\mathcal{A}^{\text{ss}}$    | Storage energy assets with seasonal method        |          | $\mathcal{A}^{\text{ss}} \subseteq \mathcal{A}^{\text{s}}$                                   |                                                                                                                                                                                                                                      |
+| $\mathcal{A}^{\text{se}}$    | Storage energy assets with energy method          |          | $\mathcal{A}^{\text{se}} \subseteq \mathcal{A}^{\text{s}}$                                   | This set contains storage assets that use investment energy method (`storage_method_energy = true` in the file [`assets-data.csv`](@ref schemas)). Otherwise fixed energy-to-power ratio method is used.                             |
```
clizbe, line 25: These notes about the code are crossing the line between Explanation (read in the bath) and Reference/How-To (look something up while implementing).  Can we do something like put a link here and explain this in a Reference/How-To section?

clizbe: But if you want to update and publish you can leave it for now and open an issue to fix it. :)

docs/src/how-to-use.md
```diff
@@ -337,3 +337,21 @@ Depending on the configuration of the energy storage assets, it may or may not b
 If no value is set in the parameter `use_binary_storage_method`, i.e., `missing` value, the storage asset can charge and discharge simultaneously.
 
 For more details on the constraints that apply when selecting this method, please visit the [`mathematical formulation`](@ref formulation) section.
+
+## [Setup a maximum or minimum outgoing energy limit](@id max-min-outgoing-energy-setup)
+
+To enable the model to utilize the constraints allowing us to consider a [maximum or minimum energy limit](@ref inter-temporal-energy-constraints) for an asset throughout the model's timeframe (e.g., a year), we need to establish a couple of parameters:
```
clizbe, line 343: Wow that's a meaty lead-in phrase.
Here's my attempt at shortening it:
```suggestion
For the model to add constraints for a [maximum or minimum energy limit](@ref inter-temporal-energy-constraints) for an asset throughout the model's timeframe (e.g., a year), we need to establish a couple of parameters:
```

If that's correct.

docs/src/how-to-use.md
```diff
@@ -337,3 +337,21 @@ Depending on the configuration of the energy storage assets, it may or may not b
...
+-   (optional) define a period partition in [`assets-timeframe-partitions.csv`](@ref schemas). If there is no partition defined, then by default the constraint is created for each period in the timeframe, otherwise, it will consider the partition definition in the file.
+
+For example, let's assume we have a year divided into 365 days because we are using days as periods in the representatives from the [_TulipaClustering.jl_](https://github.com/TulipaEnergy/TulipaClustering.jl). In addition, we define the `max_energy_timeframe_partition = 10 MWh`, meaning that the peak energy we want to have is 10MWh for each period or period partition. So, depending on the optional information we can have:
+
+-   _No definition of profile and period partitions_: The default profile is 1.p.u. for each period and since there are no period partitions then constraints will be for each period (i.e., daily). So, the outgoing energy of the asset for each day must be lower or equal to 10Mwh.
```
clizbe, line 352:
```suggestion
-   _No definition of profile and period partitions_: The default profile is 1.p.u. for each period and since there are no period partitions then constraints will be for each period (i.e., daily). So, the outgoing energy of the asset for each day must be less than or equal to 10MWh.
```

docs/src/how-to-use.md
```diff
@@ -337,3 +337,21 @@ Depending on the configuration of the energy storage assets, it may or may not b
...
+-   _Profile definition, but no definition of period partitions_: The profile definition and value will be in the [`assets-timeframe-profiles.csv`](@ref schemas) and [`profiles-timeframe.csv`](@ref schemas) files. For example, we define a profile that has the following first four values: 0.6 p.u., 1.0 p.u., 0.8 p.u., and 0.4 p.u. Since there are no period partitions then constraints will be again for each period (i.e., daily). So, the outgoing energy of the asset for the first four days must be lower or equal to 6MWh, 10MWh, 8MWh, and 4Mh.
```
clizbe, line 353:
```suggestion
-   _Profile definition, but no definition of period partitions_: The profile definition and value will be in the [`assets-timeframe-profiles.csv`](@ref schemas) and [`profiles-timeframe.csv`](@ref schemas) files. For example, we define a profile that has the following first four values: 0.6 p.u., 1.0 p.u., 0.8 p.u., and 0.4 p.u. There are no period partitions, so constraints will be for each period (i.e., daily). Therefore the outgoing energy of the asset for the first four days must be less than or equal to 6MWh, 10MWh, 8MWh, and 4MWh, respectively.
```

docs/src/how-to-use.md
```diff
@@ -337,3 +337,21 @@ Depending on the configuration of the energy storage assets, it may or may not b
...
+-   _Profile and period partitions definition_: Same profile as before, but now we define a period partition in the [`assets-timeframe-partitions.csv`](@ref schemas) file as `uniform` with a value of 2. This value means that we will aggregate every two periods (i.e., every two days). So, instead of having 365 constraints, we will have 183 constraints (182 every two days and one last constraint of 1 day). Then the profile is aggregated with the sum of the values inside the periods within the partition. Thus, the outgoing energy of the asset for the first two partitions (i.e., every two days) must be lower or equal to 16MWh, and 12MWh.
```
clizbe, line 354:
```suggestion
-   _Profile and period partitions definition_: Using the same profile as above, we now define a period partition in the [`assets-timeframe-partitions.csv`](@ref schemas) file as `uniform` with a value of 2. This value means that we will aggregate every two periods (i.e., every two days). So, instead of having 365 constraints, we will have 183 constraints (182 every two days and one last constraint of 1 day). Then the profile is aggregated with the sum of the values inside the periods within the partition. Thus, the outgoing energy of the asset for the first two partitions (i.e., every two days) must be less than or equal to 16MWh and 12MWh, respectively.
```

docs/src/how-to-use.md
```diff
@@ -337,3 +337,21 @@ Depending on the configuration of the energy storage assets, it may or may not b
...
+
+> **Note:**
+> If you want to set a limit on the maximum or minimum outgoing energy for a year with representative days, you can use the partition definition to create a single partition for the entire year to combine the profile. If you are using the default value for the profiles, which is 1.0 p.u., please note that the parameters `max_energy_timeframe_partition` and `min_energy_timeframe_partition` are defined per period, which means they represent daily limits. This implies that the energy limit for the whole year will be 365 times `max_energy_timeframe_partition` or 365 times `min_energy_timeframe_partition`.
```
clizbe, line 357: This is two different notes, right? I would split it, so people don't skip the second note if the first doesn't apply to them..

clizbe, line 357:
```suggestion
> If you want to set a limit on the maximum or minimum outgoing energy for a year with representative days, you can use the partition definition to create a single partition for the entire year. If you are using the default value for the profiles, which is 1.0 p.u., please note that the parameters `max_energy_timeframe_partition` and `min_energy_timeframe_partition` are defined per period, which means they represent daily limits. This implies that the energy limit for the whole year will be 365 times `max_energy_timeframe_partition` or 365 times `min_energy_timeframe_partition`.
```

Don't think that phrase is needed.

clizbe (changes requested):
Looks pretty good! Nice you're adding to the docs. I just wonder about including coding in the Explanation section... But if you want to release we can fix it later.

datejada:
Thanks! Yes, we can have later a general review of the docs and create an epic issue with actions to add/remove/fix stuff

docs/src/how-to-use.md
datejada: It makes sense, thanks! :)

datejada: Correct! I will split them. Thanks!

docs/src/formulation.md
datejada: I have changed it to reference the how-to section. I think we should centralize all these parameters setup in that section. That might be one the action points in the full review of the documentation.

docs/src/formulation.md
datejada: We don't have a how-to section for this, so we are leaving this here like this for now. Let's improve it later in the whole review of the documentation. Thanks!

PR #711, opened 2024-07-31
Change Symbols to Strings for input data

### LINKED ISSUE #705: Use strings instead of Symbols in the create_input_dataframes function
(this means Tulipa will now expect strings instead of Symbols)

datejada:
# Pull request details

To use DuckDB types

Closes #705 

## Collaboration confirmation

As a contributor I confirm

-   [ ] Tests are passing
-   [x] Lint is passing

### DIFF
--- a/test/test-io.jl
+++ b/test/test-io.jl
@@ -135,7 +135,7 @@ end
             println(io, line)
         end
     end
-    missing_asset = Symbol(split(lines[end], ",")[1]) # The asset the was not included
+    missing_asset = split(lines[end], ",")[1] # The asset the was not included
 
     connection = DBInterface.connect(DuckDB.DB)
     read_csv_folder(connection, dir)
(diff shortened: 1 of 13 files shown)


test/test-io.jl
```diff
@@ -135,7 +135,7 @@ end
             println(io, line)
         end
     end
-    missing_asset = Symbol(split(lines[end], ",")[1]) # The asset the was not included
+    missing_asset = split(lines[end], ",")[1] # The asset the was not included
```
abelsiqueira, line 138:
```suggestion
    missing_asset = split(lines[end], ",")[1] # The asset that was not included
```

abelsiqueira (approved):
Found a small fix in the comment, but it's otherwise good to go.

datejada:
@abelsiqueira I have mixed feelings with the benchmark result, any comments?

abelsiqueira:
Oof, don't like it. The slowdown is too large for something that we thought would be ~0% change. I will create a revert PR so we can see if there is a similar increase or if the benchmarks are broken. If there is a speed up, I think we might need to investigate before some other changes.

PR #721, opened 2024-08-04
Create profiling folder with instructions

abelsiqueira:
# Pull request details

## Describe the changes made in this pull request

Creates more details on profiling. All information is in a separate folder for now, but it can be moved to the developer documentation and merged with the current information.

## List of related issues or pull requests

No related issues

## Collaboration confirmation

As a contributor I confirm

-   [x] I read and followed the instructions in README.dev.md
-   [x] The documentation is up to date with the changes introduced in this Pull Request (or NA)
-   [x] Tests are passing
-   [x] Lint is passing

PR #733, opened 2024-08-09
Remove table tree structure

### LINKED ISSUE #729: Remove table_tree (the end)

datejada:
# Pull request details

Remove the table_tree structure to avoid creating input dataframes and use DuckDB tables instead.

Closes #729 

## Collaboration confirmation

As a contributor I confirm

-   [ ] Tests are passing
-   [ ] Lint is passing

### DIFF
--- a/src/io.jl
+++ b/src/io.jl
@@ -2,7 +2,7 @@ export create_internal_structures,
     save_solution_to_file, compute_assets_partitions!, compute_flows_partitions!
 
 """
-    graph, representative_periods, timeframe  = create_internal_structures(table_tree)
+    graph, representative_periods, timeframe  = create_internal_structures(connection)
 
 Return the `graph`, `representative_periods`, and `timeframe` structures given the input dataframes structure.
 
--- a/src/io.jl
+++ b/src/io.jl
@@ -1,103 +1,5 @@
-export create_input_dataframes,
-    create_internal_structures,
-    save_solution_to_file,
-    compute_assets_partitions!,
-    compute_flows_partitions!
-
-"""
-    table_tree = create_input_dataframes(connection)
-
-Returns the `table_tree::TableTree` structure that holds all data using a DB `connection` that
-has loaded all the relevant tables.
-Set `strict = true` to error if assets are missing from partition data.
-
-The following tables are expected to exist in the DB.
-
-> !!! warn
->
-> The schemas are currently being ignored, see issue
-[#636](https://github.com/TulipaEnergy/TulipaEnergyModel.jl/issues/636) for more information.
-
-  _ `assets_timeframe_partitions`: Following the schema `schemas.assets.timeframe_partition`.
-  _ `assets_data`: Following the schema `schemas.assets.data`.
-  _ `assets_timeframe_profiles`: Following the schema `schemas.assets.profiles_reference`.
-  _ `assets_profiles`: Following the schema `schemas.assets.profiles_reference`.
-  _ `assets_rep_periods_partitions`: Following the schema `schemas.assets.rep_periods_partition`.
-  _ `flows_data`: Following the schema `schemas.flows.data`.
-  _ `flows_profiles`: Following the schema `schemas.flows.profiles_reference`.
-  _ `flows_rep_periods_partitions`: Following the schema `schemas.flows.rep_periods_partition`.
-  _ `profiles_timeframe`: Following the schema `schemas.timeframe.profiles_data`.
-  _ `profiles_rep_periods`: Following the schema `schemas.rep_periods.profiles_data`.
-  _ `rep_periods_data`: Following the schema `schemas.rep_periods.data`.
-  _ `rep_periods_mapping`: Following the schema `schemas.rep_periods.mapping`.
-"""
-function create_input_dataframes(connection::DuckDB.DB; strict = false)
-    function read_table(table_name; allow_missing_table = false)
-        schema = get_schema(table_name)
-        if allow_missing_table
-            existence_query = DBInterface.execute(
-                connection,
-                "SELECT table_name FROM information_schema.tables WHERE table_name = '$table_name'",
-            )
-            if length(collect(existence_query)) == 0
-                return DataFrame([key => value[] for (key, value) in schema]...)
-            end
-        end
-        df = DataFrame(DBInterface.execute(connection, "SELECT * FROM $table_name"))
-        # enforcing schema to match what Tulipa expects; int -> string
-        for (key, value) in schema
-            if value <: Union{Missing,String} && !(eltype(df[!, key]) <: Union{Missing,String})
-                df[!, key] = [ismissing(x) ? x : string(x) for x in df[!, key]]
-            end
-        end
-        return df
-    end
-    df_assets_data = read_table("assets_data")
-    df_flows_data  = read_table("flows_data")
-    df_rep_periods = read_table("rep_periods_data")
-    df_rp_mapping  = read_table("rep_periods_mapping")
-
-    dfs_assets_profiles = Dict(
-        :rep_periods => read_table("assets_profiles"),
-        :timeframe => read_table("assets_timeframe_profiles"; allow_missing_table = true),
-    )
-    df_flows_profiles = read_table("flows_profiles")
-    dfs_assets_partitions = Dict(
-        period_type =>
-            read_table("assets_$(period_type)_partitions"; allow_missing_table = true) for
-        period_type in PERIOD_TYPES
-    )
-    df_flows_partitions = read_table("flows_rep_periods_partitions"; allow_missing_table = true)
-
-    dfs_profiles = Dict(
-        :rep_periods => read_table("profiles_rep_periods"),
-        :timeframe => read_table("profiles_timeframe"; allow_missing_table = true),
-    )
-
-    # Error if partition data is missing assets (if strict)
-    if strict
-        missing_assets =
-            setdiff(df_assets_data[!, :name], dfs_assets_partitions[:rep_periods][!, :asset])
-        if length(missing_assets) > 0
-            msg = "Error: Partition data missing for these assets: \n"
-            for a in missing_assets
-                msg *= "- $a\n"
-            end
-            msg *= "To assume missing asset resolutions follow the representative period's time resolution, set strict = false.\n"
-
-            error(msg)
-        end
-    end
-
-    table_tree = TableTree(
-        (assets = df_assets_data, flows = df_flows_data),
-        (assets = dfs_assets_profiles, flows = df_flows_profiles, data = dfs_profiles),
-        (assets = dfs_assets_partitions, flows = df_flows_partitions),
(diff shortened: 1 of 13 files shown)


src/io.jl
```diff
@@ -217,41 +131,41 @@ function create_internal_structures(table_tree::TableTree, connection)
...
-                    )
-                    found = true
-                    break
-                end
-            end
-            if !found
-                graph[row.name].timeframe_partitions =
-                    _parse_rp_partition(Val(:uniform), "1", 1:timeframe.num_periods)
-            end
-        end
```
abelsiqueira, line 247: I think we discussed rewriting this with TulipaIO later, right @suvayu?
I'll leave this comment so that we create an issue for it.

abelsiqueira (changes requested):
Thanks for the PR. I've left a comment/reminder in the body, but also we need to update the Project.toml file with the new release of TulipaIO, to prevent installing the old TulipaIO. It's not a huge concern, but it's good just in case.

One other thing that I noticed now is that the `schema_per_file` should perhaps be renamed to something like `schema_per_table_name`.

Finally, since the name is quite verbose, maybe we can add a helper function in the tests like:

```Julia
# In a file aux.jl, for instance?
function _read_csv_folder(con, dir)
    schemas = TEM.schema...
    return TulipaIO.read_csv_folder(con, dir; schemas)
end
```
Not necessary, though.

The other changes are approved, but I'll mark as "Request changes" for the changes above.

abelsiqueira:
I think I added `ProfileView` in PR #721 by mistake. @datejada, can you remove it?

datejada:
> I think I added `ProfileView` in PR #721 by mistake. @datejada, can you remove it? Edit: In the benchmark Project.toml

Sure, I think that's why the benchmarks are failing. Thanks! I will remove it

datejada:
> Thanks for the PR. I've left a comment/reminder in the body, but also we need to update the Project.toml file with the new release of TulipaIO, to prevent installing the old TulipaIO. It's not a huge concern, but it's good just in case.
> 
> One other thing that I noticed now is that the `schema_per_file` should perhaps be renamed to something like `schema_per_table_name`.
> 
> Finally, since the name is quite verbose, maybe we can add a helper function in the tests like:
> 
> ```julia
> # In a file aux.jl, for instance?
> function _read_csv_folder(con, dir)
>     schemas = TEM.schema...
>     return TulipaIO.read_csv_folder(con, dir; schemas)
> end
> ```
> 
> Not necessary, though.
> 
> The other changes are approved, but I'll mark as "Request changes" for the changes above.

Thanks for the suggestions. Please take a look at the latest changes to double-check that they are according to your comment 😄

datejada:
@abelsiqueira, the benchmark, gets an error due to the change in the name from `TulipaEnergyModel.schema_per_file` to `TulipaEnergyModel.schema_per_table_name`.
So, should we go back to schema_per_file in this PR and then change the name to another PR only with that (which won't be benchmarked)?

abelsiqueira:
Hum, I think the benchmarks are failing now because the API has changed. It tries to run the same script in both commits, and the old commit doesn't have the `schema_per_table_name` structure.

Short solution is accept that no benchmarks will be run. But if we want the benchmarks, the solution is to create a separate PR just changing the name of the structure and adding it to the script. That's my best guess at the moment. What do you think?

abelsiqueira:
> @abelsiqueira, the benchmark, gets an error due to the change in the name from `TulipaEnergyModel.schema_per_file` to `TulipaEnergyModel.schema_per_table_name`. So, should we go back to schema_per_file in this PR and then change the name to another PR only with that (which won't be benchmarked)?

Even better! 🥷

datejada:
@abelsiqueira, the change with the name works, but when reading the file for the EU, the schema fails (some type of conversion, that I don't see what is happening). I tried locally and got also an error, but if I run the benchmark without the schema (it does not fail the type thing because I skip it), it works for the creation of the internal structures. So, do you know how to specifically see which file has a problem? I was not able to find it. The second option will be to run the benchmark without the schema :/

abelsiqueira:
Hum, not out the top of my head. I would try reading them one by one, maybe. Let me try for a bit.

abelsiqueira:
Ah, actually the problem is deeper. The `schema_per_file` changes between `main` and this branch, so the benchmark is trying to use the old `schema_per_file` for the old commit.
We will need to change the `schema_per_file` first, seems like.

abelsiqueira:
That's why it was complaining about type "Float64", it was looking at the old schema.
PS. I restarted the benchmark to see if it was a intermittent problem.

src/io.jl
```diff
@@ -301,6 +215,21 @@ function get_schema(tablename)
...
 
+function _check_if_table_exist(connection, table_name)
+    schema = get_schema(table_name)
+
+    existence_query = DBInterface.execute(
+        connection,
+        "SELECT table_name FROM information_schema.tables WHERE table_name = '$table_name'",
+    )
+    if length(collect(existence_query)) == 0
+        columns_in_table = join(("$key $value" for (key, value) in schema), ",")
```
suvayu, line 226:
```suggestion
        columns_in_table = join(("$col $col_type" for (col, col_type) in schema), ",")
```

datejada:
> That's why it was complaining about type "Float64", it was looking at the old schema. PS. I restarted the benchmark to see if it was a intermittent problem.

ok! I disable the schema checking in the benchmark to continue with this PR. PR #733 will put back the schema checking since that PR is not going to be benchmarked. does that sound good?

src/io.jl
suvayu: Yes, this and the `groupby` & `filter`s below.

abelsiqueira:
Yes, that's a great resolution.

abelsiqueira (approved):
Thanks! Hopefully the benchmarks will be positive.

test/test-aux.jl
```diff
@@ -0,0 +1,6 @@
+# file with auxiliary functions for the testing
+
+function _read_csv_folder(connection, input_dir)
+    schemas = TulipaEnergyModel.schema_per_file
+    return TulipaIO.read_csv_folder(connection, input_dir; schemas)
+end
```
suvayu, line 6:
```suggestion
import TulipaIO

function read_csv_folder(connection, input_dir)
    schemas = TulipaEnergyModel.schema_per_file
    return TulipaIO.read_csv_folder(connection, input_dir; schemas)
end
```

I think if you do this you don't need to do the renames in `test-case-studies.jl`.  I'm a little unsure about the namespace, @abelsiqueira should be able to confirm.

You can also ignore this suggestion since it's not particularly important.

src/io.jl
```diff
@@ -217,41 +131,41 @@ function create_internal_structures(table_tree::TableTree, connection)
...
+    #=
+    For timeframe, This SQL query retrieves the names of assets from the `assets_data` table
+    along with their corresponding partition specifications from the `assets_timeframe_partitions` table,
+    if they exist. If a specification or partition is not available, it defaults to 'uniform' and '1' respectively.
+    The query only includes assets marked as seasonal (`is_seasonal` column) in the `assets_data` table.
+    =#
+    find_assets_partitions_query = """
+         SELECT assets_data.name,
+                 IFNULL(assets_timeframe_partitions.specification, 'uniform') AS specification,
+                 IFNULL(assets_timeframe_partitions.partition, '1') AS partition
```
suvayu, line 150:
```suggestion
                 IFNULL(assets_timeframe_partitions.partition, 1) AS partition
```

suvayu (changes requested):

test/test-aux.jl
abelsiqueira: That's true, I didn't think of that. It should work if `read_csv_folder` is not exported by TulipaIO and we don't  explicitly import it from TulipaIO.

abelsiqueira:
Ok, the benchmark failure now is in the current branch, i.e., parsing without the schema fails.
I think the solution will have to be making the schema change first.
@datejada, I can work on that.

datejada:
> Ok, the benchmark failure now is in the current branch, i.e., parsing without the schema fails. I think the solution will have to be making the schema change first. @datejada, I can work on that.

Yes, I see. It is better to have the schema in the main branch and then rebase this PR. So, thanks for working on that.

test/test-aux.jl
datejada: @suvayu, actually, I prefer renaming and having "_" at the beginning to make it clear that it is an internal function, and that is different from the TulipaIO one. It is clearer for no Julia experts/programmers because you don't need to think about the situation that Abel is describing.

datejada:
@suvayu I had to revert your proposed change here because actually, we expect a STRING, because the partitions can be also text (e.g., "1x4+5x7"). Later in the code there is a function that converts each case accordingly. 

```julia
IFNULL(assets_timeframe_partitions.partition, '1') AS partition
```

test/test-aux.jl
suvayu: As I said, you can ignore my comment if you prefer.

suvayu (approved):

datejada:
> ### Benchmark Results
> [5810955](https://github.com/TulipaEnergy/TulipaEnergyModel.jl/commit/5810955ee39bf23b56ae93fe387a743acdcca8b8)...	[3284786](https://github.com/TulipaEnergy/TulipaEnergyModel.jl/commit/3284786a7dbcf4b70a4275009cd1ec23677ecf0a)...	[5810955](https://github.com/TulipaEnergy/TulipaEnergyModel.jl/commit/5810955ee39bf23b56ae93fe387a743acdcca8b8).../3284786a7dbcf4...
> energy_problem/create_model	0.131 ± 0.0048 h	0.129 ± 0.0066 h	1.01
> energy_problem/input_and_constructor	15.8 ± 0.21 s	15.1 ± 0.13 s	1.04
> time_to_load	2.93 ± 0.0084 s	2.9 ± 0.045 s	1.01
> [5810955](https://github.com/TulipaEnergy/TulipaEnergyModel.jl/commit/5810955ee39bf23b56ae93fe387a743acdcca8b8)...	[3284786](https://github.com/TulipaEnergy/TulipaEnergyModel.jl/commit/3284786a7dbcf4b70a4275009cd1ec23677ecf0a)...	[5810955](https://github.com/TulipaEnergy/TulipaEnergyModel.jl/commit/5810955ee39bf23b56ae93fe387a743acdcca8b8).../3284786a7dbcf4...
> energy_problem/create_model	0.794 G allocs: 32.9 GB	0.794 G allocs: 32.9 GB	1
> energy_problem/input_and_constructor	6.41 M allocs: 0.651 GB	6.4 M allocs: 0.65 GB	1
> time_to_load	0.153 k allocs: 14.5 kB	0.153 k allocs: 14.5 kB	1
> ### Benchmark Plots
> A plot of the benchmark results have been uploaded as an artifact to the workflow run for this PR. Go to "Actions"->"Benchmark a pull request"->[the most recent run]->"Artifacts" (at the bottom).

@abelsiqueira 🚀

PR #741, opened 2024-08-14
Add vintage version for multi-year 

### LINKED ISSUE #740: Add the most complicated method

gnawin:
# Pull request details

This PR is huge, because it involves change the data structure and the model, which does not make sense if this PR gets separated.

Closes #740 

## Collaboration confirmation

As a contributor I confirm

-   [ ] Tests are passing
-   [ ] Lint is passing


gnawin:
@abelsiqueira The time resolution part, for assets/flows/constraints, seems to complain about the type as I mentioned earlier. 

- I have to remove the type of the first argument in `compute_rp_partition`. That removes that particular error for constraints. See line 208 in `time-resolution.jl`
- `compute_assets_partitions!` complains about the type as well for the assets in line 507 in `io.jl`.

I'm a bit lost here, maybe you could see what changes I made lead to these errors?

src/structures.jl
```diff
@@ -91,12 +94,13 @@ mutable struct GraphAssetData
         max_energy_timeframe_partition,
         min_energy_timeframe_partition,
     )
-        timeframe_profiles = Dict{String,Vector{Float64}}()
-        rep_periods_profiles = Dict{Tuple{String,Int},Vector{Float64}}()
-        timeframe_partitions = PeriodsBlock[]
-        rep_periods_partitions = Dict{Int,Vector{TimestepsBlock}}()
+        timeframe_profiles = Dict{Int,Dict{String,Vector{Float64}}}()
+        rep_periods_profiles = Dict{Int,Dict{Tuple{String,Int},Vector{Float64}}}()
+        timeframe_partitions = Dict{Int,PeriodsBlock}()
```
abelsiqueira, line 99: I haven't looked at the whole code, but I think this should be the following:
```suggestion
        timeframe_partitions = Dict{Int,PeriodsBlock[]}()
```

src/time-resolution.jl
```diff
@@ -65,19 +68,31 @@ function compute_constraints_partitions(graph, representative_periods)
...
-            end for
-            a in MetaGraphsNext.labels(graph), rp in 1:num_rep_periods if asset_filter(a)
+            end for a in MetaGraphsNext.labels(graph) if asset_filter(a) for
+            y in active_years[a] for iy in (
+                if any(graph[a].investable[iy] for iy in years)
+                    filter(iy -> graph[a].investable[iy], years)
+                else
+                    [0]
+                end
+            ) for rp in RP[y]
```
abelsiqueira, line 95: Maybe this is the issue with the type.

src/time-resolution.jl
```diff
@@ -65,19 +68,31 @@ function compute_constraints_partitions(graph, representative_periods)
...
                     Vector{TimestepsBlock}[]
                 end
-            end for
-            a in MetaGraphsNext.labels(graph), rp in 1:num_rep_periods if asset_filter(a)
+            end for a in MetaGraphsNext.labels(graph) if asset_filter(a) for
+            y in active_years[a] for iy in (
+                if any(graph[a].investable[iy] for iy in years)
+                    filter(iy -> graph[a].investable[iy], years)
+                else
+                    [0]
```
abelsiqueira, line 93:
```suggestion
                    [false]
```
Trying to translate.
If the graph is investable at any of the years, then loop it over the values of `investable` for each year.
Else, loop `yi` over `[0]`.
This is a problem because you are returning a vector of booleans in the first and a vector of integers in the second.

You can also try:

```suggestion
                    Bool[]
```

Unless you need the single value there.

Also, just try `for it in filter(iy -> graph[a].investable[iy], years)`. I think maybe it won't work (and you might have tried already).

abelsiqueira:
I made some comments on what, I think, is an issue. I also don't like the graph creation, but I have to read it more carefully to suggest something.

src/time-resolution.jl
gnawin: Actually I want to return a list of investment years, for which `iy` (investment year) can be looped over. So I'm trying to say, if the asset is investable at any of the years, then get the list of years for which the `graph[a].investable[iy]` is true.
Else, use a default list of [0], i.e., year 0.

src/create-model.jl
```diff
@@ -269,53 +331,66 @@ function add_expression_terms_inter_rp_constraints!(
...
+                        LinearAlgebra.dot(
+                            sub_df_flows.flow,
+                            sub_df_flows.duration ./ sub_df_flows.efficiency,
+                        ) * row_map.weight
+                else
+                    row_inter.outgoing_flow +=
+                        LinearAlgebra.dot(sub_df_flows.flow, sub_df_flows.duration) * row_map.weight
+                end
+
+                if !is_storage_level # got an error here, that's why I modified this condition
```
gnawin, line 365: @datejada This condition was `if is_storage_level`, which is the same as the previous one, is this intentional or a glitch?

src/constraints/storage.jl
```diff
@@ -100,36 +102,37 @@ function add_storage_constraints!(
             profile_aggregation(
                 Statistics.mean,
                 graph[row.asset].rep_periods_profiles,
+                row.year,
                 ("min_storage_level", row.rep_period),
                 row.timesteps_block,
                 0.0,
             ) * (
-                graph[row.asset].initial_storage_capacity +
-                (row.asset ∈ Ai ? energy_limit[row.asset] : 0.0)
```
gnawin, line 108: @datejada I also came across this: the same `energy_limit` is used for both max and min constraints?

gnawin (commented):
@datejada I came across a few things about storage when I was doing this PR, maybe you could double-check? Thanks

gnawin:
Replaced by #745

src/constraints/storage.jl
datejada: This is correct since we use a minimum storage level profile for the min constraint 😄

src/create-model.jl
datejada: It is not a glitch, but the function is confusing, so we will clean it in the refactor. Thanks for the feedback.

PR #744, opened 2024-08-20
Update documentation with groups information

### LINKED ISSUE #743: Update documentation with the group feature
Include in the documentation:

- [x] How-to section, how the user can define groups and their constraints.
- [x] Formulation
- [x] Others -> update structures

datejada:
# Pull request details

Update formulation and how-to sections

Closes #743 

## Collaboration confirmation

As a contributor I confirm

-   [x] Tests are passing
-   [x] Lint is passing

### DIFF
--- a/docs/src/formulation.md
+++ b/docs/src/formulation.md
@@ -50,19 +50,32 @@ In addition, the following flow sets represent methods for incorporating additio
 | $\mathcal{K}$     | Representative periods (rp)                         | $k \in \mathcal{K}$       | $\mathcal{K} \subset \mathbb{N}$ | $\mathcal{K}$ does not have to be a subset of $\mathcal{P}$                |
 | $\mathcal{B}_{k}$ | Timesteps blocks within a representative period $k$ | $b_{k} \in \mathcal{B}_k$ |                                  | $\mathcal{B}_k$ is a partition of timesteps in a representative period $k$ |
 
+### Sets for Groups
+
+| Name                     | Description             | Elements                       | Superset | Notes |
+| ------------------------ | ----------------------- | ------------------------------ | -------- | ----- |
+| $\mathcal{G}^{\text{a}}$ | Groups of energy assets | $g \in \mathcal{G}^{\text{a}}$ |          |       |
+
+In addition, the following subsets represent methods for incorporating additional constraints in the model.
+
+| Name                      | Description                                         | Elements | Superset                                                   | Notes                                                                                                                                                                          |
+| ------------------------- | --------------------------------------------------- | -------- | ---------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| $\mathcal{G}^{\text{ai}}$ | Group of assets that share min/max investment limit |          | $\mathcal{G}^{\text{ai}} \subseteq \mathcal{G}^{\text{a}}$ | This set contains a group of investment assets that use investment method. Please visit the [how-to section](@ref investment-group-setup) to learn how to set up this feature. |
+
 ## [Parameters](@id math-parameters)
 
 ### Parameter for Assets
 
 #### General Parameters for Assets
 
-| Name                                        | Domain           | Domains of Indices                                                | Description                                                                                 | Units          |
-| ------------------------------------------- | ---------------- | ----------------------------------------------------------------- | ------------------------------------------------------------------------------------------- | -------------- |
-| $p^{\text{inv cost}}_{a}$                   | $\mathbb{R}_{+}$ | $a \in \mathcal{A}$                                               | Investment cost of a unit of asset $a$                                                      | [kEUR/MW/year] |
-| $p^{\text{inv limit}}_{a}$                  | $\mathbb{R}_{+}$ | $a \in \mathcal{A}$                                               | Investment potential of asset $a$                                                           | [MW]           |
-| $p^{\text{capacity}}_{a}$                   | $\mathbb{R}_{+}$ | $a \in \mathcal{A}$                                               | Capacity per unit of asset $a$                                                              | [MW]           |
-| $p^{\text{init capacity}}_{a}$              | $\mathbb{R}_{+}$ | $a \in \mathcal{A}$                                               | Initial capacity of asset $a$                                                               | [MW]           |
-| $p^{\text{availability profile}}_{a,k,b_k}$ | $\mathbb{R}_{+}$ | $a \in \mathcal{A}$, $k \in \mathcal{K}$, $b_k \in \mathcal{B_k}$ | Availability profile of asset $a$ in the representative period $k$ and timestep block $b_k$ | [p.u.]         |
+| Name                                        | Domain                   | Domains of Indices                                                | Description                                                                                 | Units          |
+| ------------------------------------------- | ------------------------ | ----------------------------------------------------------------- | ------------------------------------------------------------------------------------------- | -------------- |
+| $p^{\text{inv cost}}_{a}$                   | $\mathbb{R}_{+}$         | $a \in \mathcal{A}$                                               | Investment cost of a unit of asset $a$                                                      | [kEUR/MW/year] |
+| $p^{\text{inv limit}}_{a}$                  | $\mathbb{R}_{+}$         | $a \in \mathcal{A}$                                               | Investment potential of asset $a$                                                           | [MW]           |
+| $p^{\text{capacity}}_{a}$                   | $\mathbb{R}_{+}$         | $a \in \mathcal{A}$                                               | Capacity per unit of asset $a$                                                              | [MW]           |
+| $p^{\text{init capacity}}_{a}$              | $\mathbb{R}_{+}$         | $a \in \mathcal{A}$                                               | Initial capacity of asset $a$                                                               | [MW]           |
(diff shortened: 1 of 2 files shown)


datejada:
@gnawin, this is the documentation for the group feature. I would appreciate your feedback since it is part of the new developments in the multi-year investment.

@clizbe, you always like to keep an eye on the docs, so if you have time, your feedback would also be appreciated.

Thank you both!

clizbe:
Does this implementation assume an asset can have only 1 group?

docs/src/formulation.md
```diff
@@ -50,19 +50,32 @@ In addition, the following flow sets represent methods for incorporating additio
...
+
+In addition, the following subsets represent methods for incorporating additional constraints in the model.
+
+| Name                      | Description                                         | Elements | Superset                                                   | Notes                                                                                                                                                                          |
+| ------------------------- | --------------------------------------------------- | -------- | ---------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| $\mathcal{G}^{\text{ai}}$ | Group of assets that share min/max investment limit |          | $\mathcal{G}^{\text{ai}} \subseteq \mathcal{G}^{\text{a}}$ | This set contains a group of investment assets that use investment method. Please visit the [how-to section](@ref investment-group-setup) to learn how to set up this feature. |
```
clizbe, line 63: "that use SOMETHING investment method"

An investment method? The same investment method?

docs/src/formulation.md
```diff
@@ -117,6 +130,13 @@ In addition, the following flow sets represent methods for incorporating additio
 | $p^{\text{rp weight}}_{k}$  | $\mathbb{R}_{+}$ | $k \in \mathcal{K}$                      | Weight of representative period $k$                            | [-]   |
 | $p^{\text{map}}_{p,k}$      | $\mathbb{R}_{+}$ | $p \in \mathcal{P}$, $k \in \mathcal{K}$ | Map with the weight of representative period $k$ in period $p$ | [-]   |
 
+### Parameter for Groups
+
+| Name                              | Domain           | Domains of Indices              | Description                                       | Units |
+| --------------------------------- | ---------------- | ------------------------------- | ------------------------------------------------- | ----- |
+| $p^{\text{min invest limit}}_{g}$ | $\mathbb{R}_{+}$ | $g \in \mathcal{G}^{\text{ai}}$ | Mininum investment limit (potential) of group $g$ | [MW]  |
```
clizbe, line 137: Minimum is misspelled

docs/src/how-to-use.md
```diff
@@ -354,11 +369,58 @@ For the model to add constraints for a [maximum or minimum energy limit](@ref in
...
 
-<!-- prettier-ignore-end -->
+## [Defining a group of assets](@id group-setup)
+
+A group of assets refers to a set of assets that share certain constraints. For example, the investments of a group of assets may be capped at a maximum value, which represents the potential of a specific area that is restricted in terms of the maximum allowable MW due to limitations on building licenses.
+
+In order to define the groups in the model, the following steps are necessary:
+
+1. Create a group in the [`groups-data.csv`](@ref schemas) file by defining the `name` property and its parameters.
+2. In the file [`assets-data.csv`](@ref schemas), assign to the assets that belong to the group the `name` of the group they belong in the `group` parameter/column.
```
clizbe, line 385:
```suggestion
2. In the file [`assets-data.csv`](@ref schemas), assign assets to the group by putting the `name` in the `group` parameter/column.
```

docs/src/how-to-use.md
```diff
@@ -354,11 +369,58 @@ For the model to add constraints for a [maximum or minimum energy limit](@ref in
...
+
+    > **Note:**
+    > A missing value in the parameter `group` in the [`assets-data.csv`](@ref schemas) means that the asset does not belong to any group.
+
+Groups are useful to represent several common constraints, the following group constraints are the ones available in the model.
```
clizbe, line 390:
```suggestion
Groups are useful to represent several common constraints, the following group constraints are available.
```

docs/src/how-to-use.md
```diff
@@ -354,11 +369,58 @@ For the model to add constraints for a [maximum or minimum energy limit](@ref in
...
+
+### [Setting up a maximum or minimum investment limit for a group](@id investment-group-setup)
+
+The mathematical formulation of the maximum and minimum investment limit for group constraints is available [here](@ref investment-group-constraints). The parameters to set up these constraints in the model are in the [`groups-data.csv`](@ref schemas) file.
+
+-   `invest_method = true`. This parameter enables the model to use the investment group constraints.
```
clizbe, line 396: Can we change it to "invest_limit" or "limit_investment" instead of method?
Method to me means the way of investing or something.
Or "constrain_investment" maybe?

docs/src/how-to-use.md
```diff
@@ -354,11 +369,58 @@ For the model to add constraints for a [maximum or minimum energy limit](@ref in
...
+```@example display-group-setup
+input_asset_file = "../../test/inputs/Norse/assets-data.csv" # hide
+assets = CSV.read(input_asset_file, DataFrame, header = 2) # hide
+assets = assets[.!ismissing.(assets.group), [:name, :type, :group, :investable]] # hide
+```
+
+Here we can see that the assets `Asgard_Solar` and `Midgard_Wind` belong to the `renewables` group, while the assets `Asgard_CCGT` and `Midgard_CCGT` belong to the `ccgt` group.
+
+> **Note:**
+> Remember, if the `invest_method` is set to `true` for the groups, then the assets must also have the `investable` parameter set to `true`. This is necessary for formulating the group constraints correctly. If the assets in the group are not investable but the group has a minimum limit to fulfill, it can lead to an infeasible model. For example, if there are no investable assets in the group and the group has a minimum investment constraint, it would be impossible to meet the constraint without any available investments.
```
clizbe, line 426:
```suggestion
> If the group has a `min_investment_limit`, then assets in the group have to allow investment (`investable = true`) for the model to be feasible. If the assets are not `investable` then they cannot satisfy the minimum constraint.
```

clizbe (changes requested):
Looks pretty good! Just a few comments. :)
Thanks for the PR!

docs/src/formulation.md
clizbe: Ah now I think it should be "This set contains assets that have a group investment limit." Or something like that.

datejada:
> Does this implementation assume an asset can have only 1 group?

Correct! Nice comment 😄 We can add an improvement to it in a new issue to make it more general since we might want a unit belonging to more than one group 😉 Thanks!

docs/src/how-to-use.md
datejada: mmm...the method is Juha's style that we are trying to mimic...I don't have a strong opinion on that...Maybe we can ask @g-moralesespana what he prefers as name for the parameter...

docs/src/formulation.md
datejada: Done! Thanks

docs/src/formulation.md
datejada: Done! Thanks!

docs/src/how-to-use.md
datejada: @clizbe I have created issue #747 to add the possibility of an asset belonging to more than one group and review the naming of the constraints' activation. So, we can leave it like this now and discuss it later when tackling the issue. Is that ok?

docs/src/formulation.md
```diff
@@ -50,19 +50,32 @@ In addition, the following flow sets represent methods for incorporating additio
 | $\mathcal{K}$     | Representative periods (rp)                         | $k \in \mathcal{K}$       | $\mathcal{K} \subset \mathbb{N}$ | $\mathcal{K}$ does not have to be a subset of $\mathcal{P}$                |
 | $\mathcal{B}_{k}$ | Timesteps blocks within a representative period $k$ | $b_{k} \in \mathcal{B}_k$ |                                  | $\mathcal{B}_k$ is a partition of timesteps in a representative period $k$ |
 
+### Sets for Groups
+
+| Name                     | Description             | Elements                       | Superset | Notes |
+| ------------------------ | ----------------------- | ------------------------------ | -------- | ----- |
+| $\mathcal{G}^{\text{a}}$ | Groups of energy assets | $g \in \mathcal{G}^{\text{a}}$ |          |       |
```
gnawin, line 57: When I read this I interpret $\mathcal{G}^{\text{a}}$ is a subset of $\mathcal{G}$ filtered by asset $a$, following the notion of  $\mathcal{A}^{\text{i}}$ and $\mathcal{A}$ for example. So I got a bit confused because there, assuming the same notion, we don't have (yet) $\mathcal{G}$ without any superscripts, so to me $\mathcal{G}$ and $\mathcal{G}^{\text{i}}$ is enough. Do we need to differentiate this?

gnawin (approved):
Thanks a lot Diego. LTGM. I only leave a comment/question, but did not request any changes. I'll leave it up to you to decided what is the best way to deal with that.

docs/src/formulation.md
datejada: Hi @gnawin, Thanks for the comment. I am trying to look here to the big picture, so in the future, we can have groups of flows $G^f$ (It will be helpful also for flow investments but limiting the sum of a group of flows). So, I did it like this to differentiate groups of assets and groups of flows, although groups of flows are not yet in the model. Does that make sense?

gnawin: Yes, it makes sense. I was, only vaguely, thinking of the flows, because I know it would be there but not yet, and you have a better understanding of what you are going to do of course. But my first thought was, for the flows, we still look at the assets, e.g., we only want to sum up all the flows for the group wind. That is why I didn't realize first you would need a separate set. But I can also imagine, that given the Tulipa way, making a set for flows more explicit may be more consistent and clear. 

Just some fruits for thoughts.

PR #745, opened 2024-08-21
Add basic multi-year support

abelsiqueira:
**Work in progress**

Simplify and try to refactor the code in #741 to make a more manageable pull request.

Implement basic multi-year support.

# Pull request details

Related to #740 and #741

## Collaboration confirmation

As a contributor I confirm

-   [x] Tests are passing
-   [x] Lint is passing

### DIFF
--- a/src/create-model.jl
+++ b/src/create-model.jl
@@ -1,37 +1,114 @@
-export create_model!, create_model, construct_dataframes, filter_assets, filter_flows
+export create_model!, create_model, construct_dataframes
 
 """
-    Helper function to filter assets in the graph given a key and value
+    _get_graph_asset_or_flow(graph, a)
+    _get_graph_asset_or_flow(graph, (u, v))
+
+Returns `graph[a]` or `graph[u, v]`.
 """
-filter_assets(graph, assets, key, values) =
-    filter(a -> !ismissing(getfield(graph[a], key)) && getfield(graph[a], key) == values, assets)
-filter_assets(graph, assets, key, values::Vector{String}) =
-    filter(a -> !ismissing(getfield(graph[a], key)) && getfield(graph[a], key) in values, assets)
+_get_graph_asset_or_flow(graph, a) = graph[a]
+_get_graph_asset_or_flow(graph, f::Tuple) = graph[f...]
 
 """
-    Helper function to filter flows in the graph given a key and value
+    get_graph_value_or_missing(graph, graph_key, field_key)
+    get_graph_value_or_missing(graph, graph_key, field_key, year)
+
+Get `graph[graph_key].field_key` (or `graph[graph_key].field_key[year]`) or return `missing` if
+any of the values do not exist.
+We also check if `graph[graph_key].active[year]` is true if the `year` is passed and return
+`missing` otherwise.
 """
-filter_flows(graph, flows, key, values) = filter(f -> getfield(graph[f...], key) == values, flows)
+function get_graph_value_or_missing(graph, graph_key, field_key)
+    g = _get_graph_asset_or_flow(graph, graph_key)
+    # This is not missing in the tests
+    # if g === missing
+    #     return missing
+    # end
+    # if !hasproperty(g, field_key)
+    #     return missing
+    # end
+    return getproperty(g, field_key)
+end
+function get_graph_value_or_missing(graph, graph_key, field_key, year)
+    g = get_graph_value_or_missing(graph, graph_key, field_key)
+    # This is not missing in the tests
+    # if ismissing(g)
+    #     return missing
+    # end
+    if !_get_graph_asset_or_flow(graph, graph_key).active[year]
+        return missing
+    end
+    return get(g, year, missing)
+end
+
+"""
+    safe_comparison(graph, a, value, key)
+    safe_comparison(graph, a, value, key, year)
+
+Check if `graph[a].value` (or `graph[a].value[year]`) is equal to `value`.
+This function assumes that if `graph[a].value` is a dictionary and `value` is not, then you made a mistake.
+This makes it safer, because it will not silently return `false`.
+It also checks for missing.
+"""
+function safe_comparison(graph, a, value1, args...)
+    value2 = get_graph_value_or_missing(graph, a, args...)
+    if ismissing(value1) || ismissing(value2)
+        return false
+    end
+    return cmp(value1, value2) == 0 # Will error is one is a container (vector, dict) and the other is not
+end
+
+"""
+    safe_inclusion(graph, a, value, key)
+    safe_inclusion(graph, a, value, key, year)
+
+Check if `graph[a].value` (or `graph[a].value[year]`) is in `values`.
+This correctly check that `missing in [missing]` returns `false`.
+"""
+function safe_inclusion(graph, a, values::Vector, args...)
+    value = get_graph_value_or_missing(graph, a, args...)
+    return coalesce(value in values, false)
+end
+
+"""
+    filter_graph(graph, elements, value, key)
+    filter_graph(graph, elements, value, key, year)
+
+Helper function to filter elements (assets or flows) in the graph given a key (and possibly year) and value (or values).
+In the safest case, this is equivalent to the filters
+
+```julia
+filter_assets_whose_key_equal_to_value = a -> graph[a].key == value
+filter_assets_whose_key_year_equal_to_value = a -> graph[a].key[year] in value
+filter_flows_whose_key_equal_to_value = f -> graph[f...].key == value
+filter_flows_whose_key_year_equal_to_value = f -> graph[f...].key[year] in value
+```
+"""
+filter_graph(graph, elements, value, args...) =
+    filter(e -> safe_comparison(graph, e, value, args...), elements)
+filter_graph(graph, elements, values::Vector, args...) =
+    filter(e -> safe_inclusion(graph, e, values, args...), elements)
 
 """
     dataframes = construct_dataframes(
         graph,
         representative_periods,
-        constraints_partitions,
+        constraints_partitions,, IteratorSize
+        years,
     )
 
 Computes the data frames used to linearize the variables and constraints. These are used
 internally in the model only.
 """
-function construct_dataframes(graph, representative_periods, constraints_partitions)
+function construct_dataframes(graph, representative_periods, constraints_partitions, years)
     A = MetaGraphsNext.labels(graph) |> collect
     F = MetaGraphsNext.edge_labels(graph) |> collect
-    RP = 1:length(representative_periods)
+    RP = Dict(year => 1:length(representative_periods[year]) for year in years)
 
     # Create subsets of assets
-    Ap  = filter_assets(graph, A, :type, "producer")
-    Acv = filter_assets(graph, A, :type, "conversion")
-    Auc = (Ap ∪ Acv) ∩ filter_assets(graph, A, :unit_commitment, true)
+    Ap  = filter_graph(graph, A, "producer", :type)
+    Acv = filter_graph(graph, A, "conversion", :type)
+    Auc = Dict(year => (Ap ∪ Acv) ∩ filter_graph(graph, A, true, :unit_commitment, year) for year in years)
 
     # Output object
     dataframes = Dict{Symbol,DataFrame}()
@@ -42,6 +119,7 @@ function construct_dataframes(graph, representative_periods, constraints_partiti
             # No data, but ensure schema is correct
             dataframes[key] = DataFrame(;
                 asset = String[],
+                year = Int[],
                 rep_period = Int[],
                 timesteps_block = UnitRange{Int}[],
                 index = Int[],
@@ -53,9 +131,9 @@ function construct_dataframes(graph, representative_periods, constraints_partiti
         df = DataFrame(
             (
                 (
-                    (asset = a, rep_period = rp, timesteps_block = timesteps_block) for
+                    (asset = a, year = y, rep_period = rp, timesteps_block = timesteps_block) for
                     timesteps_block in partition
-                ) for ((a, rp), partition) in partitions
+                ) for ((a, y, rp), partition) in partitions
             ) |> Iterators.flatten,
         )
         df.index = 1:size(df, 1)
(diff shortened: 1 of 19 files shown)


abelsiqueira:
My approach was:
Start from Ni's PR
Remove all investment_year (changed to year in some cases) and new constraints
Focus on having a new column year, keeping Ni's structure changes, and making the tests pass
Don't think too much about timeframe
Don't think about efficiency
The situation so far is
Tests are passing, but not 100% coverage, and docs are broken.
It is rebased on top of main.
The changes are still massive and maybe that is unavoidable. 95% of places with a graph[a].prop need a small change to add [year].
There are 6 commits, mostly independent, but not atomic.
The model right now is essentially a model per year.
I've decided to keep the new investment variables (with sparse indexes) as it is, because the alternative would be figuring out a worse temporary way of dealing with years.

abelsiqueira:
Open questions:

- Can we make this even smaller?
- How do timeframe and year interact?

src/io.jl
```diff
@@ -60,60 +74,137 @@ function create_internal_structures(connection)
...
+                " WHERE " *
+                join(("$k=$(TulipaIO.FmtSQL.fmt_quote(v))" for (k, v) in where_pairs), " AND ")
+        end
+        DuckDB.query(connection, _q)
+    end
+    function _get_stuff(table_name, col; where_pairs...)
+        result = _query(table_name, col; where_pairs...)
+        Dict(row.year => getproperty(row, Symbol(col)) for row in result)
+    end
+
```
gnawin, line 90: Do you have to/want to initialize all the years in the dict? Because I see in later codes, you always check whether if the key (year) exists or not, and if not, initialize with `get()`.

src/create-model.jl
```diff
@@ -392,22 +448,41 @@ function create_model(
...
         return length(timesteps_block) * representative_periods[rp].resolution
     end
+
+    Y = years = collect(keys(representative_periods)) # need to decide where to get the years
+
     # Maximum timestep
-    Tmax = maximum(last(rp.timesteps) for rp in representative_periods)
+    Tmax = maximum(last(rp.timesteps) for year in Y for rp in representative_periods[year])
+
     expression_workspace = Vector{JuMP.AffExpr}(undef, Tmax)
```
gnawin, line 438: how does it work in the creation of dataframes, when we don't add `year` to `expression_workspace`? because here the `expression_workspace` is only defined as the max timesteps within a rp

src/create-model.jl
```diff
@@ -392,22 +448,41 @@ function create_model(
...
+        filter_assets(key, value) =
+            filter(a -> !ismissing(getfield(graph[a], key)) && getfield(graph[a], key) == value, A)
         filter_assets(key, values::Vector{String}) =
             filter(a -> !ismissing(getfield(graph[a], key)) && getfield(graph[a], key) in values, A)
-        filter_flows(key, values) = filter(f -> getfield(graph[f...], key) == values, F)
+
+        filter_assets(key, year, value) = filter(
+            a -> begin
+                field_value = get(getfield(graph[a], key), year, missing)
+                get(graph[a].active, year, false) &&
```
gnawin, line 452: just a generic comment: if we only use one row for one unit name each year, then `graph[a].active` does not make sense anymore, we need to change this later

gnawin (commented):
Hi @abelsiqueira thanks a lot. I left a few comments/things to consider later.

In general, the next steps are:

- discuss the source for reading in the data for years
- discuss the data structure
- add the formulation (we will have two versions, unfortunately we dropped the most detailed version I worked won), and adding the missing data

test/inputs/Multi-year Investments/assets-data.csv
```diff
@@ -0,0 +1,21 @@
...
+wind2030,producer,,2030,,true,true,true,70,1,,50,0,0,,false,0,0,0,0,false,0,0,,0,false,,,
+solar2030,producer,,2030,,true,true,true,50,1,,10,0,0,,false,0,0,0,0,false,0,0,,0,false,,,
+ens2030,producer,,2030,,true,false,false,0,1,,0,1115,0,,false,0,0,0,0,false,0,0,,0,false,,,
+battery2030,storage,,2030,,true,true,false,0,1,,10,10,0,,false,0,20,0,0,false,0,0,,0,false,,,
+phs2030,storage,,2030,,true,true,false,0,1,,100,100,0,,true,0,4800,2400,0,true,0,0,,0,false,,,
+demand2030,consumer,,2030,,true,false,false,0,0,,0,0,1115,,false,0,0,0,0,false,0,0,,0,false,,,
+gas_import2050,producer,,2050,,true,false,false,0,0,,0,75000,0,,false,0,0,0,0,false,0,0,,0,false,,,
+ocgt2050,conversion,,2050,,true,false,true,25,1,,100,0,0,,false,0,0,0,0,false,0,0,,0,false,,,
+wind2050,producer,,2050,,true,false,true,70,1,,50,0,0,,false,0,0,0,0,false,0,0,,0,false,,,
+wind2050,producer,,2050,,true,true,true,70,1,,50,0,0,,false,0,0,0,0,false,0,0,,0,false,,,
```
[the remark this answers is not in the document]
abelsiqueira: Yeah, probably some mistake in the replace commands. I haven't done anything significant in these files, so don't read much in their meaning.

src/io.jl
abelsiqueira: This does not add all years because `result` is a query from `assets_data`. So only the years informed in that file are here.
So if you have `wind,2030`, `gas,2030` and `gas,2050`, your `result` for `wind` will have only 2030, while `gas` will have 2030 and 2050.

src/create-model.jl
abelsiqueira: The `expression_workspace` is a temporary pre-allocated array.
In the creation of dataframes, this is how it is used:

- Clean `workspace`
- Fill `workspace` looping over some partition
- Aggregate variables over some other partition and save then in the dataframe

So you don't need the `workspace` for anything else after that, so there is no need to store it for later, and so there is no need to separate per year, because they are completely ignored outside of the loop.
The only requirement is that the `workspace` is large enough to hold all the indices that the partition touch.

src/create-model.jl
abelsiqueira: Agreed. We need to figure out how to deal with the lifetime column and its implications in the model.

src/time-resolution.jl
```diff
@@ -13,86 +13,91 @@ which specifies the partition strategy for each resolution (i.e., lowest or high
...
 
     for (name, partitions, strategy, asset_filter) in partitions_cases
         constraints_partitions[name] = OrderedDict(
-            (a, rp) => begin
-                P = partitions(a, rp)
+            (a, y, rp) => begin
+                P = partitions(a, y, rp)
                 if length(P) > 0
-                    compute_rp_partition(partitions(a, rp), strategy)
+                    compute_rp_partition(partitions(a, y, rp), strategy)
```
gnawin, line 95: Here's another question:

I had before something like this:

```
result = compute_rp_partition(partitions(a, y, rp), strategy)
result
```

This is because when I didn't add this explicit output, it gave me an error...do you have any thoughts here why that could happen?

src/time-resolution.jl
```diff
@@ -205,10 +210,7 @@ compute_rp_partition([partition1, partition2], :highest)
  11:12
 ```
 """
-function compute_rp_partition(
-    partitions::AbstractVector{<:AbstractVector{<:UnitRange{<:Integer}}},
-    strategy,
-)
+function compute_rp_partition(partitions, strategy)
```
gnawin, line 213: I removed the type because it gave an error before, but maybe now it works with the type?

src/create-model.jl
```diff
@@ -1,37 +1,114 @@
...
+"""
+function get_graph_value_or_missing(graph, graph_key, field_key)
+    g = _get_graph_asset_or_flow(graph, graph_key)
+    # This is not missing in the tests
+    # if g === missing
+    #     return missing
+    # end
+    # if !hasproperty(g, field_key)
+    #     return missing
+    # end
```
gnawin, line 29: should we leave this?

src/create-model.jl
```diff
@@ -1,37 +1,114 @@
...
+    return getproperty(g, field_key)
+end
+function get_graph_value_or_missing(graph, graph_key, field_key, year)
+    g = get_graph_value_or_missing(graph, graph_key, field_key)
+    # This is not missing in the tests
+    # if ismissing(g)
+    #     return missing
+    # end
```
gnawin, line 37: also this?

gnawin (changes requested):
Hi @abelsiqueira thanks a lot for the PR. I left some minor comments here and there. 

I have two general questions/remarks:

- I noticed we used the parameter `active` in many places, that was on me because back then I was not fully sure about the data structure and now I have more ideas. I just want to mention that most probably we'll have to adapt that soon. Do you think this would be a _structural_ change then? I think we have no other way out for now, but still I'd like to raise this question.

- I also want to double-check with you that the changes to the model in this PR, i.e., the variables and objective, _only_ add an index to the current model. And that all the results would only have an extra year index, which is 2030 by default. Is this correct? 

  And a related question: even if we don't do multi-year, do we always have to have the year index for the outputs? Probably a separate issue, but I'm thinking maybe we could, in such cases, modify the outputs so we don't have the year index.

src/io.jl
```diff
@@ -133,18 +223,44 @@ function create_internal_structures(connection)
...
-        )
+        for year in years
+            is_active = get(graph[u, v].active, year, false)
+            if !is_active
+                continue
+            end
+            # we only compute partitions for active flows, but we initialize all years. If we don't want the all the years as keys, we should not do this.
+            graph[u, v].rep_periods_partitions[year] = Dict{Int,Vector{TimestepsBlock}}()
+            if haskey(graph[u, v].active, year) && graph[u, v].active[year]
+                # we only compute partitions for active flows.
```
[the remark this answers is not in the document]
abelsiqueira: Maybe we should add an inactive flow instead, then. The expectation is that this will be required, right?

src/io.jl
```diff
@@ -163,51 +279,73 @@ function create_internal_structures(connection)
              WHERE assets_data.is_seasonal
          """
     for row in DuckDB.query(connection, find_assets_partitions_query)
-        graph[row.name].timeframe_partitions = _parse_rp_partition(
-            Val(Symbol(row.specification)),
-            row.partition,
-            1:timeframe.num_periods,
-        )
+        for year in years
+            graph[row.name].timeframe_partitions[year] = _parse_rp_partition(
```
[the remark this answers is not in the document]
abelsiqueira: The two cases are different. Here, assignment and initialisation happen together. I am saying

```Julia
dict[2030] = new_stuff
dict[2050] = new_stuff
```

The contents of `new_stuff` is irrelevant.

In the other case below, `new_stuff` is a dictionary that we modify after creation. So we do

```Julia
dict[2030] = Dict()
dict[2030][key1] = content
dict[2030][key2] = content
```

Notice that the `key1` and `key2` elements are explicit here. In the first example, this is now the case. The equivalent version of the second code with the first example would be

```Julia
dict[2030] = Dict(key1 => content, key2 => content)
```

However, in the second example we don't know all the keys and content beforehand, so we create an empty dictionary and fill the values as they come.

src/io.jl
```diff
@@ -163,51 +279,73 @@ function create_internal_structures(connection)
...
         )
-        for ((rep_period,), df) in pairs(gp) # Loop over filtered DFs by rep_period
-            graph[asset_profile_row.asset].rep_periods_profiles[(
-                asset_profile_row.profile_type,
-                rep_period,
-            )] = df.value
+        for ((rep_period, year), df) in pairs(gp) # Loop over filtered DFs by rep_period, year
+            profiles = graph[asset_profile_row.asset].rep_periods_profiles
+            if !haskey(profiles, year)
+                profiles[year] = Dict{Tuple{Symbol,Int},Vector{Float64}}()
```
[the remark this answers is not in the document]
abelsiqueira: Answered above

src/io.jl
```diff
@@ -163,51 +279,73 @@ function create_internal_structures(connection)
...
+                graph[flow_profile_row.from_asset, flow_profile_row.to_asset].rep_periods_profiles
+            if !haskey(profiles, year)
+                profiles[year] = Dict{Tuple{Symbol,Int},Vector{Float64}}()
+            end
+
+            profiles[year][(flow_profile_row.profile_type, rep_period)] = df.value
         end
     end
 
+    # TODO: After figuring out interaction between year and timeframe, come back here
```
abelsiqueira, line 328:
```suggestion
```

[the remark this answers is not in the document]
abelsiqueira: I hope not

src/time-resolution.jl
abelsiqueira: Not really, it shouldn't be the case. Both codes should do the same thing, as far as I can see, which is why I removed the change.

src/create-model.jl
abelsiqueira: The best solution would be to have these cases in the tests. But then how to make that happen exactly, I'm not sure.
If we enable the check, then the coverage decreases.
Removing would be easier. Do you agree with that?

abelsiqueira:
@gnawin 
> * I noticed we used the parameter `active` in many places, that was on me because back then I was not fully sure about the data structure and now I have more ideas. I just want to mention that most probably we'll have to adapt that soon. Do you think this would be a _structural_ change then? I think we have no other way out for now, but still I'd like to raise this question.

I think it won't be as bad. Also, checking for `active` should be fine either way.

> * I also want to double-check with you that the changes to the model in this PR, i.e., the variables and objective, _only_ add an index to the current model. And that all the results would only have an extra year index, which is 2030 by default. Is this correct?

The changes to the model should be only "add year to variables".
All the results for existing test cases should not change. The year 2030 was added for all of them, but since it is the same for all of them, nothing changes.

>   And a related question: even if we don't do multi-year, do we always have to have the year index for the outputs? Probably a separate issue, but I'm thinking maybe we could, in such cases, modify the outputs so we don't have the year index.

That's probably better handled as post-processing. The full output data includes the year, and removing it for this case doesn't feel like a Tulipa task.

src/io.jl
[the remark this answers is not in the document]
abelsiqueira: Discussing in person, will update

gnawin (approved):
Great work and thanks a lot!! LGTM.

abelsiqueira:
Thanks for the review!

PR #757, opened 2024-08-26
Add year information to the data

### LINKED ISSUE #748: Multi-year data and timeframe relation
### What and Why

WIP - This issue will be edited directly

On the relationship between timeframe and year, this is what I understood:

- Whatever data we have must be specified by years - all hourly profiles will have 8760 hours
- Each year will be partitioned into consecutive **periods**. E.g., 365 days per year. Each period can have a different length.
- Using all periods, we construct the **representative periods (RPs).** The RPs can be used by any timeframe period of any year.
- During construction of the RPs, we create a map between timeframe periods and RPs. More specifically, we create a weight matrix to define a convex combination of RPs per timeframe period.

Implementation changes:

- New table to define years. Suggestion:
     - Table `year_data`:
        | year (?) | length (hours) |
        | --- | --- |
        | 2023 | 8760 |
        | 2024 | 8784 |
    - For now, we will ignore the length.
    - In the future, we can add a column duration.
- New table to define `timeframe` and the mapping. Suggestion:
    - Table `timeframe_data` , example:
        | year | timeframe (id) | length (hours) |
        | --- | --- | --- |
        | 2023 | 1 | 24 |
        | 2023 | 2 | 24 |
        | 2024 | 1 | 24 |
    - In TEM, we can always ignore the `length`
- Add `year` when there is `rep_period` or a `period`, since both RP and timeframe now depend on year.
    - This also applies to the file `rep-periods-mapping.csv` which has both `period` and `rep_period`. A single `year` column is added.
- For the linking files, year is added as well, although whether it makes sense is not clear.

### Possible Drawbacks

_No response_

### Related Issues

_No response_

abelsiqueira:
Add the file `utils/scripts/prepare-for-multiyear.jl` to add year
information to the input data.
Summary of changes:

- Add year and lifetime to assets-data
- Add year to flows-data
- Create new file year-data
- Create new file timeframe-data
- All files that had either a column `period` or `rep-period` now have a
  column `year` before them as well.
- The profile linking files also have a `year` information now
- The `assets-timeframe-partitions` now has year as well
- The `groups-data` now has year as well

# Pull request details

Closes #748

## Collaboration confirmation

As a contributor I confirm

-   [ ] Tests are passing
-   [ ] Lint is passing


utils/scripts/prepare-for-multiyear.jl
```diff
@@ -0,0 +1,125 @@
...
+include("../csv-modifications.jl")
+
+# Local cleanup to ensure clean files
+run(`git restore test/inputs/ benchmark/EU/`)
+
+# Add year and lifetime to assets data
+apply_to_files_named("assets-data.csv") do path
+    change_file(path) do tcsv
+        add_column(tcsv, "year", 2030; position = 5)
+        add_column(tcsv, "lifetime", ""; position = 6)
```
gnawin, line 12: The parameter `lifetime` will be needed later, but it would be needed in the form of two parameters `economic lifetime` and `technical lifetime`. So I would suggest to remove this for now, otherwise we'll need to rename it, which is also fine if that's easy given the new script.

gnawin:
Thanks Abel, I leave one comment/suggestion, but otherwise LGTM.

utils/scripts/prepare-for-multiyear.jl
abelsiqueira: I can remove and you add when you make the model changes, does that make sense?

gnawin: That sounds great. So I can add all the needed parameters later.

datejada (approved):
Thanks for the PR!

PR #771, opened 2024-09-05
Add more meaningful data for multi-year

### LINKED ISSUE #765: Create a test dataset for multi-year that makes more sense, considering existing capacities with different profiles.

gnawin:
# Pull request details

- add essential data to multi-year inputs, in particular, a column called `commission_year` is added. I'm hesitating this should be a new index or just a property of the asset, but I'm leaning towards the latter.
- note now we only have milestone years in year_data.csv, maybe we don't need that label
- rename `year` in assets-profiles to `commission_year`, and modify the name of the profile to reflect the commission_year
- use the modified name in profiles-rep-periods, so here the `year` means the year of the profile, but units of different commission years can be different profiles
- update all the other inputs according to the new format
- change the profile generation and the corresponding structure; currently I'm using dict again, which brings some problems later (I stop here now, because this may be a bigger change and I'd like to discuss first with @abelsiqueira in a working session)
- the tests are not passing yet because of the profile calculation.

Closes #765 

## Collaboration confirmation

As a contributor I confirm

-   [ ] Tests are passing
-   [ ] Lint is passing


abelsiqueira (approved):
Thanks! LGTM

PR #782, opened 2024-09-10
Add common dicts and an expression for multi-year 

### LINKED ISSUE #778: Add to model the common lists/subsets for years for both simple and compact methods

gnawin:
# Pull request details

This PR is meant to add some common sets/expressions that prepares for the parallel work for multi-year:

- Add a dict for investable years per asset
- Add a dict for pairs of (year, asset) of starting years
- Add an expression to incorporate the above-defined two dicts. This is not yet complete, but it sets a base that allows @datejada and me to branch for there.

Closes #778 

## Collaboration confirmation

As a contributor I confirm

-   [ ] Tests are passing
-   [ ] Lint is passing


gnawin:
@datejada I want to mention that this expression is currently defined for `a in Ai[y]`, which is only true for the accumulated investments. However, for the final expression we want, it should be for `a in A`, because we need to consider initial capacity as well (they don't have to be investable). I don't see it directly how to do this. Maybe we can discuss further, or if you will do something just let me know.

@abelsiqueira this PR is done by @datejada and me together, it is currently doing what we want, but we are not sure if there are better ways to code it. Do you have any suggestions?

datejada:
> @datejada I want to mention that this expression is currently defined for `a in Ai[y]`, which is only true for the accumulated investments. However, for the final expression we want, it should be for `a in A`, because we need to consider initial capacity as well (they don't have to be investable). I don't see it directly how to do this. Maybe we can discuss further, or if you will do something just let me know.
> 
> @abelsiqueira this PR is done by @datejada and me together, it is currently doing what we want, but we are not sure if there are better ways to code it. Do you have any suggestions?

Thanks @gnawin, I agree, the expression should be on A rather than `Ai`, but it is fine, I will take of it once I start my part since this expression if for the simple formulation 😉

src/create-model.jl
```diff
@@ -604,6 +604,13 @@ function create_model(
         Ai = Dict(y => filter_graph(graph, A, true, :investable, y) for y in Y)
         Fi = Dict(y => filter_graph(graph, F, true, :investable, y) for y in Y)
 
+        Yi = Dict(
+            a => [y for y in Y if a in Ai[y]] for a in A if any(graph[a].investable[y] for y in Y)
+        )
```
abelsiqueira, line 609: I found this a bit complicated, but I don't have a better solution. Maybe add a comment to indicate that this is inverting `Ai`?

src/create-model.jl
```diff
@@ -889,6 +896,15 @@ function create_model(
         JuMP.drop_zeros!.(outgoing_flow_storage_inter_rp_balance)
     end
 
+    ## Expressions for multi-year investment
+    @timeit to "multi-year investment" begin
+        @expression(
+            model,
+            accumulate_capacity_simple[y ∈ Y, a ∈ Ai[y]],
+            sum(assets_investment[i, a] for i in Yi[a] if starting_year[(y, a)] ≤ i ≤ y)
```
abelsiqueira, line 904:
```suggestion
            accumulate_capacity_simple[y ∈ Y, a ∈ Ai[y]],
            sum(assets_investment[yi, a] for yi in Yi[a] if starting_year[(y, a)] ≤ yi ≤ y)
```

abelsiqueira (approved):
Hi both. I've made some small comments. Otherwise it's good to go. Thanks.

src/create-model.jl
gnawin: Thank you! Exactly, you get it although it look a bit ugly, this is inverting `Ai`, which probably means it is correct. I'll add a comment.

PR #784, opened 2024-09-10
Add compact method for multi-year

### LINKED ISSUE #783: Add the compact method

gnawin:
# Pull request details

Closes #783 

## Collaboration confirmation

As a contributor I confirm

-   [ ] Tests are passing
-   [ ] Lint is passing


gnawin:
@abelsiqueira I'm pushing this draft PR for now just for you to have a look and see if there can be any improvements already. Thank you!

Specifically I have two questions:

- We need the extra index for commission_year/investment_year for the parameter. Maybe, if you have time, you can think of how to do this properly.
- The new expressions/variables do not seem to be very elegant. Do you have any suggestions there?

src/create-model.jl
```diff
@@ -663,6 +667,15 @@ function create_model(
...
         @variable(model, 0 ≤ assets_investment[y in Y, a in Ai[y]])  #number of installed asset units [N]
+        @variable(
+            model,
+            0 <= assets_decommission_compact[
+                a in A, # we need a new set A, we don't want e.g., demand
+                y in Y,
+                v in V;
+                starting_year_compact[y, a] ≤ v < y, # double-check formulation; this is the domain, different from the sum
+            ]
+        )
```
datejada, line 678: This variable should only exist for the assets that have the compact formulation. We need a new set with the assets that have the compact method

src/create-model.jl
```diff
@@ -613,6 +614,9 @@ function create_model(
         starting_year =
             Dict((y, a) => y - graph[a].technical_lifetime[y] + 1 for y in Y for a in Ai[y])
 
+        starting_year_compact =
+            Dict((y, a) => y - graph[a].technical_lifetime[y] + 1 for y in Y for a in A) # needs to be modified, because we want a that is either investable or existing
+
```
datejada, line 619: After looking at the changes twice, the starting year depends on the method, but it is the same parameter. So, instead of calculating the value in two different parameters, we need only one parameter, `starting_year`, whose value depends on the method.

abelsiqueira:
Hi @gnawin, it looks like the complexity exists because of the indexes, so I think we have to use DataFrames.
In other words, you have conditionals that make the indexes sparse, so maybe we have to store the indexes in a linearlized fashion. It depends on what you're going to do with it later, but the idea would be to move the indexes for which the variables and/or expressions are defined to dataframes.

Note that `investment` variables also have sparse indexes, but since they were small and simple, and we were in a rush to implement multi-year, we chose to let them have sparse indexes temporarily. The idea being that they will be moved to table format later.
The same could be done here, but the complexity might build up. It depends on how you need to use these expressions later.

src/create-model.jl
```diff
@@ -905,6 +918,22 @@ function create_model(
...
+                v in V;
+                starting_year_compact[(y, a)] ≤ v ≤ y,
+            ],
+            graph[a].initial_units[y] + # need extra v index
+            (haskey(assets_investment, (v, a)) ? assets_investment[v, a] : 0) - sum(
+                assets_decommission_compact[a, i, v] for
+                i in Y if starting_year_compact[(y, a)] < i ≤ y &&
+                haskey(assets_decommission_compact, (a, i, v))
+            )
+        )
```
datejada, line 936: If you create a new set of investable assets with a compact method, this equation simplifies a lot. Rebase to the latest main, and look at the PR that solves #786 for a reference

datejada (approved):
LGTM

Issue #791, opened 2024-09-11
Update binary to avoid simultaneously charging/discharging constraints to consider multi-year investment methods


datejada:
Blocked by: https://github.com/TulipaEnergy/TulipaEnergyModel.jl/issues/786 and https://github.com/TulipaEnergy/TulipaEnergyModel.jl/issues/783

Issue #462, opened 2024-02-12
Multi-year investment


datejada:
### Description

Update the model to handle multi-year investments. The checks below should be converted into issues when this development starts, to keep the changes as atomic as possible.

One big challenge in this implementation is the data handling. For instance, the capacity of the assets will be year-dependent, meaning that they will no longer be a single value, but a vector or time series. Therefore we will have a large data set for the input data.

### Must-haves and nice-to-haves

For this epic, we want a working version that achieves the same features as COMPETES and allows different levels of detail for performance considerations. This means we want the following.

- [x] The main temporal structure would be milestone years for both investment and operation.
- [x] Simple weights for the in-between years.
- [x] Economic representation
- [x] Use total cost for Capex calculation.
- [x] Decommissioning for existing and new capacities (note: we allow inputs for existing capacities in the forms of both individual plants and aggregated capacities, as we have for COMPETES)

This leaves us the following nice-to-haves.

- Complex temporal structures such as milestone years for investment, and more years for operation
- Complex weights for the in-between years such as the mapping matrix.
- ~Add annualized cost for Capex calculation.~ (done in #810 #826)
- Mothballing, retrofitting.
- ~[Can be a new epic] A new "entity" called "group" to have a set of common constraints, such as investment potential for a group of assets (e.g., different wind techs a total potential in the NL)~(done in #732). Also, use methods such that one can have a common investment variable (if only the efficiencies are different).

### Sub issues

- [x] #690 which covers all the must-have features
- [x] #732 
- [x] #745 
- [x] #766
  - Separate static data and non-static/yearly data.
  - Differentiate milestone years and non-milestone years. We need to put some information on whether this is a milestone year (that we can invest) or this is a year for the vintage/existing capacities. Names can be discussed.
- [x] #765
- [x] #777
- [x] #778
- Different levels of detail 
  - [x] ~#722~ (in favor of #786 ) 
  - [x] ~#740~ (after discussion, we will drop this)
  - [x] #783
  - [x] #786
  - [x] #792
  - [x] #806
- [x] ~Add existing capacity, and create/modify the needed parameters~ (will be done within different levels of detail)
- [x] ~Add decommissioning in the model, and create the needed parameters~ (will be done within different levels of detail)
- [x] #797
- [x] #798
- [x] #790
- [x] #791
- [x] #789
- [x] #838
- [x] #802
- Add economic parameters
  - [x] #810
  - [x] #816

### Improvements
[...]

PR #793, opened 2024-09-11
Add simple method for multi year investment on assets

### LINKED ISSUE #786: Add the simplest method for multi-year investment

datejada:
# Pull request details

Changes to consider the simple method in the investment of assets.

Other related issues are:
- Update on ramping constraints #790 
- Update on binary method for storage #791 
- Add simple method for transport assets #792 

They are done in separate issues to keep the changes as atomic as possible.

Closes #786 

## Collaboration confirmation

As a contributor I confirm

-   [x] Tests are passing
-   [x] Lint is passing


datejada:
@abelsiqueira depending on the answer from @gnawin here https://github.com/TulipaEnergy/TulipaEnergyModel.jl/issues/462#issuecomment-2343312397 the code might simplify a bit more depending if we don't allow missing values on the investment method.

But, for now, I would appreciate your feedback. Also, these changes might help to have a better overview for the ones in #784

src/create-model.jl
```diff
@@ -902,8 +926,12 @@ function create_model(
...
         @expression(
             model,
-            accumulate_capacity_simple[y ∈ Y, a ∈ Ai[y]],
-            sum(assets_investment[yi, a] for yi in Yi[a] if starting_year[(y, a)] ≤ yi ≤ y)
+            accumulate_capacity_simple_method[y ∈ Y, a ∈ investable_assets_using_simple_method[y]],
+            graph[a].initial_units[y] +
+            sum(assets_investment[yi, a] for yi in Yi[a] if starting_year[(y, a)] ≤ yi ≤ y) - sum(
+                assets_decommission_simple_method[yi, a] for
+                yi in Yi[a] if starting_year[(y, a)] ≤ yi ≤ y
+            )
```
abelsiqueira, line 934:
```suggestion
            sum(
            assets_investment[yi, a] - assets_decomission_simple_method[yi, a]
            for yi in Yi[a] if starting_year[(y, a)] ≤ yi ≤ y)
            )
```
Linter will probably complain.

src/create-model.jl
```diff
@@ -662,10 +671,20 @@ function create_model(
...
-        @variable(model, 0 ≤ assets_investment[y in Y, a in Ai[y]])  #number of installed asset units [N]
         @variable(model, 0 ≤ flows_investment[y in Y, (u, v) in Fi[y]])
+
+        ### Investment variables
+        @variable(model, 0 ≤ assets_investment[y in Y, a in Ai[y]])  #number of installed asset units [N]
+        @variable(
+            model,
+            0 ≤ assets_decommission_simple_method[
+                y in Y,
+                a in investable_assets_using_simple_method[y],
```
gnawin, line 682: I think we need a different set here for a. We want to be able to decommission both investable assets and existing assets. Suppose we have some gas units that we don't want to invest in, but they are still there an can be decommissioned. That's why in the other PR, I made a comment/to-do, to define a set for A that covers both investable assets and existing assets.

abelsiqueira (approved):
Thanks for the PR. I noticed a place where we can join the `sum`s. It shouldn't be a performance bottleneck, so you can choose whatever makes the code more readable.

src/create-model.jl
```diff
@@ -902,8 +926,12 @@ function create_model(
     @timeit to "multi-year investment" begin
         @expression(
             model,
-            accumulate_capacity_simple[y ∈ Y, a ∈ Ai[y]],
-            sum(assets_investment[yi, a] for yi in Yi[a] if starting_year[(y, a)] ≤ yi ≤ y)
+            accumulate_capacity_simple_method[y ∈ Y, a ∈ investable_assets_using_simple_method[y]],
```
gnawin, line 929: - Same with above, we can have accumulated capacity even without investments, i.e., for existing capacity. So a needs a different set.
- Another thing here is we may have some existing capacity in 2030, but no longer in 2050, then we would create an expression in 2050 that is 0. Is that a problem? I think we should try to not create empty expressions, but here since it is only existing units, I don't know how to avoid this.

src/create-model.jl
```diff
@@ -902,8 +926,12 @@ function create_model(
...
+            graph[a].initial_units[y] +
+            sum(assets_investment[yi, a] for yi in Yi[a] if starting_year[(y, a)] ≤ yi ≤ y) - sum(
+                assets_decommission_simple_method[yi, a] for
+                yi in Yi[a] if starting_year[(y, a)] ≤ yi ≤ y
```
gnawin, line 933: - Same here, the year set and asset set should probably be reconsidered.

src/create-model.jl
```diff
@@ -914,7 +942,22 @@ function create_model(
...
+            )
+        )
+
+        assets_fixed_cost = @expression(
+            model,
+            sum(
+                graph[a].fixed_cost[y] *
+                graph[a].capacity[y] *
+                accumulate_capacity_simple_method[y, a] for y in Y for
+                a in investable_assets_using_simple_method[y]
```
gnawin, line 954: - Same here, we also want it to work with only initial/existing units that we don't invest in anymore.

gnawin (changes requested):
Hi @datejada thanks for the PR, I left a few comments. We can discuss in person today if you want.

src/create-model.jl
datejada: Thanks! I like it as you proposed. I did it locally, to avoid the Linter complaining during the CI/CD

datejada:
@gnawin, then let's review it together in person because the code reflects the equations that are in the pdf 🤷

datejada:
@gnawin, my takeaway is that we might need an extra parameter to cover the cases you describe more easily. Otherwise, I don't see a combination that allows existing capacity to be decommissioned without being an investable asset (with the current parameters' definitions).

src/create-model.jl
datejada: Latest changes account for the reviewed formulation

src/create-model.jl
datejada: Latest changes account for the reviewed formulation

src/create-model.jl
datejada: Latest changes account for the reviewed formulation

src/create-model.jl
datejada: Latest changes account for the reviewed formulation

src/constraints/investment.jl
```diff
@@ -22,6 +24,9 @@ function add_investment_constraints!(
         if graph[a].capacity[y] > 0 && !ismissing(graph[a].investment_limit[y])
             bound_value = _find_upper_bound(graph, y, a)
             JuMP.set_upper_bound(assets_investment[y, a], bound_value)
+            if a in decommissionable_assets_using_simple_method
+                JuMP.set_upper_bound(assets_decommission_simple_method[y, a], bound_value)
+            end
```
gnawin, line 29:
```suggestion

```

src/constraints/investment.jl
```diff
@@ -10,9 +10,11 @@ function add_investment_constraints!(
     graph,
     Y,
     Ai,
+    decommissionable_assets_using_simple_method,
     Ase,
     Fi,
     assets_investment,
+    assets_decommission_simple_method,
```
gnawin, line 17:
```suggestion
```

src/constraints/investment.jl
```diff
@@ -10,9 +10,11 @@ function add_investment_constraints!(
     graph,
     Y,
     Ai,
+    decommissionable_assets_using_simple_method,
```
gnawin, line 13:
```suggestion
```

src/create-model.jl
```diff
@@ -1029,9 +1093,11 @@ function create_model(
         graph,
         Y,
         Ai,
+        decommissionable_assets_using_simple_method,
         Ase,
         Fi,
         assets_investment,
+        assets_decommission_simple_method,
```
gnawin, line 1100:
```suggestion
```

src/create-model.jl
```diff
@@ -1029,9 +1093,11 @@ function create_model(
         graph,
         Y,
         Ai,
+        decommissionable_assets_using_simple_method,
```
gnawin, line 1096:
```suggestion
```

gnawin (changes requested):
Hi @datejada thanks a lot for the PR! 

As discussed, let's for now leave out the upper bounds for the decommission variables and do that in a separate PR. 

The naming of things can be improved but let's do it later after my PR. 

And regarding your comments on the initial capacity/units, let's continue the discussion but also do it separately.

gnawin (approved):

Issue #701, opened 2024-07-29
Refactor the model


clizbe:
### Discussed in https://github.com/TulipaEnergy/TulipaEnergyModel.jl/discussions/688

<div type='discussions-op-text'>

<sup>Originally posted by **datejada** July  1, 2024</sup>

### Overview of the changes

- [x] #942
- [x] #885
We can control and improve both the creation of the data/tables and the create_model function better if these concerns are separated.
This includes the "expression" data, e.g., data that informs the incoming and outgoing flow.
  - This includes clustering stuff (representative periods, timeframe, years, etc.)
  - Partition should also be created before the model, though we are missing the "scenario specification" data as a separate entity
- Sets
  - [x] #892
- Variables
  - [x] #884
  The way that we control the indexing - to prevent sparsity issues - is to precompute the indexes of the variables. The way that we are doing it right now is to compute tables for these indexes. Assuming that keeps being the case, we have to define what is necessary for these tables and define their names.
  These tables will probably also save the final result, so they are used outside of the model as well.
    - If we use tables as DuckDB tables, currently have `1:3` in the column but we have to change to two columns (?) 1 and 3.
    - Maybe it makes sense to keep these tables (only) as DF, so we keep storing ranges and possibly JuMP variables in them
    - Create a TulipaVariable struct and store the DuckDB table with the indexes (i.e., each row corresponds to a index), and store also a Vector of JuMP variables
    - [x] #898
    - [x] #901
    - [x] #903
    - [x] Create investment variables using the indices in the variable structure
  - [x] #923
    This involves:
    - move the construction from `construct_dataframes` to `compute_variables_indices`
    - change many places that access the variables through `dataframes` to instead unpack `x = model[:x]`, or instead use `variables`. We should try both strategies and see what makes more sense
  - [x] Move JuMP variables out of the DF
  - [x] #909
    Make sure that we differentiate them from the input tables in DuckDB/TulipaIO
- Constraints and expressions
  - [x] #927
  The way that the (balance) constraints work is by asset. The table defines the necessary information for each of these assets. In the current implementation, one of the columns of this table is the JuMP expression with the incoming flow (and one for the output, and possibly more for extra stuff).
  Constructing the incoming and outgoing flows is a huge issue. It was slow and we had to figure out a nice way to do it.
  Furthermore, if we want to use an external table format, we can't save the expressions in the table. So the format will need to be changed.
[...]

PR #803, opened 2024-09-17
Create model discount parameters

### LINKED ISSUE #802: Create economic parameters
Create structure to hold model parameters
Structure should be inside EnergyProblem
Should be read after `connection` is created because we need the data to define defaults.
Either create at `EnergyProblem` constructor or `create_model!`.

gnawin:
# Pull request details

- create a toml file for data on the model level (should be scenario data in the future) only for multi-year
- this file is directly read in `create_model`. Note this means all other cases will use this toml. The discount year will matter to other test cases. Since all other cases have a year of 2030, making a discount year of 2030 would mean doing nothing.
- the generalization of this data and separation of model data and scenario data should be done in the refactor #701 (under _AND THEN_)    

Closes #802 

## Collaboration confirmation

As a contributor I confirm

-   [x] Tests are passing
-   [x] Lint is passing


gnawin:
Hi both @abelsiqueira @datejada, I have this draft PR to hear your opinions. 

The idea is to create two parameters that apply to the model, i.e., which do not depend on the assets. Since we currently don't have a place for that, I just put them in graph-assets-data. Do you think there is a better way to do it now? 

For example, I thought about creating a new file. But since my references are the current files, which look differently from this potential file (just two names with two parameters), I don't yet see a straightforward way to do it.

datejada:
> Hi both @abelsiqueira @datejada, I have this draft PR to hear your opinions.
> 
> The idea is to create two parameters that apply to the model, i.e., which do not depend on the assets. Since we currently don't have a place for that, I just put them in graph-assets-data. Do you think there is a better way to do it now?
> 
> For example, I thought about creating a new file. But since my references are the current files, which look differently from this potential file (just two names with two parameters), I don't yet see a straightforward way to do it.

This proposal is not ideal, we will repeat values unnecessarily. I was looking for inspiration in other models and this data is normally stored in a new entity/file/table with the model properties/constants/parameters.

What is the best way to go forward in Tulipa? I am not sure, but I am leaning toward a separate file (CSV, TOML, YAML, or JSON) to store this information and any other constant that we might want to define or need in the model.

abelsiqueira:
I agree, this should be a separate file. I recommend using TOML:

```toml
model_discount_rate = 0.03
model_discount_year = 2020
```

Since these are parameters, not data, it makes sense that they have a different, simpler, format. Also, that they will be manage outside of TulipaIO.

gnawin:
Thank you both @datejada @abelsiqueira. 

I created a toml file. Since it is outside TulipaIO, so I suppose it's not part of the connection (which currently is tied to `read_csv` in TulipaIO), but rather something else. For that I added an extra argument in `io.jl`. Note that this extra argument will need to added to lots of other places. 

I just want to check with you if this is fine before I make changes accordingly.

abelsiqueira:
It is not part of the connection. Some questions before judging how to handle it:

- Is the file mandatory? Do you need it in all uses of Tulipa? If you don't have one, is there a reasonable default for the parameters?
- Where do you begin to need these parameters? Which function?
- Is it part of the EnergyProblem? Does it make sense to keep these parameters inside it?

gnawin:
@abelsiqueira thanks for thinking along

abelsiqueira:
If it's planned as a temporary solution, I would just open the file directly in `create_model!` and then we discuss these questions in more details later.
Also, I would suggest not creating the file for every test/input, since it's only necessary for multi-year. This way you ensure you're not making it necessary.
Also, it's good to add to #701 to remember that we have to change it afterwards.

gnawin:
@abelsiqueira I made the changes accordingly, but read in the data in create_model instead of create_model!, this way I don't have to add additional arguments for now.

abelsiqueira (changes requested):
That's too hard-coded. It's not conditional on the problem, so it is reading the parameters for all inputs, not only the multi-year. Furthermore, it is reading a fixed file, which is the same as explicitly defining the variables in the file (i.e., hard-code them).
My thinking was you give the file as argument to `create_model`. If nothing is passed, you don't open anything. You would still need to change a few places, but you don't carry anything around.

If you need a quick hack to continue other PRs, you might as well just explicitly define `model_discount_rate` and `model_discount_year` in the `create_model!` function. But then it might be more useful to do it in the specific PR, than do it here.

What is the goal of #802? Enable reading model parameters for general use? Or just create these two parameters somewhere? Or something else, maybe? What is it blocking?

gnawin:
Closed in favor of #813.

PR #807, opened 2024-09-19
Refactor group constraints to improve performance

### LINKED ISSUE #789: Update group constraints according to the investment methods
The group feature needs an update after the changes in the investment methods for the multi-year investment.

Blocked by: #786 and #783

datejada:
# Pull request details

The constraints do not change with the multi-year investment since they are meant to limit the investments that you can make each year rather than limiting the accumulated investments.

The changes here clarify this in the documentation and improve the code's performance.

Closes #789 

## Collaboration confirmation

As a contributor I confirm

-   [x] Tests are passing
-   [x] Lint is passing

### DIFF
--- a/docs/src/formulation.md
+++ b/docs/src/formulation.md
@@ -541,6 +541,8 @@ The following constraints aggregate variables of different assets depending on t
 
 These constraints apply to assets in a group using the investment method $\mathcal{G}^{\text{ai}}$. They help impose an investment potential of a spatial area commonly shared by several assets that can be invested there.
 
+> **Note**: These constraints are applied to the investments each year. The model does not yet have investment limits to a group's accumulated invested capacity.
+
 ##### Minimum Investment Limit of a Group
 
 ```math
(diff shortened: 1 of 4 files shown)


abelsiqueira:
The error in the benchmark run should be fixed after changing `actions/upload-artifact` to use version `4` in the `.github/workflows/benchmark.yml`.

datejada:
@abelsiqueira The benchmark is complaining about a deprecated version. If you show how to update it I can fix that 😄 

I want to run the benchmark in the EU case study because the changes improve the performance of the group constraint. I ran the benchmarks locally in the Norse case, and I got the following:

**Main**: 
- Time for  creating the model (in seconds): 235.7008262

Value of three runs, so avg. 78.6 seconds

**This branch**
- Time for  creating the model (in seconds): 212.0622996

Value of three runs, so avg. 70.7 seconds

datejada:
@abelsiqueira For the EU, there is not much impact, which is expected since there are no groups in that case, and we get a reduction in the time for the Norse, which has groups. So, I think all seems good with the changes.

abelsiqueira (approved):
Thanks. Indeed, it doesn't change much because this part is small in comparison.
Also, the problematic `if` is only the first one in line 17 (inside the `sum` in the `@expression`). The other ones are "normal" ifs. This should still be more efficient, but it is not as terrible as `if`s inside `@expression`s.

PR #814, opened 2024-09-20
Update storage binary charging constraints

### LINKED ISSUE #791: Update binary to avoid simultaneously charging/discharging constraints to consider multi-year investment methods

datejada:
# Pull request details

Update on the constraints for storage assets using the binary method to avoid charging and discharging simultaneously. 

Closes #791 

## Collaboration confirmation

As a contributor I confirm

-   [ ] Tests are passing
-   [x] Lint is passing


gnawin:
@datejada Yes, I see the point, your added compact method for capacity_in is not covered in any tests. And I agree this should be resolved if we add a conversion/storage with compact method in the multi-year case study. I guess there is no need to add different profiles because we can just leave it blank. Would you mind modifying the multi-year case a bit to reflect this? Thanks.

datejada:
> @datejada Yes, I see the point, your added compact method for capacity_in is not covered in any tests. And I agree this should be resolved if we add a conversion/storage with compact method in the multi-year case study. I guess there is no need to add different profiles because we can just use the default. Would you mind modifying the multi-year case a bit to reflect this? Thanks.

Perfect! Thanks for the reply. I will add it to the case study

src/constraints/capacity.jl
```diff
@@ -139,22 +135,25 @@ function add_capacity_constraints!(
...
+                            ("availability", row.rep_period),
+                            row.timesteps_block,
+                            1.0,
+                        ) *
+                        accumulated_units_compact_method[accumulated_set_using_compact_method_lookup[(
+                            row.asset,
+                            row.year,
+                            v,
+                        )]] for v in V_all if
+                        (row.asset, row.year, v) in accumulated_set_using_compact_method
```
abelsiqueira, line 156: This conditional is inside the `sum`, so it might be slow.
The solution would be defining something like

```
some_set[a, y] = [v for v in V_all if (a, y, v) in accumulated_set_using_compact_method]
```

And looping over `some_set[row.asset, row.year]`.
This might be affected by the refactor, but since this array of expression is already defined as an array of expressions looped over the rows of a table, maybe this code won't be improved simply by doing the refactor.

abelsiqueira (approved):
Thanks for the PR.
I made a comment over a loop inside a sum inside an expression. It's not blocking for me, we can update that after the refactor, so I'm approving.

src/constraints/capacity.jl
datejada: Good point to consider in the refactor and then see how we tackle this. Thanks!

Issue #806, opened 2024-09-18
Add the simple method for storage assets


datejada:
Adapt storage assets to have the simple method considering:

- [x] #820
- [x] #822
- [x] #823
- [x] #824
- [ ] #837
- [x] Update objective function coefficients for the investment in energy (checked in favor of #835 )

PR #823, opened 2024-09-24
Update energy storage constraints with multi-year investment

datejada:
# Pull request details

Update the energy storage limit to consider the accumulated energy investment in the multi-year development. The storage assets will only use the simple method.

Part of #806 

## Collaboration confirmation

As a contributor I confirm

-   [x] Tests are passing
-   [x] Lint is passing

### DIFF
--- a/src/constraints/capacity.jl
+++ b/src/constraints/capacity.jl
@@ -135,43 +135,20 @@ function add_capacity_constraints!(
     # - Create capacity limit for incoming flows
     assets_profile_times_capacity_in =
         model[:assets_profile_times_capacity_in] = [
-            if row.asset ∈ decommissionable_assets_using_compact_method
-                @expression(
-                    model,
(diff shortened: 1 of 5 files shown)


abelsiqueira (approved):
LGTM. Just as a repeated note, we're adding a lot of `sum(... if ..)` inside expressions, which is slower. It's not terrible, but it's something to keep in mind. And as a reminder to ourselves, we plan to remove these during/after the refactor.

datejada:
> LGTM. Just as a repeated note, we're adding a lot of `sum(... if ..)` inside expressions, which is slower. It's not terrible, but it's something to keep in mind. And as a reminder to ourselves, we plan to remove these during/after the refactor.

I 100% agree; I also want to get rid of those combinations of `sum(... if ..)`. I am adding that to the list of refactoring tasks. Thanks!

PR #829, opened 2024-09-26
Add simple method transport investment

### LINKED ISSUE #792: Add the simple method for investments in transport assets

gnawin:
# Pull request details

Closes #792 

## Collaboration confirmation

As a contributor I confirm

-   [x] Tests are passing
-   [x] Lint is passing


src/create-model.jl
```diff
@@ -625,7 +625,13 @@ function create_model(
...
         Ah  = filter_graph(graph, A, "hub", :type)
         Acv = filter_graph(graph, A, "conversion", :type)
-        Ft  = Dict(y => filter_graph(graph, F, true, :is_transport, y) for y in Y)
+        Ft  = filter_graph(graph, F, true, :is_transport)
+        # Ft  = Dict(y => filter_graph(graph, F, true, :is_transport, y) for y in Y) # do I need the year index?
+
+        # Ft = filter(!isempty, collect(values(Ft))) |> collect
+        # if !isempty(Ft)
+        #     Ft = Ft[1]
+        # end
```
abelsiqueira, line 634:
```suggestion
```
This can be removed

src/create-model.jl
```diff
@@ -1243,6 +1284,18 @@ function create_model(
...
+            sum(
+                graph[u, v].fixed_cost[y] / 2 *
+                graph[u, v].capacity *
+                (
+                    accumulated_flows_export_units[y, (u, v)] +
+                    accumulated_flows_import_units[y, (u, v)]
+                ) for y in Y for (u, v) in Fi[y]
+            )
+        )
+
```
abelsiqueira, line 1298: This doesn't appear to be used anywhere

src/constraints/transport.jl
```diff
@@ -6,30 +6,22 @@ add_transport_constraints!(model, graph, df_flows, flow, Ft, flows_investment)
...
-                    1.0,
-                ) *
-                graph[row.from, row.to].capacity *
-                (
-                    graph[row.from, row.to].initial_export_units[row.year] +
-                    flows_investment[row.year, (row.from, row.to)]
-                )
-            )
-        else
+        if (row.from, row.to) in Ft
```
abelsiqueira, line 24: With this change `upper_bound_transport_flow` and `lower_...` now have a bunch of empty because the `if` doesn't have an else.
I still have to understand where this is used, but in principle you could change the `for row in ...` to add a `if (row.from, row.to) in Ft` at the end. This will make the `upper_....` vector limited to that size only.

abelsiqueira: So it looks like this expression has length `length(df_flows)`, which means that if we change the loop to `for row in .... if ... in Ft`, the size will be wrong.

However, in the next section you define `df` which is exactly the filter that we want.

So it looks like you could move the `df` computation to above this and redefine this expression as a loop over the rows of `df`.

Later, when you access this expression via `row.index`, it will break because the index is not correct anymore.
So, instead of that index, we actually want the indexes of `df`. So we can change to something like

```Julia
...
   this_expression[idx]
...
   for (idx, row) in enumerate(eachrow(df))
```

Alternatively, you can `zip` and loop over the values of both:

```Julia
...
    element
...
    for (row, element) in zip(eachrow(df), this_expression)
```

I hope this wasn't too confusing

abelsiqueira (changes requested):
Hi both, thanks for the PR.
I've made a few suggestions that I hope aren't too confusing.
I also noticed that `flows_fixed_cost` doesn't appear to be used anywhere. Is that the case?

src/create-model.jl
gnawin: Good catch! It should be in the obj.

src/constraints/transport.jl
gnawin: Thanks! I think I got it!

gnawin:
@abelsiqueira thanks for the review! I updated accordingly to my understanding. Does it now do what you meant?

abelsiqueira (approved):
LGTM

PR #821, opened 2024-09-24
Add necessary data to prepare multi-year transport

gnawin:
# Pull request details

## Describe the changes made in this pull request

- Add missing parameters to graph-flows-data.csv: technical_lifetime, economic_lifetime, discount_rate,
- Add file vintage-flows-data with: from_asset, to_asset, commission_year, fixed_cost, investment_cost
- Change `initial_export_capacity` to `initial_export_units`, `initial_import_capacity` to `initial_import_units` 

## List of related issues or pull requests

Part of #792 

## Collaboration confirmation

As a contributor I confirm

-   [ ] I read and followed the instructions in README.dev.md
-   [ ] The documentation is up to date with the changes introduced in this Pull Request (or NA)
-   [ ] Tests are passing
-   [ ] Lint is passing

Issue #461, opened 2024-02-12
Implement data validation


clizbe:
### Description

Please edit this description with new bullets as you think of them:

- [ ] (Outdated?) If GraphAssetData.investable is an array of years, investment_cost should also be an array of equal length. (After multi-year investment is implemented.)
- [ ] #1100
- [x] #1098
- [x] #1101
- [ ] (Outdated?) If a storage asset has a missing is_seasonal, the default is `false` (non-seasonal storage).
- [ ] #1103
- [ ] (Outdated?) If storage inflows are missing, then assign the default = 0
- [ ] #1104
- [x] #437 
- [x] #453 
- [ ] https://github.com/TulipaEnergy/TulipaIO.jl/issues/78
- [x] #1105
- [x] #1109
- [x] #1110
- [x] #1111
- [x] #1112
- [x] #1091
- [ ] Validate `storage_loss_from_stored_energy` stays between 0 and strictly less than 1
- [ ] Consider/fix the interaction of the schemas arg and populate_defaults related to this bug: #1171 
- [ ] Validate or allow missing row in asset-commission
   If an asset exists in `asset` but has no corresponding row in `asset-commission`, the code errors with a cryptic error related to multiplying `Missing` and `JuMP.AffExpr` in the objective. We should either allow the missing row or validate that they don't happen.
   This was uncovered with TulipaBuilder.jl, so steps to reproduced are missing for now.
- [ ] Validate or allow missing row in asset-milestone
   If an asset exists in `asset` but has no corresponding row in `asset-milestone`, the code errors with a cryptic error related to multiplying `Missing` and `JuMP.AffExpr` in the objective. We should either allow the missing row or validate that they don't happen.
   This was uncovered with TulipaBuilder.jl, so steps to reproduced are missing for now.
- [ ] Validate or allow missing row in flow-commission
   If a flow exists in `flow` but has no corresopnding row in `flow-commission`, the code errors with a cryptic error related to converting `Missing` to `Float64` in `model-preparation.jl`. We should either allow the missing row or validate that they don't happen.
   This was uncovered with TulipaBuilder.jl, so steps to reproduced are missing for now.
- [ ] Validate or allow missing row in flow-milestone
   If a flow exists in `flow` but has no corresponding row in `flow-milestone`, the code errors with a cryptic error related to multiplying `Missing` and `JuMP.AffExpr` in the objective. We should either allow the missing row or validate that they don't happen.
[...]

PR #832, opened 2024-09-27
Change flow data to make multi-year case feasible

### LINKED ISSUE #831: Make multi-year test feasible

gnawin:
# Pull request details

Multi-year becomes unbounded after #821, which is about changing flow data. The flow investment cost has been read incorrectly as 0, so the variable flow investment becomes unbounded.

The PR reads in the correct data. 

Closes #831 

## Collaboration confirmation

As a contributor I confirm

-   [ ] Tests are passing
-   [ ] Lint is passing


gnawin:
In addition to the description above, I also added an upper limit for flow investment. This is because without the upper limit, the `solve_model!` takes forever. But it works fine with it.

I attach two lp files, one with the upper limit, and the other without. The only difference between the two cases is the upper limit.

Any ideas why? @abelsiqueira @datejada 

[change-investment-cost-add-upper-bound-can-solve.txt](https://github.com/user-attachments/files/17164239/change-investment-cost-add-upper-bound-can-solve.txt)
[change-investment-cost-cannot-solve.txt](https://github.com/user-attachments/files/17164240/change-investment-cost-cannot-solve.txt)
Screenshot 2024-09-27 143802

datejada:
> In addition to the description above, I also added an upper limit for flow investment. This is because without the upper limit, the `solve_model!` takes forever. But it works fine with it.
> 
> 
> 
> I attach two lp files (I manually changed the file type from `.lp` to `.txt` so I can upload them), one with the upper limit, and the other without. The only difference between the two cases is the upper limit.
> 
> 
> 
> Any ideas why? @abelsiqueira @datejada 
> 
> 
> 
> [change-investment-cost-add-upper-bound-can-solve.txt](https://github.com/user-attachments/files/17164239/change-investment-cost-add-upper-bound-can-solve.txt)
> 
> [change-investment-cost-cannot-solve.txt](https://github.com/user-attachments/files/17164240/change-investment-cost-cannot-solve.txt)
> 
> Screenshot 2024-09-27 143802
> 
> 

Hi @gnawin, thanks for the files, I have downloaded them and did some tests locally using Gurobi. The problem is unbounded when you don't have an upper bound in the investment, but HiGHS doesn't identify that and tries to solve it, and then most probably getting into a bug or so such that it is not able to solve it and it looks that it takes forever. I will report it to the HiGHS team. Anyway, the data must have an investment limit. This check should be added to the issue #461 to avoid this problem in the future. Would you mind adding it?

datejada (commented):
The flow variable is free for transport assets. So, we need the constraint that makes the accumulated units non-negative. I will talk to German tomorrow, but it always feels safer to add it.

datejada (approved):

PR #842, opened 2024-10-07
Limit DuckDB version to 1.0 until duckdb/duckdb#13911 is fixed

abelsiqueira:
DuckDB.jl 1.1 on Windows is broken with an error
'Could not load symbol duckdb_vector_size'
This limits the DuckDB version until that is fixed.

refs:
- duckdb/duckdb#13911


datejada (approved):
Thank you!

PR #843, opened 2024-10-07
DuckDB_jll needs explicit compat bounds

abelsiqueira:
DuckDB 1.0 allows DuckDB_jll 1.1.0, which is not correct.
This limits DuckDB_jll as well.

refs:
- duckdb/duckdb#13911


datejada (approved):
Great! Thanks

Issue #676, opened 2024-06-27
Apply BestieTemplate.jl


abelsiqueira:
### What and Why

The [BestieTemplate.jl](https://github.com/abelsiqueira/BestieTemplate.jl) was created using the best practices that we introduced in this package, so for many parts, this package is up-to-date in those practices.
However, since it was created, there were improvements in the BestieTemplate that we should apply here and the easier way to do it is by adopting the template.

The template was applied to [TulipaIO](https://github.com/TulipaEnergy/TulipaIO.jl) recently (in two parts, check the merged PRs). It was named COPIERTemplate.jl at the moment.

Some of the things:

- Link checker was removed from pre-commit and moved to a separate Linting with a different faster package (lychee)
- Dependabot was added (#444)
- An action checking for updates of the pre-commit versions was added
- It helps with this checklist (#630)
- Better solution for testing on the PR 
- Maybe [All Contributors](https://allcontributors.org)

If we do decide to apply it, we should wait until at least Monday, when the Registration of the rename will be done.

### Possible Drawbacks

There are different choices, such as the issue and PR templates are different, the formatting is slightly different, and the README.dev and CONTRIBUTING files moved to the docs.

It takes some time to apply it the first time, since you have to fix the conflicts (though I can help with that).

I am biased, since I'm lead on the project.

clizbe:
Sounds good to me. Of course you're biased, but so far drinking your Kool-Aid has worked out well for us. :)

What do you mean by the README.dev and CONTRIBUTING files moved to the docs?

PR #846, opened 2024-10-08
Specify the domain of decommission variables

### LINKED ISSUE #828: Add checks to the domain of decommission variables and make sure they only exist when there are existing capacities or investments - compact method.
~This should avoid making the multi-year case study unbounded~ (see #831)

Best to alter the structure of code as little as possible (so #833 is not a good way to go), but allow more repetitions if needed.

gnawin:
# Pull request details

Closes #828 

## Collaboration confirmation

As a contributor I confirm

-   [ ] Tests are passing
-   [ ] Lint is passing

### DIFF
--- a/src/TulipaEnergyModel.jl
+++ b/src/TulipaEnergyModel.jl
@@ -38,7 +38,8 @@ include("time-resolution.jl")
 include("economic-parameters.jl")
 
 constraints_folder_path = joinpath(@__DIR__, "constraints")
-for file in filter(f -> endswith(f, ".jl"), readdir(constraints_folder_path))
+constraints_files = filter(f -> endswith(f, ".jl"), readdir(constraints_folder_path))
+for file in constraints_files
     include(joinpath(constraints_folder_path, file))
 end
 
--- a/src/TulipaEnergyModel.jl
+++ b/src/TulipaEnergyModel.jl
@@ -38,8 +38,7 @@ include("time-resolution.jl")
 include("economic-parameters.jl")
 
 constraints_folder_path = joinpath(@__DIR__, "constraints")
-constraints_files = filter(f -> endswith(f, ".jl"), readdir(constraints_folder_path))
-for file in constraints_files
+for file in filter(f -> endswith(f, ".jl"), readdir(constraints_folder_path))
     include(joinpath(constraints_folder_path, file))
 end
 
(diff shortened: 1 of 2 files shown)


gnawin:
@abelsiqueira @datejada It is more or less straightforward to do this for the decommission variable of the compact method, so I did this before the refactor. 

The simple method is a bit more complex because it is indexed differently, i.e., not using an indice set yet, but it will be the same idea (and the conditions are very similar as well). Since we'll refactor, I did not do it for the simple method and rather do it in the factor.

gnawin:
@abelsiqueira both the link and the codecov/project fails are confusing to me, could you have a look? Thanks!

abelsiqueira:
For completeness, repeating some thigns from the chat here.

The link check failure is probably due to changes in the treatment of automated link checkers by the end points. Recommended solution is to ignore at the moment, and check after we apply Bestie (#676). If the error persists we can add it to the ignore list. This is safe to do for this website at least. We can think again for the others. 
On a side note, it is actually possible to add it to the ignore list at `.markdown-link-check.json` right now, but we would need to go check how to run it locally

The Codecov error is probably because version "1" points to "1.11" now (https://github.com/JuliaLang/www.julialang.org/pull/2160) and that affected how the coverage is processed.
Possible solutions:
- Set a range of acceptable coverage instead of the 100% goal - it will be harder and harder to keep it at 100%, so this might be an incentive and the right time to do it
- Change the `for` and remove the intermediary variable `constraints_files`.

abelsiqueira (approved):
LGTM

gnawin:
> For completeness, repeating some thigns from the chat here.
> 
> The link check failure is probably due to changes in the treatment of automated link checkers by the end points. Recommended solution is to ignore at the moment, and check after we apply Bestie (#676). If the error persists we can add it to the ignore list. This is safe to do for this website at least. We can think again for the others. On a side note, it is actually possible to add it to the ignore list at `.markdown-link-check.json` right now, but we would need to go check how to run it locally
> 
> The Codecov error is probably because version "1" points to "1.11" now ([JuliaLang/www.julialang.org#2160](https://github.com/JuliaLang/www.julialang.org/pull/2160)) and that affected how the coverage is processed. Possible solutions:
> 
> * Set a range of acceptable coverage instead of the 100% goal - it will be harder and harder to keep it at 100%, so this might be an incentive and the right time to do it
> * Change the `for` and remove the intermediary variable `constraints_files`.

After some searching, it seems that I have to use an inline comment, but it does not work 
```
<!-- markdownlint-disable-next-line -->
[Visit this link](https://code.visualstudio.com)
```
Is this what I should do?

PR #876, opened 2024-10-14
Change lower bound and test value

### LINKED ISSUE #844: The model can be unbounded under certain set-ups
### What and Why

From the multi-year case, it is found that the model can be unbounded if a flow is used for both production and transport. 

Image a user case of a wind-electrolyzer combination, where we want to invest in a offshore wind turbine, and its production goes exclusively to a electrolyzer/demand onshore. In this case, we can use one flow variable for that, i.e., flow wind-demand. Since this is flow also a transport asset, it, in principle, can go backwards (from demand to wind). But in this case, we want unidirectional flow.

- If no investments in transport are modeled, this can be coped by setting up the initial import capacity to 0.  
- If we want to model investments as well, then the investments will be for two directions, and we will have reverse flows, i.e.,  negative flow wind-demand. 
Then It becomes a problem because the production flow will be unbounded (since it is a transport and currently we set transport flows free), so that it can go to minus infinity, unless there is a given bound for the transport which prevents the flow from going to minus infinity.

We can always avoid this, if we use a hub in-between, but it creates "unnecessary" constraints.

Points to discussion:
- do we want to support such a use case?
  - If not, we need to add a validation.
  - If yes, the need to see how to change the model.

I have discussed this with @datejada, and we would like to discuss with everyone. It's more of a modeling choice so @g-moralesespana @clizbe. 

### Possible Drawbacks

_No response_

### Related Issues

_No response_

gnawin:
## Related issues

Closes #844 

## Checklist

- [x] Tests are passing


datejada (approved):
LGTM

Issue #885, opened 2024-10-16
Create all data and tables necessary for the model before the  function


abelsiqueira:
We can control and improve both the creation of the data/tables and the create_model function better if these concerns are separated.
This includes the "expression" data, e.g., data that informs the incoming and outgoing flow.
This includes clustering stuff (representative periods, timeframe, years, etc.)
Partition should also be created before the model, though we are missing the "scenario specification" data as a separate entity

- [ ] Clean up `create_model`. Create files and functions to make it look nicer
- [ ] Move dataframe updates to before create_model if possible. Mark places that are waiting for #884 
- [x] Sets can be created beforehand as well

abelsiqueira:
Except for moving the expressions to before creation of the model, this is done. The feasibility of changing the expressions will be discussed at another point.

PR #886, opened 2024-10-16
Refactor some of the code related to model creation

abelsiqueira:
- Create new files src/utils.jl and src/model-preparation.jl to help split the code from src/create-model.jl.
- The file src/utils.jl contains useful functions for create_model.
- The file src/model-preparation should be useful in the future for data and structure creation that should happen before the model.
- Create a simple sets NamedTuple for this section.
- Create function add_expressions_to_dataframe! with a large chunk of code from create_model that adds incoming and outgoing flows to some dataframes.
- Create function create_variables! that creates the variables in create_model.
- Both of these functions actually involve things happening inside create_model, but for now we are keeping them there.

---

- Move the sets creation to a function create_sets.
- Store the sets in a NamedTuple


abelsiqueira:
This doesn't close #885, but it can be merged

datejada (approved):
LGTM. Thanks!

PR #887, opened 2024-10-16
Create TulipaVariable struct and define variable tables

### LINKED ISSUE #884: Define variable tables

datejada:
## Related issues

Closes #884

## Checklist

- [x] Tests are passing


abelsiqueira (approved):
Took me a long time to realise what was happening with the profile.
`create_model` allocates -0.5 GB (which is nothing in comparison to the 22 GB), so it improves only 2%.
`input` allocates +0.5 GB (which is a lot in comparison to the 0.7 GB), so it worsens by 40%.

In practice, though, it is the same. Something to keep in mind for future benchmarks

PR #889, opened 2024-10-17
Multi year docs

gnawin:
## Related issues

Closes #819 

## Checklist

- [x] I am following the [contributing guidelines](https://github.com/TulipaEnergy/TulipaEnergyModel.jl/blob/main/docs/src/90-contributing.md)

- [x] Tests are passing
- [x] Lint workflow is passing
- [ ] Docs were updated and workflow is passing

PR #894, opened 2024-10-21
Fix index

gnawin:
## Related issues

Small fix. 
Related to the docs of multi-year #889. 

## Checklist

- [x] Tests are passing


datejada:
> Hi @datejada thanks for adding the missing parts and merging #889. I added the y index for the set (now almost everything has a year index...).
> 
> And if you have time, please go through the new docs on the **storage**, because I'm not sure if I documented them 100% correctly, especially the **storage balance** https://tulipaenergy.github.io/TulipaEnergyModel.jl/dev/40-formulation/#Constraints-for-Energy-Storage-Assets. I think I got lost a bit by the code of the energy limit.
> 
> Thanks!

Thanks for the PR. No worries for the formulation of the storage, it is okay. During the refactor we are cleaning the code so it is easier to follow and then I will change the formulation if needed (which I don't think it is necessary).

datejada (approved):

PR #900, opened 2024-10-22
Create flow variables using the indices table

### LINKED ISSUE #898: Create flow variables using the indices in the variable structure

datejada:
## Related issues

Closes #898

## Checklist

- [x] Tests are passing

### DIFF
--- a/src/constraints/capacity.jl
+++ b/src/constraints/capacity.jl
@@ -299,15 +299,13 @@ function add_capacity_constraints!(
     ]
 
     # - Lower limit for flows associated with assets
-    assets_with_non_negative_outgoing_flows_indices =
-        filter(:from => in(Ap ∪ Acv ∪ As), flows_indices; view = true)
-    for row in eachrow(assets_with_non_negative_outgoing_flows_indices)
-        JuMP.set_lower_bound(flow[row.index], 0.0)
-    end
-
-    assets_with_non_negative_incoming_flows_indices =
-        filter(:to => in(Acv ∪ As), flows_indices; view = true)
-    for row in eachrow(assets_with_non_negative_incoming_flows_indices)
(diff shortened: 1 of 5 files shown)


src/constraints/capacity.jl
```diff
@@ -296,12 +299,15 @@ function add_capacity_constraints!(
...
+    assets_with_non_negative_outgoing_flows_indices =
+        filter(:from => in(Ap ∪ Acv ∪ As), flows_indices; view = true)
+    for row in eachrow(assets_with_non_negative_outgoing_flows_indices)
+        JuMP.set_lower_bound(flow[row.index], 0.0)
+    end
+
+    assets_with_non_negative_incoming_flows_indices =
+        filter(:to => in(Acv ∪ As), flows_indices; view = true)
+    for row in eachrow(assets_with_non_negative_incoming_flows_indices)
+        JuMP.set_lower_bound(flow[row.index], 0.0)
```
abelsiqueira, line 311: You should be able to write this with a single filter (or DataFrames.subset, I never know which is best)

```julia
df_with_good_name = DataFrames.subset(
  flows_indices,
  [:from, :to] => ByRow((from, to) -> from in Ap || from in Acv || from in As || to in Acv || to in As),
)
```

I've separated the union into different `in` checks, because the `union` creates a temporary set (AFAIK).

abelsiqueira (changes requested):
I made a small suggestion that might affect efficiency. Since the loop is made for a lot of flow variables, it might make a difference. I tagged to run the benchmark. We can check both versions.

datejada:
@abelsiqueira Thanks! I have included the change in the new version of the code

abelsiqueira (approved):
Thanks

PR #902, opened 2024-10-22
Create units_on variables using the indices table

### LINKED ISSUE #901: Create units on variables using the indices in the variable structure

datejada:
## Related issues

Closes #901

## Checklist

- [x] Tests are passing

### DIFF
--- a/src/variables/unit-commitment.jl
+++ b/src/variables/unit-commitment.jl
@@ -1,31 +1,31 @@
 export add_unit_commitment_variables!
 
 """
-    add_unit_commitment_variables!(model, dataframes, sets)
+    add_unit_commitment_variables!(model, ...)
 
-Adds unit commitment variables to the optimization `model` based on the `:units_on` dataframe.
+Adds unit commitment variables to the optimization `model` based on the `:units_on` indices.
 Additionally, variables are constrained to be integers based on the `sets` structure.
 
 """
-function add_unit_commitment_variables!(model, dataframes, sets)
-    df_units_on = dataframes[:units_on]
+function add_unit_commitment_variables!(model, sets, variables)
+    units_on_indices = variables[:units_on].indices
 
-    model[:units_on] =
-        df_units_on.units_on = [
-            @variable(
-                model,
-                lower_bound = 0.0,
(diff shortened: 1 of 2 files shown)


src/variables/unit-commitment.jl
```diff
@@ -1,31 +1,31 @@
...
+    ]
 
     ### Integer Unit Commitment Variables
-    for row in eachrow(df_units_on)
-        if !(row.asset in sets.Auc_integer[row.year])
-            continue
-        end
-
-        JuMP.set_integer(model[:units_on][row.index])
+    integer_units_on_indices = filter(
```
datejada, line 22: The following code is more efficient, but we need to ensure that the indices dataframe has the correct columns, which doesn't happen when there is no assets using the unit commitment. If we ensure that the indices table always has the required columns, no matter if the rows are zero, then we can use this code instead of the current changes (using filter).

```suggestion
    integer_units_on_indices = DataFrames.subset(
        units_on_indices,
        [:asset, :year] => DataFrames.ByRow((a, y) -> a in sets.Auc_integer[y]),
    )
```

src/variables/unit-commitment.jl
```diff
@@ -1,31 +1,31 @@
...
+        row ->
+            (row.year in keys(sets.Auc_integer)) && (row.asset in sets.Auc_integer[row.year]),
+        units_on_indices,
+    )
```
datejada, line 26: The following code is more efficient, but we need to ensure that the indices dataframe has the correct columns, which doesn't happen when no assets are using the unit commitment. If we ensure that the indices table always has the required columns, no matter if the rows are zero, we can use this code instead of the current changes (using the filter function).

```suggestion
    integer_units_on_indices = DataFrames.subset(
        units_on_indices,
        [:asset, :year] => DataFrames.ByRow((a, y) -> a in sets.Auc_integer[y]),
    )
```

datejada: Yes, it does. I agree that having some extra columns in the indices will help to filter them (e.g., assets with integer unit commitment) using SQL queries in DB. So, let's keep the filter for now, since it will change once we move the indices to DB.
Thanks!

abelsiqueira (approved):
Great, thanks!

PR #917, opened 2024-10-28
Add `highest_out` to `create_constraints_indices`

### LINKED ISSUE #914: Create constraints indices for `highest_out`
### Description

Similar to what has been done in #896.

gnawin:
## Related issues

Closes #914 

## Checklist

- [x] Tests are passing


gnawin:
@abelsiqueira could you help on this? thanks.

src/tmp.jl
```diff
@@ -231,6 +231,54 @@ function tmp_create_constraints_indices(connection)
...
+    # FROM asset_time_resolution
+    # UNION
+    DuckDB.execute(
+        connection,
+        "CREATE OR REPLACE TABLE cons_indices_highest_out AS
+        SELECT
+            main.asset,
+            main.year,
+            main.rep_period,
+            sub.time_block_start,
```
abelsiqueira, line 267:
```suggestion
            COALESCE(sub.time_block_start, 1) AS time_block_start,
```
I think this is the same issue that I had in #919 but for the benchmark tests, and that now is fixed in #920. The problem is that there are no outgoing flows (in this case), so `sub.time.block_start` doesn't exist. By using `COALESCE`, we make sure that if it doesn't exist, we use the range `1:num_timesteps`

gnawin:
@abelsiqueira the tests indeed passed, but the link checker complains. Could you have a look? Or pause/close this. Thank you!

abelsiqueira (approved):
LGTM. Benchmarks are not running because it's a fork, but should be a similar efficiency to previous PR (slight decrease with possibilities of improvements).

PR #920, opened 2024-10-29
Fix case where the there are no time blocks to determine in a constraint

abelsiqueira:
For FI_Hydro_Reservoir in the EU case, there are no matches in the incoming
flows, so there is nothing constraining the time block. In this case, it uses
1 as time_block_start and num_timesteps and tmie_block_end of the single
block.


clizbe (approved):
Good old coalesce!
LGTM

Issue #923, opened 2024-11-07
Refactor variables


abelsiqueira:
A Variable should have two things: an indices table and JuMP vector.
Currently, many variables are still in the old format (see `src/variables/investments.jl`).
To solve this, we have to:

- Create variables indices in `compute_variables_indices` for all variables
- Directly use `connection` instead of `dataframes` to construct indices
- Create JuMP variables from the indices for all variables (which means updating the whole model)

Expected challenges:

- Some sets are heavily dependent on other sets, so the logic needs to be translated to DuckDB tables
- Changing the `@variable` definition to use the table requires:
  - Changing how some values are computed (e.g., `weight_for_assets_investment_discounts`) to use DuckDB tables as well
  - Changing all places that use the variable to loop over tables

Issue #684, opened 2024-07-01
Create a meaningful year data for the Norse Case


datejada:
### What and Why

This is to make the pipeline more meaningful

PR #925, opened 2024-11-08
Refactor variables to use TulipaVariable

abelsiqueira:
Create field lookup to allow changing the variables without
having to redefine the whole model creation

Related to #923

### DIFF
--- a/src/constraints/investment.jl
+++ b/src/constraints/investment.jl
@@ -1,23 +1,24 @@
 export add_investment_constraints!
 
 """
-add_investment_constraints!(graph, Ai, Ase, Fi, assets_investment, assets_investment_energy, flows_investment)
+    add_investment_constraints!(graph, Ai, Ase, Fi, assets_investment, assets_investment_energy, flows_investment)
 
 Adds the investment constraints for all asset types and transport flows to the model
 """
-
-function add_investment_constraints!(
-    graph,
-    sets,
-    assets_investment,
-    assets_investment_energy,
-    flows_investment,
-)
+function add_investment_constraints!(graph, sets, variables)
+    # TODO: Since this function is defining bound constraints, it doesn't need the `model`
+    # When we refactor the signatures to look the same, we should consider naming it differently
+    # TODO: Verify if it's possible and reasonable to move the bound definition to when the
+    # indices are created
     # unpack from sets
     Ai = sets[:Ai]
     Ase = sets[:Ase]
     Fi = sets[:Fi]
     Y = sets[:Y]
+
+    assets_investment = variables[:assets_investment].lookup
+    assets_investment_energy = variables[:assets_investment_energy].lookup
+    flows_investment = variables[:flows_investment].lookup
 
     # - Maximum (i.e., potential) investment limit for assets
     for y in Y, a in Ai[y]
--- a/src/variables/create.jl
+++ b/src/variables/create.jl
@@ -1,0 +1,137 @@
+export compute_variables_indices
+
+# TODO: Allow changing table names to make unit tests possible
+# The signature should be something like `...(connection; assets_data="t_assets_data", ...)`
+function compute_variables_indices(connection, dataframes)
+    variables = Dict(
+        :flow => TulipaVariable(dataframes[:flows]),
+        :units_on => TulipaVariable(dataframes[:units_on]),
+        :storage_level_intra_rp => TulipaVariable(dataframes[:storage_level_intra_rp]),
+        :storage_level_inter_rp => TulipaVariable(dataframes[:storage_level_inter_rp]),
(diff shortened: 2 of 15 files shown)


src/variables/create.jl
```diff
@@ -0,0 +1,79 @@
+export compute_variables_indices
+
+function compute_variables_indices(connection, dataframes)
```
abelsiqueira, line 3: Add "TODO: Allow changing table names to make unit tests possible."
Something like `compute_variables_indices(connection, assets_data="t_assets_data")`

src/constraints/investment.jl
```diff
@@ -6,19 +6,17 @@ add_investment_constraints!(graph, Ai, Ase, Fi, assets_investment, assets_invest
...
 """
 
-function add_investment_constraints!(
-    graph,
-    sets,
-    assets_investment,
-    assets_investment_energy,
-    flows_investment,
-)
+function add_investment_constraints!(graph, sets, variables)
```
abelsiqueira, line 9: Add "TODO: Differentiate the type of constraint creation (bound vs not bound)"
Add "TODO: Verify if it's possible to move the bound to the indices creation"

src/variables/create.jl
```diff
@@ -53,23 +53,55 @@ function compute_variables_indices(connection, dataframes)
...
-    #             AND assets_data.commission_year=commission_year_data.commission_year
-    #         WHERE
-    #             investment_method='compact'
-    #             AND assets_data.year - graph_assets_data.technical-lifetime + 1 <= assets_data.commission_year < assets_data.year
-    #             AND ass
-    #             ",
-    #     ) |> DataFrame,
-    # )
+    # For the compact method, not all combinations of (assets, year, commission_year) exist
+    # In particular, some of the commission years for which SOMETHING  TODO: for which what, again?
```
gnawin, line 59:
```suggestion
    # In particular, if the commission years are also milestone years where the asset is investable. 
```

src/variables/create.jl
abelsiqueira: Done

src/constraints/investment.jl
abelsiqueira: Done

src/variables/create.jl
abelsiqueira: Thank you

src/variables/create.jl
```diff
@@ -0,0 +1,156 @@
...
+
+    # For the compact method, not all combinations of (assets, year, commission_year) exist
+    # In particular, if the commission years are also milestone years where the asset is investable. 
+    # which means that we use the UNION to add them
+    # We only filter afterwards to increase readability, but it would be technically more
+    # efficient to filter for each inner query before computing the union
+    # NOTE: To add another column from assets_data, it has to be added on the two inner SELECT
+    # statements (not the year_data one)
+    variables[:assets_decommission_compact_method] = TulipaVariable(
+        DuckDB.query(
```
datejada, line 66: Thanks @abelsiqueira, @gnawin and I reviewed this query and it is almost there. So, two things:

1. We tested it in the multi-year investment data, and it works. However, we want to cover another edge case outside that use case (so, in issue #684 I added that to the list to improve it). If we change the input to have the asset = ccgt in year = 2030 and commission_year = 2030 with investable = false, then it should not appear in the result of this query. As it is, it seems. Ni and I tried to fix it but couldn't (sorry about that). If we cover that case, then this change is okay.
2. For the indices of the decommission variables, we don't need the investable column (we need the investment integer for creating the variable). So, we were wondering, why do you keep it? Maybe we are missing something. If we don't need it, then we should take it out.

datejada (changes requested):
Hi @abelsiqueira, thanks for the PR. We have a comment on the decommission variable indices. Please let us know if you need to discuss it, and we can comment on Teams.

src/variables/create.jl
abelsiqueira: 1. Can we add that to the actual tests? Otherwise there is guarantee that it works as expected. 
   I also don't see the relation between this new data row and the compact formulation. There are other rows that result in `investable = false` in the query, so `investable` is not taken into account

2. I was just using for checking that things propagate

src/variables/create.jl
```diff
@@ -0,0 +1,137 @@
...
+            LEFT JOIN flow
+                ON flow.from_asset = flow_milestone.from_asset
+                AND flow.to_asset = flow_milestone.to_asset
+            WHERE
+                flow_milestone.investable = true",
+        ) |> DataFrame,
+    )
+    dataframes[:flows_investment] = variables[:flows_investment].indices
+
+    # TODO: Verify that year=commission_year is the correct approach. Alternative is creating a table where (name, year) is the key
```
abelsiqueira, line 32:
```suggestion
```

PR #926, opened 2024-11-09
Create all variables directly from connection and independent of dataframes

### LINKED ISSUE #923: Refactor variables

abelsiqueira:
**Blocked by #925**

Create all variables with connection instead of dataframes variable.

With the `lookup` created in #925, this allow us to

Close #923

### DIFF
--- a/src/tmp.jl
+++ b/src/tmp.jl
@@ -11,7 +11,9 @@ function _append_given_durations(appender, row, durations)
             DuckDB.append(appender, row.to_asset)
         end
         DuckDB.append(appender, row.year)
-        DuckDB.append(appender, row.rep_period)
+        if haskey(row, :rep_period)
+            DuckDB.append(appender, row.rep_period)
+        end
         if haskey(row, :efficiency)
             DuckDB.append(appender, row.efficiency)
         end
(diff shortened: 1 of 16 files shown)


src/tmp.jl
```diff
@@ -157,6 +159,148 @@ function tmp_create_partition_tables(connection)
...
+    DuckDB.end_row(appender)
+end
+
+function tmp_create_lowest_resolution_table(connection)
+    # These are generic tables to be used to create some variables and constraints
+
+    # The logic:
+    # - t_union has groups in order, then s:e ordered by s increasing and e decreasing
+    # - in a group (asset, year, rep_period) we can take the first range then
+    # - continue in the sequence selecting the next largest e
```
abelsiqueira, line 252: Remove or update

abelsiqueira: Use

- 1:4 5:8 9:12
- 1:3 4:4 5:9 10:12

1 2 3 4 
o o o o
o o o
           o o o o
                o o o o
                o o o o o
                              o o o o
                                 o o o

o o o o
               o

gnawin (approved):
LGTM

abelsiqueira:
Thanks!

Issue #796, opened 2024-09-13
Add formatter option to have explicit returns


abelsiqueira:
### What and Why

The following returns `nothing` implicitly:

```Julia
function foo()
  dict = Dict()
  for x = 1:3
    dict[x] = x^2
  end
end
```

The correct version must have a `return dict` in the end. Since this is implicit, it's easy to miss. I recommend adding a formatter option to automatically add `return`.

PR #928, opened 2024-11-14
927 tulipa constraints

abelsiqueira:
- **Refactor variables to use TulipaVariable**
- **[fixup] Refactor more variables**
- **[fixup] Missed a few things**
- **[fixup] Move assets_decomission_compact_method to TulipaVariable**
- **Add TODOs from PR review**
- **Refactor creation of investment variable using a helper function**
- **Update src/variables/create.jl**
- **Move flows variables out of dataframes**
- **Move units_on variable out of dataframes**
- **Move storage_level_intra_rp out of dataframes**
- **Move storage_level_inter_rp out of dataframes**
- **Separate creation of union and lowest resolution tables from storage_level_intra_rp creation**
- **Create is_charging without dataframes - no more dataframes for variables**
- **Create TulipaConstraint structure**
- **Move constraint :lowest from dataframes to TulipaConstraint**


src/create-model.jl
```diff
@@ -80,20 +75,9 @@ function create_model(
...
-        outgoing_flow_lowest_storage_resolution_intra_rp,
-        incoming_flow_highest_in_out_resolution,
-        outgoing_flow_highest_in_out_resolution,
-        incoming_flow_highest_in_resolution,
-        outgoing_flow_highest_out_resolution,
-        incoming_flow_storage_inter_rp_balance,
-        outgoing_flow_storage_inter_rp_balance,
-    ) = add_expressions_to_dataframe!(
-        dataframes,
+    add_expressions_to_dataframe!(
```
datejada, line 78: Maybe we need to rename this function —something like "add_expressions_for_constraints" to be more specific.

test/test-time-resolution.jl
datejada, line 1: Should we test the SQL queries instead? We might want to keep testing the functionality even though the method used to determine the partition differs.

src/tmp.jl
```diff
@@ -240,26 +240,52 @@ function tmp_create_union_tables(connection)
...
+        "CREATE OR REPLACE TEMP TABLE t_union_in_flows AS
+        SELECT DISTINCT to_asset as asset, year, rep_period, time_block_start, time_block_end
+        FROM flow_time_resolution
+        ",
+    )
+
+    # Outgoing flows
+    DuckDB.execute(
+        connection,
+        "CREATE OR REPLACE TEMP TABLE t_union_out_flows AS
```
datejada, line 255: These new tables are available in the connection object. Having an overview of all tables available in the connection will be nice, including which ones are read from the files/DB/csv and the ones that are created internally in the model.

src/tmp.jl
```diff
@@ -240,26 +240,52 @@ function tmp_create_union_tables(connection)
     # These are equivalent to the partitions in time-resolution.jl
     # But computed in a more general context to be used by variables as well
 
+    # Incoming flows
+    DuckDB.execute(
+        connection,
+        "CREATE OR REPLACE TEMP TABLE t_union_in_flows AS
```
gnawin, line 246: Is it okay to just use `CREATE`?

src/tmp.jl
```diff
@@ -240,26 +240,52 @@ function tmp_create_union_tables(connection)
...
+        SELECT DISTINCT to_asset as asset, year, rep_period, time_block_start, time_block_end
```
gnawin, line 247: Also here, okay to use only "SELECT"? Do we have duplicates?

src/tmp.jl
```diff
@@ -240,26 +240,52 @@ function tmp_create_union_tables(connection)
...
+    )
+
+    # Union of all assets and outgoing flows
+    DuckDB.execute(
+        connection,
+        "CREATE OR REPLACE TEMP TABLE t_union_assets_and_out_flows AS
+        SELECT DISTINCT asset, year, rep_period, time_block_start, time_block_end
+        FROM asset_time_resolution
+        UNION
+        FROM t_union_out_flows
```
gnawin, line 268: Sorry I'm also learning SQL a bit in the process: is it fine to neglect `SELECT` after `UNION` :)

src/model-preparation.jl
```diff
@@ -532,106 +390,47 @@ function add_expressions_to_dataframe!(
...
-        incoming_flow_lowest_storage_resolution_intra_rp,
-        outgoing_flow_lowest_storage_resolution_intra_rp,
-        incoming_flow_highest_in_out_resolution,
-        outgoing_flow_highest_in_out_resolution,
-        incoming_flow_highest_in_resolution,
-        outgoing_flow_highest_out_resolution,
-        incoming_flow_storage_inter_rp_balance,
-        outgoing_flow_storage_inter_rp_balance,
-    )
+    return
```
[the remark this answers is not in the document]
abelsiqueira: It's just a coding style choice, it doesn't make a difference (https://docs.julialang.org/en/v1/manual/functions/#Returning-nothing). I don't think there is a JuliaFormatter option to force one or the other. There is an option to force writing `return` (https://domluna.github.io/JuliaFormatter.jl/dev/#always_use_return,  #796 ), but it's something else.
If we do decide on anything, we should add it to https://github.com/TulipaEnergy/TulipaEnergyModel.jl/blob/main/docs/src/91-developer.md#code-format-and-guidelines

test/test-time-resolution.jl
abelsiqueira: It would be good to have some tests, indeed, but the code is not robust yet to allow that, because everything is hardcoded. This is related to what @suvayu said about accepting table names as keywords that is written down in one of the TODOs.

src/tmp.jl
abelsiqueira: That's the Miro description. When finished it can be rewritten into an official SQL description if desired, to have something in code format.

src/constraints/create.jl
```diff
@@ -0,0 +1,174 @@
...
+        ) |> DataFrame,
+    )
+
+    constraints[:storage_level_inter_rp] = TulipaConstraint(
+        DuckDB.query(
+            connection,
+            "SELECT * FROM storage_level_inter_rp
+            ",
+        ) |> DataFrame,
+    )
```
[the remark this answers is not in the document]
abelsiqueira: Yes, that's correct

src/tmp.jl
abelsiqueira: It should be, in the final version we can change to that, if preferred. However, for development it makes it easier to rerun some functions without rerunning everything. If only `CREATE` is used and the table exists, it will fail and then you have to manually delete the table or create a new `connection` object.

src/tmp.jl
abelsiqueira: For this table it is ok to write only `SELECT` because they are filtered later in the resolution computation, but might as well do it now.
There are probably duplicates because this ignores `from_asset`.

src/tmp.jl
abelsiqueira: Not for all SQL languages, but DuckDB allows writing `FROM table` without the `SELECT`. It means `SELECT * FROM table`. We can be explicit, if preferred

src/constraints/create.jl
```diff
@@ -0,0 +1,174 @@
+export compute_constraints_indices
+
+function compute_constraints_indices(connection)
+    constraints = Dict{Symbol,TulipaConstraint}()
+
+    constraints[:lowest] = TulipaConstraint(
```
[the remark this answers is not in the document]
abelsiqueira: This still uses the old table names. Miro shows the tables after renaming and some splitting, which hasn't happened yet to avoid further complications. There are some purple stickers in the bottom of the constraint tables showing where they come from.

src/create-model.jl
datejada: Great! Thanks

src/model-preparation.jl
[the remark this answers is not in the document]
datejada: Thanks! I like enforcing the return, but I don't have a strong opinion on the `nothing`. So, I created #940 for this.

test/test-time-resolution.jl
datejada: Great! then good to know that it is in the TODOs. Thanks

src/tmp.jl
datejada: Great, I have created #941 to keep it in our minds.

datejada (approved):
Thanks for the replies and updates. I created a couple of issues to work on after we have all the tables on MIRO in the code and a final overview of them.

PR #946, opened 2024-12-04
Refactor investment constraints

### LINKED ISSUE #945: Don't get data from graph object in src/constraints/investment.jl

datejada:
Refactor the investment constraints to use tables instead of sets in their creation.

## Related issues

Closes #945

## Checklist

- [x] Tests are passing

### DIFF
--- a/benchmark/benchmarks.jl
+++ b/benchmark/benchmarks.jl
@@ -30,9 +30,9 @@ end samples = 3 evals = 1 seconds = 86400
 energy_problem = EnergyProblem(connection)
 
 SUITE["energy_problem"]["create_model"] = @benchmarkable begin
-    create_model!($energy_problem)
+    create_model!($connection, $energy_problem)
 end samples = 3 evals = 1 seconds = 86400
-create_model!(energy_problem)
+create_model!(connection, energy_problem)
 
 # SUITE["energy_problem"]["output"] = @benchmarkable begin
 #     save_solution_to_file($OUTPUT_FOLDER_BM, $energy_problem)
--- a/benchmark/benchmarks.jl
+++ b/benchmark/benchmarks.jl
@@ -30,9 +30,9 @@ end samples = 3 evals = 1 seconds = 86400
 energy_problem = EnergyProblem(connection)
 
 SUITE["energy_problem"]["create_model"] = @benchmarkable begin
-    create_model!($connection, $energy_problem)
+    create_model!($energy_problem)
 end samples = 3 evals = 1 seconds = 86400
-create_model!(connection, energy_problem)
+create_model!(energy_problem)
 
 # SUITE["energy_problem"]["output"] = @benchmarkable begin
 #     save_solution_to_file($OUTPUT_FOLDER_BM, $energy_problem)
--- a/src/constraints/investment.jl
+++ b/src/constraints/investment.jl
@@ -21,16 +21,22 @@ function add_investment_constraints!(graph, sets, variables)
     flows_investment = variables[:flows_investment].lookup
 
     # - Maximum (i.e., potential) investment limit for assets
-    for y in Y, a in Ai[y]
-        if graph[a].capacity > 0 && !ismissing(graph[a].investment_limit[y])
-            bound_value = _find_upper_bound(graph, y, a)
-            JuMP.set_upper_bound(assets_investment[y, a], bound_value)
+    sub_df = filter(row -> row.capacity > 0, variables[:assets_investment].indices)
+    for (assets_investment, row) in zip(variables[:assets_investment].container, eachrow(sub_df))
+        if !ismissing(graph[row.asset].investment_limit[row.milestone_year])
+            bound_value = _find_upper_bound(graph, row.milestone_year, row.asset)
+            JuMP.set_upper_bound(assets_investment, bound_value)
         end
-        if (a in Ase[y]) && # for a in Ase, i.e., storage assets with energy method
-           graph[a].capacity_storage_energy > 0 &&
-           !ismissing(graph[a].investment_limit_storage_energy[y])
-            bound_value = _find_upper_bound(graph, y, a; is_bound_for_energy = true)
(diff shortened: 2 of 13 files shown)


benchmark/benchmarks.jl
```diff
@@ -30,9 +30,9 @@ end samples = 3 evals = 1 seconds = 86400
 energy_problem = EnergyProblem(connection)
 
 SUITE["energy_problem"]["create_model"] = @benchmarkable begin
-    create_model!($energy_problem)
+    create_model!($connection, $energy_problem)
```
abelsiqueira, line 33: We have the connection inside the `energy_problem` (named `db_connection`, maybe we can rename), so this can be avoided

datejada: Yes, we can! it is done in the latest changes

src/constraints/investment.jl
```diff
@@ -1,62 +1,131 @@
...
-    flows_investment = variables[:flows_investment].lookup
 
     # - Maximum (i.e., potential) investment limit for assets
-    for y in Y, a in Ai[y]
-        if graph[a].capacity > 0 && !ismissing(graph[a].investment_limit[y])
-            bound_value = _find_upper_bound(graph, y, a)
-            JuMP.set_upper_bound(assets_investment[y, a], bound_value)
+    constraint_indices = _create_asset_investment_indices(connection)
+    filtered_indices =
+        filter(row -> row.capacity > 0 && !ismissing(row.investment_limit), constraint_indices)
```
abelsiqueira, line 15: Instead of return `DataFrame`, we can return the query directly. This makes it more memory efficient. Then, you also don't use `for row in eachrow(table)`, and use directly `for row in table` (or the zip equivalent).

Then, filtering like this doesn't work, and instead we filter inside the loop - I'll also comment something related below

src/constraints/investment.jl
```diff
@@ -1,62 +1,131 @@
...
+    if !isempty(filtered_indices)
+        filtered_indices.bound_value .= _find_upper_bound(
+            filtered_indices.investment_limit,
+            filtered_indices.capacity,
+            filtered_indices.investment_integer,
+        )
+        for (assets_investment, row) in
+            zip(variables[:assets_investment].container, eachrow(filtered_indices))
```
abelsiqueira, line 23: There is no guarantee that `filtered_indices` and `variables[:assets_investment]` have the same length, so this `zip` can silently stop early and wrongly link different elements - I spent a few hours with this issue.

src/constraints/investment.jl
```diff
@@ -1,62 +1,131 @@
...
+            indices.milestone_year,
+            indices.investment_integer_storage_energy,
+            asset.capacity_storage_energy,
+            asset_commission.investment_limit_storage_energy,
+        FROM indices_for_assets_investment_energy AS indices
+        LEFT JOIN asset
+            ON asset.asset = indices.asset
+        LEFT JOIN asset_commission
+            ON asset_commission.asset = indices.asset
+                AND asset_commission.commission_year = indices.milestone_year
```
abelsiqueira, line 97: Also add a `ORDER BY index` to make sure that follow the initial order to correctly zip with the variables.

src/constraints/investment.jl
```diff
@@ -1,62 +1,131 @@
...
 
     # - Maximum (i.e., potential) investment limit for assets
-    for y in Y, a in Ai[y]
-        if graph[a].capacity > 0 && !ismissing(graph[a].investment_limit[y])
-            bound_value = _find_upper_bound(graph, y, a)
-            JuMP.set_upper_bound(assets_investment[y, a], bound_value)
+    constraint_indices = _create_asset_investment_indices(connection)
+    filtered_indices =
+        filter(row -> row.capacity > 0 && !ismissing(row.investment_limit), constraint_indices)
+    if !isempty(filtered_indices)
```
abelsiqueira, line 16: When we return the DuckDB table instead of the DF, there is no cheap way to verify this, so instead just remove it.

src/constraints/investment.jl
```diff
@@ -1,62 +1,131 @@
...
+                AND flow.to_asset = indices.to_asset
+        LEFT JOIN flow_commission
+            ON flow_commission.from_asset = indices.from_asset
+                AND flow_commission.to_asset = indices.to_asset
+                AND flow_commission.commission_year = indices.milestone_year
+        ",
+    ) |> DataFrame
+end
+
+function _find_upper_bound(investment_limit, capacity, investment_integer)
```
abelsiqueira, line 124: With the proposed changes, I think the inputs will be numbers, not arrays

src/constraints/investment.jl
```diff
@@ -1,62 +1,131 @@
...
-            JuMP.set_upper_bound(assets_investment[y, a], bound_value)
+    constraint_indices = _create_asset_investment_indices(connection)
+    filtered_indices =
+        filter(row -> row.capacity > 0 && !ismissing(row.investment_limit), constraint_indices)
+    if !isempty(filtered_indices)
+        filtered_indices.bound_value .= _find_upper_bound(
+            filtered_indices.investment_limit,
+            filtered_indices.capacity,
+            filtered_indices.investment_integer,
+        )
```
abelsiqueira, line 21: This allocates, but with the change to use DuckDB table, we will have to compute them for each row instead, so it won't be an issue anymore

src/variables/create.jl
```diff
@@ -113,7 +113,8 @@ function compute_variables_indices(connection)
     variables[:flows_investment] = TulipaVariable(
         DuckDB.query(
             connection,
-            "SELECT
+            "CREATE OR REPLACE TABLE indices_for_flows_investment AS
```
abelsiqueira, line 116: I used `var_...` in #950 because of the names in Miro.

abelsiqueira (changes requested):
Thanks! I've made some comments to one of the variables and one of the queries that should apply to all. Let me know what you think

datejada:
Close in favour of #954

PR #949, opened 2024-12-05
Remove old code for compute_assets_partitions

### LINKED ISSUE #948: Update `compute_assets_partitions!`  to use DuckDB more efficiently

abelsiqueira:
Since the code does not exist anymore and the partitions are being computed
in a better way in src/tmp.jl, then

Closes #948


datejada (approved):
LGTM

PR #950, opened 2024-12-05
Store a table name inside TulipaVariable and TulipaConstraint

abelsiqueira:
With the table_name, we use a different constructor, based only on
the connection and table name.

### DIFF
--- a/src/constraints/create.jl
+++ b/src/constraints/create.jl
@@ -1,174 +1,182 @@
 export compute_constraints_indices
 
 function compute_constraints_indices(connection)
-    constraints = Dict{Symbol,TulipaConstraint}()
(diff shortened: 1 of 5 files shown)


abelsiqueira:
@datejada, I think this might be useful in the future, so I did it in a separate PR. 
If you check this link, https://github.com/TulipaEnergy/TulipaEnergyModel.jl/pull/950/files?w=1, you can see the differences excluding whitespace changes, which in this case is very helpful.

This PR creates the tables for variables and constraints and then creates the TulipaVariable and TulipaConstraint dictionaries using only the connection and the table name. This PR also stores the table name in the variable and constraint objects, which might be useful and cleaner as well.

datejada (approved):
After review together with @abelsiqueira, the PR is approved

PR #956, opened 2024-12-10
Use `@timeit` on function calls

### LINKED ISSUE #934: Try using `@timeit to` on function calls, not on `begin ... end` blocks
### Description

When an error happens inside the `begin ... end` block, the`@timeit` macro will hide the line where the error happened by just saying that the error happens at the macro line, and they listing `macro expansion` errors - which is unhelpful

gnawin:
## Related issues

Closes #934 

- Remove EnergyProblem.timings
- Use `@timeit` on function calls

## Checklist

- [x] Tests are passing

### DIFF
--- a/src/create-model.jl
+++ b/src/create-model.jl
@@ -14,7 +14,7 @@ function create_model!(energy_problem; kwargs...)
     groups = energy_problem.groups
     model_parameters = energy_problem.model_parameters
     years = energy_problem.years
-    sets = create_sets(graph, years)
+    sets = @timeit to "create_sets" create_sets(graph, years)
     energy_problem.model = @timeit to "create_model" create_model(
         energy_problem.db_connection,
         graph,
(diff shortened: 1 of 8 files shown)


abelsiqueira (approved):
LGTM, thanks @gnawin !

gnawin:
@abelsiqueira @datejada After checking the time log [log.txt](https://github.com/user-attachments/files/18078212/log.txt) for the Tiny case, here are a few observations:
- `create_interna_structures` is 27.6% of the time, which seems mostly spent on the functions related to reading tables (e.g., `TulipaIO.get_table`, `DBInterface.execute`). Probably there is nothing we can do better, but interesting to see.
- `compute_dual_variables ` is 3 times more expensive than `total solver time`. Seeing the code, it makes sense, but still, good to realize.

abelsiqueira:
Nice insights. We should definitely do that as part of the efficiency improvement for the EU dataset

datejada:
Thanks @gnawin; the compute duals will change once we change how to store the solution and hopefully, it will be faster than the current approach 😄

Issue #642, opened 2024-05-13
Rename the variables and constraints dataframes in the model


datejada:
### What and Why

Rename constraint partitions to be closer to the constraint name itself (e.g., power or balance).
We are also updating the definition of `TulipaConstraint` to allow multiple `@constraint`s to be defined for the same `TulipaConstraint`, i.e, same indices. We do require that each `@constraint` use a full index set, i.e., no filtering of the rows of a constraint table.

### Related Issues

**variables**

- [ ] Rename variable table names to names in Miro - e.g. `var_storage_level_intra_rp` becomes `var_storage_rep_period`
- [ ] Rename variable key names and other places to reflect the table names - e.g., `variables[:storage_level_intra_rp]` becomes `variables[:storage_rep_period]`

**constraints**

- [x] #957
- [x] #958
- [x] Rename, split, create new constraints for each file:
  - [x] #961
    - [x] `assets_profile_times_capacity_out` should be duplicated and attached to corresponding tables
    - [ ] split `capacity_outgoing` into compact and simple and use for `max_output_...`
    - [x] Remove `acc..._inv..._lim...` expression
    - [x] Create new tables `max_output..._binary_part1` and `...part2` for outputs and for inputs (4 tables). Use a name that reflect using investment limit and accumulated investment, binary, etc.
  - [x] Rename, split, create new constraints for consumer.jl
  - [x] Rename, split, create new constraints for conversion.jl
  - [x] Rename, split, create new constraints for energy.jl
  - [x] #969
  - [x] Rename, split, create new constraints for hub.jl
  - [x] #964
    - [ ] Create constraint `limit_units_on`. Copy from `units_on` variable and update Miro
    - [ ] Create constraints `min_output_...` and `max_output_...` similar to `ramping_with_unit_commitment` (file ramp...jl). This requires two tables, since they don't have the same indices
    - [ ] Rename expression `units_on`, at least inside the file ramp...jl
    - [ ] Create a new table `max_ramp_with_unit_commitment` to attach `max_ramp_up_...` and `min_ramp_up...`
    - [ ] Create new table `max_ramp_without_unit_commitment` to attach related `max_ramp` constraints. Notice that `k = 1` will be an exception case, for now set to `0.0 == 0.0`.
    - [ ] `assets_profile_times_capacity_out` should be duplicated and attached to corresponding tables
  - [x] #966
  - [x] Rename, split, create new constraints for transport.jl

[...]

PR #959, opened 2024-12-11
Update TulipaConstraint to allow attaching more constraints to the same indices table

### LINKED ISSUE #957: Update `TulipaConstraint` to store the dual variables and the list of constraints that use the same set of indices

abelsiqueira:
With the idea of allowing multiple constraints per indices table, this links the indices to the constraints. It also creates a function `attach_constraint!` to help keep things in check.

It fails the [avoid fields with abstract types](https://docs.julialang.org/en/v1/manual/performance-tips/#Avoid-fields-with-abstract-type) recommendation (`ConstraintRef` has three parametric types), so it might be unnecessarily slow.
I see two solutions:

1. Don't store the container at all, just the names. This is probably sufficient and cleaner.
2. Store the multiple types that ConstraintRef can have, either with multiple fields, dict of dicts, or more complicated structures. I don't like this idea because it makes the code more complex and involves more JuMP specific things. However, if the container for the containers is required, this is necessary (but so I far I don't see it being required).

EDIT: I added a commit with option 1 implemented.

@datejada, what do you think?

Closes #957

### DIFF
--- a/src/structures.jl
+++ b/src/structures.jl
@@ -94,7 +94,9 @@ function attach_constraint!(
 end
 
 function attach_constraint!(model::JuMP.Model, cons::TulipaConstraint, name::Symbol, container)
-    # This should be the empty case container = Any[]
+    # This should be the empty case container = Any[] that happens when the
+    # indices table in empty in [@constraint(...) for row in eachrow(indices)].
+    # It resolves to [] so the element type cannot be inferred
     @assert length(container) == 0
     @assert cons.num_rows == 0
     empty_container = JuMP.ConstraintRef{JuMP.Model,Missing,JuMP.ScalarShape}[]
--- a/src/structures.jl
+++ b/src/structures.jl
@@ -100,7 +100,9 @@ function attach_constraint!(model::JuMP.Model, cons::TulipaConstraint, name::Sym
     # indices table in empty in [@constraint(...) for row in eachrow(indices)].
     # It resolves to [] so the element type cannot be inferred
     if length(container) > 0
-        error("This variant is supposed to capture empty containers. This container is not empty for $name")
+        error(
+            "This variant is supposed to capture empty containers. This container is not empty for $name",
+        )
     end
     if cons.num_rows > 0
         error("The number of rows in indices table should be 0 for $name")
--- a/src/structures.jl
+++ b/src/structures.jl
@@ -100,7 +100,7 @@ function attach_constraint!(model::JuMP.Model, cons::TulipaConstraint, name::Sym
     # indices table in empty in [@constraint(...) for row in eachrow(indices)].
     # It resolves to [] so the element type cannot be inferred
     if length(container) > 0
-        error("This variant is supposed to capture empty containers. This container is not empty")
+        error("This variant is supposed to capture empty containers. This container is not empty for $name")
     end
     if cons.num_rows > 0
         error("The number of rows in indices table should be 0")
--- a/src/structures.jl
+++ b/src/structures.jl
@@ -119,7 +119,7 @@ This checks that the `container` length matches the stored `indices` number of r
 """
 function attach_expression!(cons::TulipaConstraint, name::Symbol, container::Vector{JuMP.AffExpr})
     if length(container) != cons.num_rows
-        error("The number of expressions does not match the number of rows in the indices")
+        error("The number of expressions does not match the number of rows in the indices of $name")
     end
     cons.expressions[name] = container
     return nothing
--- a/src/structures.jl
+++ b/src/structures.jl
@@ -103,7 +103,7 @@ function attach_constraint!(model::JuMP.Model, cons::TulipaConstraint, name::Sym
         error("This variant is supposed to capture empty containers. This container is not empty for $name")
     end
     if cons.num_rows > 0
-        error("The number of rows in indices table should be 0")
+        error("The number of rows in indices table should be 0 for $name")
     end
     empty_container = JuMP.ConstraintRef{JuMP.Model,Missing,JuMP.ScalarShape}[]
     model[name] = empty_container
--- a/src/structures.jl
+++ b/src/structures.jl
@@ -88,7 +88,7 @@ function attach_constraint!(
     container::Vector{<:JuMP.ConstraintRef},
 )
     if length(container) != cons.num_rows
-        error("The number of constraints does not match the number of rows in the indices")
+        error("The number of constraints does not match the number of rows in the indices of $name")
     end
     push!(cons.constraint_names, name)
     model[name] = container
--- a/src/structures.jl
+++ b/src/structures.jl
@@ -110,21 +110,22 @@ Attach a expression named `name` stored in `container`, and optionally set `mode
 This checks that the `container` length matches the stored `indices` number of rows.
 """
 function attach_expression!(cons::TulipaConstraint, name::Symbol, container::Vector{JuMP.AffExpr})
-    @assert length(container) = cons.num_rows
+    @assert length(container) == cons.num_rows
     cons.expressions[name] = container
     return nothing
 end
 
-function attach_expression!(
-    cons::TulipaConstraint,
-    name::Symbol,
-    container::Vector{JuMP.AffExpr},
-    model::JuMP.Model,
-)
-    attach_expression!(cons, name, container)
-    model[name] = container
-    return nothing
-end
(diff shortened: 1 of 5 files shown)


datejada:
@abelsiqueira option 1 is preferred, and I see you started with it. So, Thanks!

src/structures.jl
```diff
@@ -55,17 +55,77 @@ end
...
+    duals::Dict{Symbol,Vector{Float64}}
 
     function TulipaConstraint(connection, table_name::String)
         return new(
             DuckDB.query(connection, "SELECT * FROM $table_name") |> DataFrame,
             table_name,
+            only([ # only makes sure that a single value is returned
+                row.num_rows for
+                row in DuckDB.query(connection, "SELECT COUNT(*) AS num_rows FROM $table_name")
+            ]), # This loop is required to access the query resulted, because it's a lazy struct
```
datejada, line 70: Maybe it's a silly question, but why can it return more than one value? I would expect it to always return only one value, although the `only` is a safeguard to ensure that.

datejada, line 70: Maybe it's a silly question, but why can it return more than one value? I would expect it to always return only one value, although the `only` is a safeguard to ensure that.

src/structures.jl
```diff
@@ -55,17 +55,77 @@ end
...
+    attach_constraint!(model, cons, name, container)
+
+Attach a constraint named `name` stored in `container`, and set `model[name] = container`.
+This checks that the `container` length matches the stored `indices` number of rows.
+"""
+function attach_constraint!(
+    model::JuMP.Model,
+    cons::TulipaConstraint,
+    name::Symbol,
+    container::Vector{<:JuMP.ConstraintRef},
```
datejada, line 88: I didn't know that there is a ConstraintRef in JuMP. So, can we use [JuMP.VariableRef](https://jump.dev/JuMP.jl/stable/api/JuMP/#JuMP.VariableRef) to do the same? In that sense, we don't don't store the variables twice. Then the TulipaConstraint and TulipaVariable structure will become more reference to the model objects that make it easier to find stuff without duplicating info. What do you think?

src/structures.jl
```diff
@@ -55,17 +55,77 @@ end
...
+    attach_expression!(cons, name, container)
+    attach_expression!(model, cons, name, container)
+
+Attach a expression named `name` stored in `container`, and optionally set `model[name] = container`.
+This checks that the `container` length matches the stored `indices` number of rows.
+"""
+function attach_expression!(cons::TulipaConstraint, name::Symbol, container::Vector{JuMP.AffExpr})
+    @assert length(container) = cons.num_rows
+    cons.expressions[name] = container
+    return nothing
```
datejada, line 115: This is our friend Codecov doing it stuff, correct? I mean, it is an apparent hit in the code, but the only version misses it for whatever reason. I don't care about this as a problem, but just to be sure, I get it right.

src/structures.jl
abelsiqueira: I didn't understand what you mean. The `TulipaVariable` should not be duplicating info, just pointing to the right place inside the model.
We could store only the name of the variable and use `model[variable.var_name]`, if that's what you mean. However, that would require passing the `model` as well everywhere, and since we are already accessing the `variables` dictionary via the variable name, it would not add anything. I.e., the key to the `TulipaVariable` inside the Dict `variables` would be stored inside the `TulipaVariable` itself, and we would still require `model` to get the actual variable. We could try making this change just to make sure that it's not creating duplicate memory, but we haven't seen such a large increase when we created the `TulipaVariable` in the past. Anyway, I digress and maybe that's not even what you mean

src/structures.jl
abelsiqueira: I fixed this. The `attach_expression!` was indeed not being used. I commented out the `model` version until we actually use it, though

abelsiqueira:
Thanks for the review @datejada, I commented on all of them

src/structures.jl
```diff
@@ -55,17 +55,77 @@ end
...
+Attach a constraint named `name` stored in `container`, and set `model[name] = container`.
+This checks that the `container` length matches the stored `indices` number of rows.
+"""
+function attach_constraint!(
+    model::JuMP.Model,
+    cons::TulipaConstraint,
+    name::Symbol,
+    container::Vector{<:JuMP.ConstraintRef},
+)
+    @assert length(container) == cons.num_rows
```
datejada, line 90: Can we have a message here? I mean, it will help us debug in the future.

datejada (commented):
Hi @abelsiqueira, thanks for the PR. There were some comments and questions during the review, but generally speaking, it looks cleaner with the changes, and I like the approach. Thanks.

src/structures.jl
datejada: I was confused for a moment, but we do have the `Vector{JuMP.VariableRef}` in the TulipaVariable. So, all good; thanks for the reply; it helped me to realize that we already have what I was trying to propose 😆

src/structures.jl
```diff
@@ -55,17 +55,77 @@ end
...
+    push!(cons.constraint_names, name)
+    model[name] = container
+    return nothing
+end
+
+function attach_constraint!(model::JuMP.Model, cons::TulipaConstraint, name::Symbol, container)
```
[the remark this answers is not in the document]
abelsiqueira: You're right, this should be explained better, I will improve the comment. This is called for 

```julia
container = [@constraint(...) for row in eachrow(indices)]
```

So when `indices` is an empty table, it becomes `container = []`, which doesn't have a type so it becomes `Any[]`. This captures that case.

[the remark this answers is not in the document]
datejada: 100% clear now! Thanks

abelsiqueira:
Thanks for the review! I've updated the PR with the error messages and further comments. Let me know what you think

src/structures.jl
abelsiqueira: Ah nice

src/structures.jl
```diff
@@ -55,17 +55,88 @@ end
...
+"""
+    attach_expression!(cons, name, container)
+    attach_expression!(model, cons, name, container)
+
+Attach a expression named `name` stored in `container`, and optionally set `model[name] = container`.
+This checks that the `container` length matches the stored `indices` number of rows.
+"""
+function attach_expression!(cons::TulipaConstraint, name::Symbol, container::Vector{JuMP.AffExpr})
+    if length(container) != cons.num_rows
+        error("The number of expressions does not match the number of rows in the indices")
```
datejada, line 122:
```suggestion
        error("The number of expressions does not match the number of rows in the indices of $name")
```

src/structures.jl
```diff
@@ -55,17 +55,88 @@ end
...
+This checks that the `container` length matches the stored `indices` number of rows.
+"""
+function attach_constraint!(
+    model::JuMP.Model,
+    cons::TulipaConstraint,
+    name::Symbol,
+    container::Vector{<:JuMP.ConstraintRef},
+)
+    if length(container) != cons.num_rows
+        error("The number of constraints does not match the number of rows in the indices")
```
datejada, line 91:
```suggestion
        error("The number of constraints does not match the number of rows in the indices of $name")
```

src/structures.jl
```diff
@@ -55,17 +55,88 @@ end
...
+    model[name] = container
+    return nothing
+end
+
+function attach_constraint!(model::JuMP.Model, cons::TulipaConstraint, name::Symbol, container)
+    # This should be the empty case container = Any[] that happens when the
+    # indices table in empty in [@constraint(...) for row in eachrow(indices)].
+    # It resolves to [] so the element type cannot be inferred
+    if length(container) > 0
+        error("This variant is supposed to capture empty containers. This container is not empty")
```
datejada, line 103:
```suggestion
        error("This variant is supposed to capture empty containers. This container is not empty for $name")
```

src/structures.jl
```diff
@@ -55,17 +55,88 @@ end
...
+    end
+    if cons.num_rows > 0
+        error("The number of rows in indices table should be 0")
```
datejada, line 106:
```suggestion
        error("The number of rows in indices table should be 0 for $name")
```

datejada (approved):
Great! Thanks for the changes; I am approving this PR. You can merge after all the checks have been done.

abelsiqueira:
Thanks! I'll merge after the benchmarks run and rebase #642

Issue #960, opened 2024-12-11
Remove `active` parameter from flows and assets


datejada:
### Description

Since it will be part of the user-friendly files, the `active` parameter is no longer needed for the Tulipa input files.

Might make this issue obsolete:
- #266

clizbe:
I was gonna grab this one then realized it might have major conflicts with what you're working on. :/

PR #962, opened 2024-12-12
Move some of the active checks earlier in the workflow

abelsiqueira:
This doesn't remove the active as proposed in #960, but it moves the check to earlier in the workflow - to the highest and lowest resolution.

Relates to #960

### DIFF
--- a/src/tmp.jl
+++ b/src/tmp.jl
@@ -375,8 +375,17 @@ function tmp_create_lowest_resolution_table(connection)
         current_group = ("", 0, 0)
         @timeit to "append $table_name rows" for row in DuckDB.query(
             connection,
-            "SELECT * FROM $union_table
-            ORDER BY asset, year, rep_period, time_block_start
+            "SELECT t_union.* FROM $union_table AS t_union
+            LEFT JOIN asset_both
+                ON asset_both.asset = t_union.asset
+                AND asset_both.milestone_year = t_union.year
+                AND asset_both.commission_year = t_union.year
+            WHERE asset_both.active = true
+            ORDER BY
+                t_union.asset,
+                t_union.year,
+                t_union.rep_period,
+                t_union.time_block_start
             ",
         )
             if (row.asset, row.year, row.rep_period) != current_group
@@ -418,6 +427,7 @@ function tmp_create_highest_resolution_table(connection)
     # - keep all unique time_block_start
     # - create corresponing time_block_end
 
+    # filtering by active
     for union_table in
         ("t_union_" * x for x in ("in_flows", "out_flows", "assets_and_out_flows", "all_flows"))
         table_name = replace(union_table, "t_union" => "t_highest")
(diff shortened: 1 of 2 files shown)


src/tmp.jl
```diff
@@ -418,28 +427,34 @@ function tmp_create_highest_resolution_table(connection)
...
             ) AS t_union
+            LEFT JOIN asset_both
+                ON asset_both.asset = t_union.asset
+                AND asset_both.milestone_year = t_union.year
+                AND asset_both.commission_year = t_union.year
             LEFT JOIN rep_periods_data
                 ON t_union.year = rep_periods_data.year
                     AND t_union.rep_period = rep_periods_data.rep_period
-            ORDER BY asset, t_union.year, t_union.rep_period, time_block_start
+            WHERE asset_both.active = true
```
datejada, line 456: Just for my understanding, the JOIN here is to have available in the table the `active` parameter. So, when we take out `active` from the Tulipa files in #960 we don't have to do this JOIN. Is that correct?

src/tmp.jl
```diff
@@ -375,8 +375,17 @@ function tmp_create_lowest_resolution_table(connection)
...
         @timeit to "append $table_name rows" for row in DuckDB.query(
             connection,
-            "SELECT * FROM $union_table
-            ORDER BY asset, year, rep_period, time_block_start
+            "SELECT t_union.* FROM $union_table AS t_union
+            LEFT JOIN asset_both
+                ON asset_both.asset = t_union.asset
+                AND asset_both.milestone_year = t_union.year
+                AND asset_both.commission_year = t_union.year
+            WHERE asset_both.active = true
```
datejada, line 383: same comment here as in https://github.com/TulipaEnergy/TulipaEnergyModel.jl/pull/962/files#r1882306145

src/tmp.jl
abelsiqueira: The stored table itself won't have the `active` parameter because I haven't selected it, but the intermediary table resulting from the join will, which is then filtered in this `WHERE`. So the `JOIN` is only so that we have `active` available for filtering.

datejada: So, deleting the active also will also delete this extra join and make the code more simple 😄 

Thanks!

datejada (approved):
Hi, @abelsiqueira; it looks cleaner and simpler to have it in the calculation of the highest-resolution table. Thanks! This will also simplify the work in #960

src/tmp.jl
abelsiqueira: During #960, this will also be removed as there will be no mention of `active` anywhere. I was thinking about where the removal of `active` will happen, and the earliest point would be not having `active` at all in the input of TulipaEnergyModel. That means that during the workflow processing, at some point we check for `active` and remove. In the current workflow, it would be at the `read_csv_folder` function in TulipaIO. It's not clear yet the implications, and it's not urgent, so I'm not actively thinking about it.
For this PR, I noticed the constraint creations had a bunch of active checks, and the constraints could look a bit simpler if they didn't, so I shifted left.

abelsiqueira:
Thanks for the review!

Issue #961, opened 2024-12-12
Rename, split, create new constraints for capacity.jl


abelsiqueira:
- [x] `assets_profile_times_capacity_out` should be duplicated and attached to corresponding tables
- [ ] split `capacity_outgoing` into compact and simple and use for `max_output_...`
- [x] Remove `acc..._inv..._lim...` expression
- [x] Create new tables `max_output..._binary_part1` and `...part2` for outputs and for inputs (4 tables). Use a name that reflect 
- [x] Remove lower bound of flow and move to variable creation (move for src/variables/flow)

PR #963, opened 2024-12-12
Update capacity.jl to use the new attach functions and update tables

abelsiqueira:
- **Update capacity.jl to use attach and update tables related to it**

Part of #961

### DIFF
--- a/src/constraints/create.jl
+++ b/src/constraints/create.jl
@@ -12,9 +12,11 @@ function compute_constraints_indices(connection)
             :capacity_incoming,
             :capacity_incoming_storage_with_binary,
             :capacity_incoming_investable_storage_with_binary,
+            :capacity_incoming_investable_storage_limit_with_binary,
             :capacity_outgoing,
             :capacity_outgoing_storage_with_binary,
             :capacity_outgoing_investable_storage_with_binary,
+            :capacity_outgoing_investable_storage_limit_with_binary,
             :ramping_with_unit_commitment,
             :ramping_without_unit_commitment,
             :balance_storage_rep_period,
@@ -135,6 +137,26 @@ function _create_constraints_tables(connection)
     DuckDB.query(
         connection,
         "CREATE OR REPLACE TEMP SEQUENCE id START 1;
+        CREATE OR REPLACE TABLE cons_capacity_incoming_investable_storage_limit_with_binary AS
+        SELECT
+            nextval('id') AS index,
+            t_high.*
+        FROM t_highest_in_flows AS t_high
+        LEFT JOIN asset
+            ON t_high.asset = asset.asset
+        LEFT JOIN asset_milestone
+            ON t_high.asset = asset_milestone.asset
+            AND t_high.year = asset_milestone.milestone_year
+        WHERE
+            asset.type in ('storage')
+            AND asset.use_binary_storage_method in ('binary', 'relaxed_binary')
+            AND asset_milestone.investable
+        ",
+    )
+
+    DuckDB.query(
+        connection,
+        "CREATE OR REPLACE TEMP SEQUENCE id START 1;
         CREATE OR REPLACE TABLE cons_capacity_outgoing AS
         SELECT
             nextval('id') AS index,
@@ -149,7 +171,7 @@ function _create_constraints_tables(connection)
     DuckDB.query(
         connection,
         "CREATE OR REPLACE TEMP SEQUENCE id START 1;
-        CREATE OR REPLACE TABLE cons_capacity_outgoing_binary AS
+        CREATE OR REPLACE TABLE cons_capacity_outgoing_storage_with_binary AS
         SELECT
             nextval('id') AS index,
             t_high.*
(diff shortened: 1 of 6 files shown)


abelsiqueira:
This is not the whole of #961, but it's already a lot of changes - because it's one of the hardest files, I think.
Benchmarks say things are faster, though it might be randomness still. Otherwise, a reason could be that things are more type stable, with the `attach` functions, but I'm guessing.

@datejada, what do you think? Does it look clean enough?

datejada (commented):
@abelsiqueira I have committed my proposed changes. Let me know if you agree on them to continue. Thanks!

src/constraints/ramping-and-unit-commitment.jl
```diff
@@ -22,28 +22,31 @@ function add_ramping_constraints!(model, variables, constraints, graph, sets)
...
-                row.year,
-                ("availability", row.rep_period),
-                row.time_block_start:row.time_block_end,
-                1.0,
-            ) * graph[row.asset].capacity
-        ) for row in eachrow(cons_with.indices) if is_active(graph, row.asset, row.year)
-    ]
+    profile_times_capacity = Dict(
+        key => begin
+            table_name = Symbol("ramping_$(key)_unit_commitment")
```
datejada, line 27: Also, it reads better below if the keys are `:with_unit_commitment` and `:without_unit_commitment`

So, please change it either here or in the other PR you are working for the ramping. Thanks!

src/constraints/create.jl
```diff
@@ -149,7 +171,7 @@ function _create_constraints_tables(connection)
     DuckDB.query(
         connection,
         "CREATE OR REPLACE TEMP SEQUENCE id START 1;
-        CREATE OR REPLACE TABLE cons_capacity_outgoing_binary AS
+        CREATE OR REPLACE TABLE cons_capacity_outgoing_storage_with_binary AS
```
abelsiqueira, line 174: Took me a while, but if I understood correctly, this table should have a `WHERE` condition to check `NOT asset.investable` for it to be equal to the old code.

My logic (and I will use initials to avoid writing for 30 minutes) is that initially you had two tables: COB and COBI

- For COB, if investable, use `(c * aiu + ail) * (1 - ch)` else use `c * aiu * (1 - ch)`
- FOR COBI, use `c * (aiu * (1 - ch) + aiusm)`

Now you have three tables, COSB, COISB, COISLB.

- For COSB, `c * aiu * (1 - ch)`, so COSB = COB + not investable
- For COISB, `c * (aiu * (1 - ch) + aiusm)`, so COISB = COBI
- For COISLB, `(c * aiu + ail) * (1 - ch)`, so COISLB is COB + investable (which is COBI, I think)

If that is the case, you can also reuse the COISB table - you still separate the constraints to avoid the if condition, you just attach to the same constraint.
I checked the formulation, and indeed it separates like these tables, but I think that the formulation is also missing a `a \not \in A_y^I` or similar for that constraint.

Let me know if this makes sense or not

abelsiqueira (commented):
Thanks for the commits. I left a comment and I'll check your comments on this PR now

src/constraints/ramping-and-unit-commitment.jl
abelsiqueira: I agree. I will change in the next PR, because I had to completely change the keys to the table names, since they are used in other tables as well

src/constraints/create.jl
```diff
@@ -10,7 +10,13 @@ function compute_constraints_indices(connection)
             :balance_consumer,
             :balance_hub,
             :capacity_incoming,
+            :capacity_incoming_storage_with_binary,
```
datejada, line 13:
```suggestion
            :capacity_incoming_non_investable_storage_with_binary,
```

src/constraints/create.jl
```diff
@@ -10,7 +10,13 @@ function compute_constraints_indices(connection)
...
+            :capacity_incoming_investable_storage_with_binary,
+            :capacity_incoming_investable_storage_limit_with_binary,
             :capacity_outgoing,
+            :capacity_outgoing_storage_with_binary,
```
datejada, line 17:
```suggestion
            :capacity_outgoing_non_investable_storage_with_binary,
```

src/constraints/create.jl
```diff
@@ -92,6 +98,62 @@ function _create_constraints_tables(connection)
...
+            nextval('id') AS index,
+            t_high.*
+        FROM t_highest_in_flows AS t_high
+        LEFT JOIN asset
+            ON t_high.asset = asset.asset
+        WHERE
+            asset.type in ('storage')
+            AND asset.use_binary_storage_method in ('binary', 'relaxed_binary')
+        ",
+    )
```
datejada, line 115: update to include that it is non investable

src/constraints/create.jl
```diff
@@ -106,6 +168,62 @@ function _create_constraints_tables(connection)
...
+        SELECT
+            nextval('id') AS index,
+            t_high.*
+        FROM t_highest_out_flows AS t_high
+        LEFT JOIN asset
+            ON t_high.asset = asset.asset
+        WHERE
+            asset.type in ('storage')
+            AND asset.use_binary_storage_method in ('binary', 'relaxed_binary')
+        ",
```
datejada, line 184: Add the condition to make it only for non investable storage assets

src/constraints/create.jl
datejada: It makes sense, I left a couple of comments where I see we need the changes and also a rename in one of the constraints to make it explicit.

datejada: We need to update the formulation, too

abelsiqueira: Thanks. Do you want to finish the PR with these changes?

datejada (approved):
Thanks! LGTM

PR #970, opened 2024-12-16
Create transport table and update transport.jl

abelsiqueira:
Part of #642


datejada (approved):
Thanks for the PR. The changes follow the same pattern we have for the other constraints.

PR #972, opened 2024-12-16
Rename, split, create new constraints for group.jl

### LINKED ISSUE #969: Rename, split, create new constraints for group.jl

datejada:
## Related issues

Closes #969 

## Checklist

- [ ] Tests are passing


datejada:
@abelsiqueira this PR is ready for review. There are two things that might be better to address in separate issues:

1. Getting rid of the sets to create the expression for the group constraints
2. After 1, deleting the group structure. It will be useless after that

Let me know what you think about doing it like this or if you prefer to include those changes here as well.

P.S. `group` is poor choice for names in SQL 😆 it gave a couple of headaches. e.g., I still need to call the first column of the file as `name` since using `group` gave errors with SQL 🙄

abelsiqueira (approved):
Thanks, looks good to me! I agree with doing the other tasks in a separate PR

abelsiqueira:
There is apparently a conflict with `src/constraints/create.jl`

abelsiqueira:
It's approved, so feel free to merge whenever you're ready

PR #976, opened 2024-12-18
Remove unnecessary expression and move lower bound of flow variable from src/constraints/capacity.jl 

abelsiqueira:
- **Remove unnecessary accumulated_investment_limit expression**
- **Move lower bound of flow variable from capacity.jl to variable creation**

Related to #961


datejada (approved):
Reviewed during online meeting

Issue #942, opened 2024-12-04
Change graph data access and constraint creation to use the indices tables


abelsiqueira:
The goal of the issue is to remove most/all data from `graph`. The easiest way to do this will be to find a constraint or similar loop that accesses `graph[...]...` and create the necessary columns in the corresponding DB. Sometimes there is no corresponding DB, so a necessary step is moving that loop to use a DB.

Some things to do:

- Use TulipaIO instead of directly accessing the data
- Inside each function, create a joined table appending more columns to the `indices` tables
- Remove the dependency on data stored in the graph and use the new table instead
- Change loops over sets to loop over rows of the new table

One noteworthy point is that to remove profiles we need to store the `profile_aggregation` in the new table.

After all `graph` access is removed, remove the objects from `graph`, or even `graph` itself.

To avoid becoming a huge PR, we can create smaller PRs or commits per file or per loop. This issue can be worked on by multiple people at the same time then.

The list of files that have `graph[...]` is below. Some might not be applicable for this issue:

- [x] #947
- [x] #943
- [x] Don't get data from graph object in src/constraints/energy.jl
- [x] #977
- [x] #945 
- [x] Don't get data from graph object in src/constraints/ramping-and-unit-commitment.jl
- [x] #955
- [x] Don't get data from graph object in src/constraints/transport.jl
- [ ] Don't get data from graph object in src/expressions/multi-year.jl
- [ ] Don't get data from graph object in src/expressions/storage.jl
- [ ] Don't get data from graph object in src/io.jl
- [ ] Don't get data from graph object in src/model-preparation.jl
- [ ] Don't get data from graph object in src/objective.jl
- [ ] Don't get data from graph object in src/solve-model.jl
- [ ] Don't get data from graph object in src/utils.jl
- [ ] Don't get data from graph object in src/variables/storage.jl

For the `add_*_constraints` functions, the `graph` object should NOT be an argument. The `sets` object should be avoided if possible. If you encounter a place where that is not possible, let's discuss.

PR #979, opened 2024-12-18
Remove graph from src/constraints/energy.jl and transport.jl

abelsiqueira:
- **Remove graph from src/constraints/energy.jl**
- **Remove graph from src/constraints/transport.jl**

Waiting on #978

Related to #942


datejada (approved):
Thanks for the PR. LGTM

PR #980, opened 2024-12-19
Remove graph from src/constraints/ramping-and-unit-commitment.jl

abelsiqueira:
Blocked by #978 and #979 

- Create a field coefficients in TulipaConstraint to allow attaching numeric columns to constraints without modifying the DuckDB table.
- Remove graph from src/constraints/ramping-and-unit-commitment.jl

Part of #942


datejada (approved):
Reviewed during online meeting

PR #987, opened 2025-01-07
Remove need for representative_periods and graph in add_expression_terms_rep_period_constraints

abelsiqueira:
Remove graph and representative_periods from add_expression_terms_rep_period_constraints.
Refactor the function to use the DuckDB tables instead of the DataFrames. 

This now uses DuckDB to compute the groups and the intersection. The type instability made the code a bit slower so to improve, we have to manually specify the types of some outputs with `::Type`. This made the code faster than it was before!

## Related issues

Part of #942

## Checklist

- [x] Tests are passing
- ~~Docs were updated and workflow is passing~~

### DIFF
--- a/src/model-preparation.jl
+++ b/src/model-preparation.jl
@@ -4,11 +4,10 @@ export create_sets, prepare_profiles_structure
 """
     add_expression_terms_rep_period_constraints!(df_cons,
                                                df_flows,
-                                               workspace,
-                                               representative_periods,
-                                               graph;
+                                               workspace;
                                                use_highest_resolution = true,
                                                multiply_by_duration = true,
+                                               add_min_outgoing_flow_duration = false,
                                                )
 
 Computes the incoming and outgoing expressions per row of df_cons for the constraints
@@ -21,22 +20,14 @@ This strategy is based on the replies in this discourse thread:
   - https://discourse.julialang.org/t/help-improving-the-speed-of-a-dataframes-operation/107615/23
 """
 function add_expression_terms_rep_period_constraints!(
+    connection,
     cons::TulipaConstraint,
-    flow::TulipaVariable,
-    workspace,
-    representative_periods,
-    graph;
+    flow::TulipaVariable;
     use_highest_resolution = true,
     multiply_by_duration = true,
     add_min_outgoing_flow_duration = false,
 )
-    # Aggregating function: If the duration should NOT be taken into account, we have to compute unique appearances of the flows.
-    # Otherwise, just use the sum
-    agg = multiply_by_duration ? v -> sum(v) : v -> sum(unique(v))
-
-    grouped_cons = DataFrames.groupby(cons.indices, [:year, :rep_period, :asset])
-
-    # grouped_cons' asset will be matched with either to or from, depending on whether
+    # cons' asset will be matched with flow's to or from, depending on whether
     # we are filling incoming or outgoing flows
     cases = [
         (expr_key = :incoming, asset_match = :to, selected_assets = ["hub", "consumer"]),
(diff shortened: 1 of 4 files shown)


src/model-preparation.jl
```diff
@@ -47,6 +38,57 @@ function add_expression_terms_rep_period_constraints!(
         ),
     ]
     num_rows = size(cons.indices, 1)
+    # TODO: Move this new workspace definition out of this function if and when it's used by other functions
```
datejada, line 41: I think this makes sense since the workspace will be reuse to create the expressions for the binary charging and the unit commitment. But, I agree that for now, let's keep it here to make atomic commits in this PR.

src/model-preparation.jl
```diff
@@ -47,6 +38,57 @@ function add_expression_terms_rep_period_constraints!(
...
+    # 1.1.1.2. Store (var_idx, coefficient) in workspace[timestep]
+    # 1.2. Loop over each constraint in the group: (cons_idx, cons_time_block_start, cons_time_block_end)
+    # 1.2.1. Aggregate all variables in workspace[timestep] for timestep in the
+    #   time block to create a list of variable indices and their coefficients [(var_idx1, coef1), ...]
+    # 1.2.2. Compute the expression using the variable container, the indices and coefficients
+    #
+    # Notes:
+    # - On step 1.2.1, the aggregation can be either by uniqueness of not,
+    #   i.e., if the variable happens in more that one `workspace[timestep]`,
+    #   should we add up the coefficients or not. This is defined by the keyword `multiply_by_duration`
```
datejada, line 69: can we have the numbers in the comments below? so you can read here and then go down and see what is comenting about (e.g., having below 1.2.1 abc...)...Also, maybe this makes more sense to have in the documentation of the function rather than here...What do you think?

src/model-preparation.jl
```diff
@@ -47,6 +38,57 @@ function add_expression_terms_rep_period_constraints!(
...
+        DuckDB.query(
+            connection,
+            "CREATE TEMP TABLE $grouped_cons_table_name AS
+            SELECT
+                cons.asset,
+                cons.year,
+                cons.rep_period,
+                ARRAY_AGG(cons.index) AS index,
+                ARRAY_AGG(cons.time_block_start) AS time_block_start,
+                ARRAY_AGG(cons.time_block_end) AS time_block_end,
```
datejada, line 86: I might need to read more about the 'ARRAY_AGG' function in SQL, but is here guaranting you get and array type if there is no data? I mean, to keep the type stablity...more to learn about it than a comment

src/model-preparation.jl
```diff
@@ -64,60 +106,128 @@ function add_expression_terms_rep_period_constraints!(
...
-            for i in eachindex(workspace)
-                workspace[i] = JuMP.AffExpr(0.0)
-            end
+
+        # The grouped variable table is created below for each case (from=asset, to=asset)
+        grouped_var_table_name = "t_grouped_$(flow.table_name)_match_on_$(case.asset_match)"
+        if !_check_if_table_exists(connection, grouped_var_table_name)
+            DuckDB.query(
+                connection,
+                "CREATE TEMP TABLE $grouped_var_table_name AS
```
datejada, line 115: Although we create some extra temporal tables in the connection, it seems that is faster and less memory consuming than the previous DB. Do you have an intuition why?

src/model-preparation.jl
```diff
@@ -64,60 +106,128 @@ function add_expression_terms_rep_period_constraints!(
...
+                    ARRAY_AGG(var.time_block_start) AS time_block_start,
+                    ARRAY_AGG(var.time_block_end) AS time_block_end,
+                    ARRAY_AGG(var.efficiency) AS efficiency,
+                FROM $(flow.table_name) AS var
+                GROUP BY var.$(case.asset_match), var.year, var.rep_period
+                ",
+            )
+        end
+
+        resolution_query = multiply_by_duration ? "rep_periods_data.resolution" : "1.0::FLOAT8"
```
datejada, line 130: is this ::FLOAT8 part of the type stability requirements, is it correct?

src/model-preparation.jl
```diff
@@ -64,60 +106,128 @@ function add_expression_terms_rep_period_constraints!(
...
+            "SELECT
+                cons.asset,
+                cons.year,
+                cons.rep_period,
+                cons.index AS cons_idx,
+                cons.time_block_start AS cons_time_block_start,
+                cons.time_block_end AS cons_time_block_end,
+                var.index AS var_idx,
+                var.time_block_start AS var_time_block_start,
+                var.time_block_end AS var_time_block_end,
```
datejada, line 144: I like that it is clear the start/end of the constraint and the start/end of the variables. It is readable

datejada:
Thanks @abelsiqueira, I have a couple of comments that are for my understanding of the changes rather than improvements. The SQL is understandable and readable. The best is that it is fast and uses less memory.

src/model-preparation.jl
abelsiqueira: Thanks

src/model-preparation.jl
abelsiqueira: Very good question. I have some doubts, but my intuition is the following.
First, I'm gonna split the work in this function into the data manipulation part and the rest (create expression, aggregating, etc.).

First, data manipulation.

The previous code was creating DataFrames groups and their corresponding sub-DataFrames. This code creates pretty much the same thing, except that instead of a `subdf`, we have `array_agg` for each relevant column.
The previous attempts to make this faster were based on performing the whole computation of intervals in DuckDB, but that required cross joining two large tables before grouping. With the `array_agg`, we manage to group first and then use a simpler left join instead - which is what the Julia code was doing. So, in a way, we are doing the same kind of work in DuckDB as we were in DataFrames, so we remain in the same speed complexity because DuckDB is fast enough.

One noteworthy thing here is that computing the groups/array_agg of constraints and flows is done only once per case because we are creating the temporary tables. So aggregating millions of lines of flows happens only twice (to=asset, from=asset), even though we call the functions a few times. So that probably helps to make DuckDB faster.

Second, the rest.

Changing from `JuMP.add_expression` to the `Dict` was done because the `unique` was slow for type inference reasons, I believe. Maybe it was also slow before, so I just went to check, and apparently, this PR + add_expression is slower, but the previous code + Dict is slower. So `Dict` is the best solution here with DuckDB. PS. I also think that `unique` from before could end being wrong in some special situations, but it's not easy to test, so I won't claim it.

The other changes were type annotations. They are REALLY necessary for the DuckDB code because at least some parts of DataFrames are type stable in this function, and for DuckDB, too many things were not. Adding type annotation to the DataFrames version could probably improve it too, but I'm not sure how much. One big slow part for DuckDB were the arrays, so `row`, `row.time_block_start`, `timestep in time_block`, etc., were all not inferred.

In essence, my intuition is that the data part is pretty much the same, but DuckDB is probably faster, and the rest part is slower for the DuckDB version, but it can be made faster with more type annotations, and even some larger changes.

About the memory consumption. It is possible that Julia doesn't count the memory used by DuckDB for these operations, so it should be taken with a grain of salt. Apart from that, being type stable improves memory, but I'm not sure exactly why, it's a Julia thing.

src/model-preparation.jl
abelsiqueira: Yes. Otherwise it can be interpreted as a different type by DuckDB.

src/model-preparation.jl
abelsiqueira: 🙏

abelsiqueira (commented):
Hi @datejada, thanks for the review and comments. Let me know if my replies answer your questions. I will push updates and re-request a review 👍

datejada (approved):
@abelsiqueira Thanks for the reply to the comments and the changes. I learnt more about SQL and type stability 😄

PR #1000, opened 2025-01-15
Implement solution saving and exporting for refactored code

### LINKED ISSUE #115: Return tables (a.k.a. DuckDB) or dataframes as the output format
Define the output format for downstream analysis and visualization.

@suvayu said: _Learn from HyChain WP2_

### LINKED ISSUE #818: Create outputs for the new results of the variables
Add the following to outputs:
- `assets_decommission_simple_method`
- `assets_decommission_compact_method`
- and decommission for flow, assets_investment_energy (waiting for #792 #806)

abelsiqueira:
- Update function to compute dual
- Create function to save solution and dual values to DuckDB connection
- Update `save_solution_to_file`
- Update tests

## Related issues

Closes #115 
Closes #818 

## Checklist

- [x] Tests are passing

### DIFF
--- a/src/io.jl
+++ b/src/io.jl
@@ -1,4 +1,4 @@
-export create_internal_structures, save_solution_to_file
+export create_internal_structures, export_solution_to_csv_files
 
 """
     graph, representative_periods, timeframe  = create_internal_structures(connection)
@@ -317,16 +317,16 @@ function _create_empty_unless_exists(connection, table_name)
 end
 
 """
-    save_solution_to_file(output_folder, energy_problem)
+    export_solution_to_csv_files(output_folder, energy_problem)
 
 Saves the solution from `energy_problem` in CSV files inside `output_file`.
 Notice that this assumes that the solution has been computed by [`save_solution!`](@ref).
 """
-function save_solution_to_file(output_folder, energy_problem::EnergyProblem)
+function export_solution_to_csv_files(output_folder, energy_problem::EnergyProblem)
     if !energy_problem.solved
         error("The energy_problem has not been solved yet.")
     end
-    save_solution_to_file(
+    export_solution_to_csv_files(
         output_folder,
         energy_problem.db_connection,
         energy_problem.variables,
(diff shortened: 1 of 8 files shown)


src/solve-model.jl
```diff
@@ -1,171 +1,53 @@
...
     if model === nothing
         error("Model is not created, run create_model(energy_problem) first.")
     end
 
-    energy_problem.solution =
-        solve_model!(model, energy_problem.variables, optimizer; parameters = parameters)
+    solve_model(model, optimizer; parameters = parameters)
     energy_problem.termination_status = JuMP.termination_status(model)
-    if energy_problem.solution === nothing
+    if energy_problem.termination_status != JuMP.OPTIMAL
```
datejada, line 22: This option in JuMP is more generic, I propose using it instead of only the optimal condition:

```suggestion
    if !JuMP.is_solved_and_feasible(model)
```

src/solve-model.jl
```diff
@@ -183,24 +65,129 @@ function solve_model(
...
+function save_solution!(connection, model, variables, constraints; compute_duals = true)
+    # Check if it's solved
+    status = JuMP.termination_status(model)
+    if status == JuMP.OPTIMIZE_NOT_CALLED
+        @error("The model was not solved yet, use one of the solve_model functions")
+        return
+    elseif status != JuMP.OPTIMAL
+        @error("The model was not solved to optimality, cannot get solution")
+        return
+    end
```
datejada, line 106: Here the same, I propose simplifying like this:

```suggestion
    if !JuMP.is_solved_and_feasible(model; dual = compute_duals)
        @error("The model has a termination status: $JuMP.termination_status(model), with primal status $JuMP.primal_status(model), and dual status $JuMP.dual_status(model)")
        return
    end
```

src/io.jl
```diff
@@ -320,158 +320,52 @@ end
     save_solution_to_file(output_folder, energy_problem)
 
 Saves the solution from `energy_problem` in CSV files inside `output_file`.
+Notice that this assumes that the solution has been computed by [`save_solution!`](@ref).
 """
 function save_solution_to_file(output_folder, energy_problem::EnergyProblem)
```
datejada, line 325: To differentiate more from the `save_solution` function and account for future options to export to another format, I propose renaming this function as `export_solution_to_csv_files`

datejada (commented):
Thanks for the PR @abelsiqueira; I have left some comments. Let me know what you think.

abelsiqueira:
Thanks for the review! I've updated the PR

datejada (approved):
Thanks for the changes! I don´t have extra comments.

PR #1004, opened 2025-01-22
Refactor many files related to multi-year expressions remove graph and completely removes the sets structure

abelsiqueira:
- Create `TulipaExpression`, similar to `TulipaConstraint`, but for expressions only.
- Remove the concept of simple and compact decommission. There is only a single type of decommission, which is the old compact.
- Update the definitions of accumulated units and related things touching multi-year.
- Update the definition of `accumulated_energy_capacity`.
- Remove `graph` and `sets` from `src/constraints/capacity.jl`
- Remove `graph` and `sets` from `src/expressions/multi-year.jl`
- Remove `graph` and `sets` from `src/expressions/storage.jl`
- Update part of `src/objective.jl` with new expressions.
- Fix ens data to not be decommissionable and set fixed_cost to 0 in multi-year  data.
- Change fixed_cost_storage_energy to 0 in Storage data.
- Fix Variable Resolution test objective value to account for energy capacity
- Remove graph from `src/objective.jl`
- Remove sets completely
- Rename accumulated to available everywhere

Part of #942


src/expressions/storage.jl
```diff
@@ -1,47 +1,91 @@
...
+            ARRAY_AGG(expr_acc.index) AS acc_indices,
+        FROM asset_milestone
+        LEFT JOIN asset
+            ON asset_milestone.asset = asset.asset
+        LEFT JOIN expr_accumulated_units AS expr_acc
+            ON asset_milestone.asset = expr_acc.asset
+            AND asset_milestone.milestone_year = expr_acc.milestone_year
+        GROUP BY
+            asset_milestone.asset,
+            asset_milestone.milestone_year
```
abelsiqueira, line 20: `WHERE asset.type = 'storage'`

src/constraints/storage.jl
```diff
@@ -248,6 +258,7 @@ function _append_storage_data_to_indices(connection, table_name)
             inflows_profile.profile_name AS inflows_profile_name,
             max_storage_level_profile.profile_name AS max_storage_level_profile_name,
             min_storage_level_profile.profile_name AS min_storage_level_profile_name,
+            expr_acc.index AS acc_energy_capacity_index
```
abelsiqueira, line 261: Note to self, check if you're missing a loop over matching indices

src/model-parameters.jl
```diff
@@ -59,3 +59,24 @@ function ModelParameters(connection::DuckDB.DB, path::String = ""; kwargs...)
...
+#             discount_rate REAL,
+#             discount_year INTEGER,
+#         );
+#         INSERT INTO model_parameters VALUES (
+#             $discount_rate,
+#             $discount_year,
+#         );
+#         ",
+#     )
+# end
```
abelsiqueira, line 82:
```suggestion
```

datejada:
> 169

Maybe we can address this slowdown here, #1021. What do you think? @abelsiqueira

test/test-case-studies.jl
```diff
@@ -68,7 +68,7 @@ end
     connection = DBInterface.connect(DuckDB.DB)
     _read_csv_folder(connection, dir)
     energy_problem = TulipaEnergyModel.run_scenario(connection)
-    @test energy_problem.objective_value ≈ 28.45872 atol = 1e-5
+    @test energy_problem.objective_value ≈ 778.45872 atol = 1e-5
```
datejada, line 71: @abelsiqueira, this update looks strange to me since the tiny variable resolution case study does not have investments. Maybe other parameter needs to be updated...but I would be surprised if there is this change in the objective function...
And I think, it is due to the `fixed_cost_storage_energy` that has a value of 5, but since this case is not investment, it should be 0.
Please double-check. Thanks!

src/variables/investments.jl
```diff
@@ -84,8 +77,8 @@ function add_investment_variables!(model, graph, sets, variables)
             row -> row.investment_integer_storage_energy,
         ),
         (
-            :assets_decommission_energy_simple_method,
-            row -> (row.milestone_year, row.asset),
+            :assets_decommission_energy,
```
datejada, line 80: @gnawin maybe we need this case in the multi-year investment to cover for it. It seems we don't have storage with storage decommission. But, we should cover it, since, in principle we should be able to have the compact for all assets and types of investments.

src/variables/create.jl
```diff
@@ -269,19 +257,23 @@ function _create_variables_tables(connection)
...
-        FROM asset_milestone
+        FROM asset_both
         LEFT JOIN asset
-            ON asset.asset = asset_milestone.asset
+            ON asset.asset = asset_both.asset
         WHERE
             asset.storage_method_energy = true
             AND asset.type = 'storage'
-            AND asset.investment_method = 'simple'",
+            AND asset.investment_method = 'simple' -- TODO: Keep this or not?
```
datejada, line 273: I think all related to simplicity should be discarded. @gnawin, what do you think? This part of the code should be uniform for all assets and types of investments.

src/variables/investments.jl
```diff
@@ -52,26 +52,19 @@ function add_investment_variables!(model, graph, sets, variables)
...
             _ -> 0.0,
             _ -> Inf,
             row -> row.investment_integer,
         ),
         (
-            :flows_decommission_using_simple_method,
-            row -> (row.milestone_year, (row.from_asset, row.to_asset)),
+            :flows_decommission,
+            row -> ((row.from_asset, row.to_asset), row.milestone_year, row.commission_year),
             _ -> 0.0,
```
gnawin, line 64: Does this mean we create an explicit lower bound for this variable, does this create a constraint (for the lower bound)?

src/variables/investments.jl
```diff
@@ -52,26 +52,19 @@ function add_investment_variables!(model, graph, sets, variables)
...
             _ -> Inf,
             _ -> false,
```
gnawin, line 66:
```suggestion
            _ -> row.investment_integer,
```

`flows_decommission` should also depend on the `investment_integer`, same as the investment.

src/structures.jl
```diff
@@ -39,6 +39,7 @@ mutable struct TulipaVariable
     indices::DataFrame
     table_name::String
     container::Vector{JuMP.VariableRef}
+    expressions::Dict{Symbol,Vector{JuMP.AffExpr}}
```
gnawin, line 42: What is the difference between `TulipaVariable.expressions` and `TulipaExpression.expressions`? I see mostly (if not all) the latter is used.

gnawin (changes requested):

test/test-case-studies.jl
gnawin: @datejada yes, the difference is due to the `fixed_cost_storage_energy`, @abelsiqueira and I checked this issue last time.

I recall our conclusion was that this cost `fixed_cost_storage_energy` is considered no matter the investment method, i.e., if you have capacity (even if not investment) then you pay fixed costs. 

I check the older versions why this cost was not taken into account, and I figured out that this cost was only considered when there is an investment method. I think this was not correct.

We can either change this obj value, or to make the data to 0, just what we did for the storage case study. Shall we change the data instead? @abelsiqueira @datejada

abelsiqueira: Tagging @gnawin

src/variables/create.jl
gnawin: @datejada Yes, I agree.

test/test-case-studies.jl
abelsiqueira: Btw, it does come from `fixed_cost_storage_energy = 5` times `capacity_storage_energy = 150`.

datejada: Great! all clear 😄 So, please change the data to have `fixed_cost_storage_energy = 0`
Thanks!

src/variables/investments.jl
abelsiqueira: This is equivalent to the `JuMP.set_lower_bound` function, so it simplifies the creation of the lower bound of the variable. It will not create an explicit constraint.

src/variables/investments.jl
abelsiqueira: Fixed locally, please check again after update.

src/structures.jl
abelsiqueira: This is a left over from an attempt to solve the accumulated expressions. I forgot to remove it. There are only `TulipaExpression.expressions` and `TulipaConstraint.expressions`. The only difference is that the latter has constraints information (names and containers). It should be possible to generalise, but that is far the current issue.

src/variables/investments.jl
gnawin: I see. Indeed this happens because we don't have a case for `storage_energy_method==true`. There is battery but it is set to false, so it is only the capacity. Maybe we can change the data to true to see what happens, but it may influence the test as a whole, which we would have to check.

@abelsiqueira would you mind changing the data to see what happens? Thanks a lot!

abelsiqueira:
Hi both, thanks for the comments. I replied in the appropriate comment, and here are the comments to the reviews:

Replying to @datejada: 
> @abelsiqueira, thanks for the PR; great work here 😄 I left a couple of comments, but here is a recap:
> 
> 1. the objective function of the tiny variable temporal resolution should not change, so it seems easy to fix it from the data, but let me know.
> 2. the storage decommission should be the same as for the other ones, yo keep it consistent. So, there is no "simple" there. But, please @gnawin check if I am wrong. In any case, I think the "simple" reference in the query should go away.

@gnawin, I tagged you on 1 to decide what is correct with @datejada.
For 2, I will remove "simple" and let you know what happens.

> Maybe we can address this slowdown here, https://github.com/TulipaEnergy/TulipaEnergyModel.jl/issues/1021. What do you think? @abelsiqueira

If the issue is the slowdown of this PR, regardless of #1021, I can look into it this week instead of my previous plan. If you already found a place that we're using the wrong thing, please point it out and I'll try to fix. But if you're suggesting investigating #1021 here, then I think it's better to wait after merging, because it can be done in parallel with the rest of my planned work. 

Replying to @gnawin:
> Just some minor issues (see comments), one regarding the legacy of simple here and there. I think the best is the remove it, because now everything is compact (right? because I see simple is mentioned sometimes). As we discussed, in the future we will add an simple version, but probably we'll build upon the current foundation (we would then need the differentiation between simple and compact but most probably we don't need to worry about it in the version

Indeed, I removed `simple` from where I forgot - except for the formulation in the docs, of course. I addressed the other comments in the last commit, hopefully.

src/variables/create.jl
abelsiqueira: I removed this

test/test-case-studies.jl
abelsiqueira: Done

src/variables/investments.jl
datejada: Hmm...still the codecov complains, so it might be something else. But, I will leave @gnawin the last decision on whether we can want to dig into why is not creating this variable or not in this PR

datejada:
> Hi both, thanks for the comments. I replied in the appropriate comment, and here are the comments to the reviews:
> 
> Replying to @datejada:
> 
> > @abelsiqueira, thanks for the PR; great work here 😄 I left a couple of comments, but here is a recap:
> > 
> > 1. the objective function of the tiny variable temporal resolution should not change, so it seems easy to fix it from the data, but let me know.
> > 2. the storage decommission should be the same as for the other ones, yo keep it consistent. So, there is no "simple" there. But, please @gnawin check if I am wrong. In any case, I think the "simple" reference in the query should go away.
> 
> @gnawin, I tagged you on 1 to decide what is correct with @datejada. For 2, I will remove "simple" and let you know what happens.
> 
> > Maybe we can address this slowdown here, #1021. What do you think? @abelsiqueira
> 
> If the issue is the slowdown of this PR, regardless of #1021, I can look into it this week instead of my previous plan. If you already found a place that we're using the wrong thing, please point it out and I'll try to fix. But if you're suggesting investigating #1021 here, then I think it's better to wait after merging, because it can be done in parallel with the rest of my planned work.
> 
> Replying to @gnawin:
> 
> > Just some minor issues (see comments), one regarding the legacy of simple here and there. I think the best is the remove it, because now everything is compact (right? because I see simple is mentioned sometimes). As we discussed, in the future we will add an simple version, but probably we'll build upon the current foundation (we would then need the differentiation between simple and compact but most probably we don't need to worry about it in the version
> 
> Indeed, I removed `simple` from where I forgot - except for the formulation in the docs, of course. I addressed the other comments in the last commit, hopefully.

Great! Thanks. Regarding the improvements, I suggest exploring the speedups on #1021 rather than here, so yes, after merging this PR.

datejada (approved):
Hi @abelsiqueira, Thanks for the updates. So, I don't have more comments for you. So, approved from my side.
I am leaving the decision on whether to change or not the objective function of the multi-year investment to @gnawin, depending on what he thinks is best at this point. We can always merge now and create an issue to investigate that specific case of the codecov.

gnawin (approved):

PR #1005, opened 2025-01-27
Update .all-contributorsrc and run generate changing some links

abelsiqueira:
I've changed the links of Juha and Matthijs to be the GitHub link to help the link checker.
I changed it on the `.all-contributorsrc` file and ran the all contributors generator locally. It ended changing more than I was expecting, but apparently these links are correct and the old ones were not.
Next time we update the all-contributors file, we see how much it changes and decide how to proceed.

Also, it looks like the ieeexplore.ieee.org site is down, but that is not necessarily something that happens often.

## Related issues

There is no related issue.

## Checklist

- [x] Tests are passing


datejada (approved):
Thanks for the changes, hopefully the links won't give more problems in the future 😊

PR #1012, opened 2025-02-04
Add explicit asset timeframe partitions

### LINKED ISSUE #1009: Add default values for `assets_timeframe_partitions` table
### Description

The table created from assets-timeframe-partitions.csv has no defaults, while the assets-rep-periods-partitions.csv uses a default of "uniform" and partition = 1. So, this confuses the user.

### Validation and testing

The assets-timeframe-partitions.csv should be created with default "uniform" and partition = 1 for the assets defined as seasonal (`is_seasonal = true`). As it is done in the partitions for representative periods.

### Motivation

To have the same criteria in both files to avoid confusion.

### Target audience

developers

### Can you help?

_No response_

datejada:
This PR unifies the default value of the partitions for the representatives and the timeframe. So, if no data is defined in the partition files, the default partition is `uniform` of value `1`.

## Related issues

Closes #1009

## Checklist

- [ ] Tests are passing

### DIFF
--- a/src/tmp.jl
+++ b/src/tmp.jl
@@ -193,6 +193,25 @@ function tmp_create_partition_tables(connection)
 
     DBInterface.execute(
         connection,
+        "CREATE OR REPLACE TABLE explicit_assets_timeframe_partitions AS
+        SELECT
+            asset.asset,
+            asset_commission.commission_year AS year,
+            COALESCE(atp.specification, 'uniform') AS specification,
+            COALESCE(atp.partition, '1') AS partition,
+        FROM asset AS asset
+        LEFT JOIN asset_commission
+            ON asset.asset = asset_commission.asset
+        LEFT JOIN assets_timeframe_partitions AS atp
+            ON asset.asset = atp.asset
+            AND asset_commission.commission_year = atp.year
+        WHERE
+            asset.is_seasonal = true
+        ",
+    )
+
+    DBInterface.execute(
+        connection,
         "CREATE OR REPLACE TABLE asset_timeframe_time_resolution(
             asset STRING,
             year INT,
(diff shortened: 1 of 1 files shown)


datejada:
@abelsiqueira so, from the benchmark we can see the performance problem when we have several representatives and we use the constraints over clustered year (inter-temporal constraints) by defining the timeframe partitions.

I will create a PR with the equivalent problem in version 0.10.4 to have a reference.

Also, Joaquim and Oscar recommended PProf.jl find type instabilities and bottlenecks in the code. I will give it a try on a different PR.

I will keep you posted

src/tmp.jl
```diff
@@ -191,6 +191,25 @@ function tmp_create_partition_tables(connection)
...
+            COALESCE(atp.specification, 'uniform') AS specification,
+            COALESCE(atp.partition, '1') AS partition,
+        FROM asset AS asset
+        LEFT JOIN asset_commission
+            ON asset.asset = asset_commission.asset
+        LEFT JOIN assets_timeframe_partitions AS atp
+            ON asset.asset = atp.asset
+            AND asset_commission.commission_year = atp.year
+        WHERE
+            asset.is_seasonal = true
```
abelsiqueira, line 209:
```suggestion
            asset.is_seasonal
```

abelsiqueira (approved):
Thanks, small simple change, pre-approved.
It is a bit concerning that this doesn't break the tests at all. It's either not affecting anything, or affecting things that don't change anything. Either way, it's another thing we have to create tests for.
Maybe we modify `Norse` to remove one of the rows in `assets-timeframe-partitions.csv`?

datejada:
> Thanks, small simple change, pre-approved. It is a bit concerning that this doesn't break the tests at all. It's either not affecting anything, or affecting things that don't change anything. Either way, it's another thing we have to create tests for. Maybe we modify `Norse` to remove one of the rows in `assets-timeframe-partitions.csv`?

Thanks @abelsiqueira! I have updated the Storage case since it used the `uniform` partition = 1, so it does not change anything in the case study. In addition, I have updated the docs to mention the defaults and files that are allowed to be missing, and I have also updated the concepts section as a consequence.

Once the tests pass, I will merge them; thanks!

PR #1017, opened 2025-02-07
Improve performance to create constraints over clustered year

abelsiqueira:
This is built on top of #1015.

Closes https://github.com/TulipaEnergy/TulipaEnergyModel.jl/issues/1014

Notes:

- This is essentially the refactor for the relevant function. So `graph` and other similar things were removed in favour of DuckDB.
- The idea is pretty much the same as in the similar rep_periods function. The main differences are:
  - Now, the constraints are defined over **period** blocks, because it is over clustered year.
  - This means we need the `rep_periods_mapping` table (and friends) to join the **period** from constraints with the **rep_period** from variables taking into account the **weight**.
  - This create a two-level group for the main loop, one for `(asset, year)` and internally for `rep_period`.
- In addition to the expressions, the `:inflows_profile_aggregation` is also computed here. Although, in the end the solution I found was to create a completely separate query to aggregate the profile - and probably it doesn't make sense to compute this in this function (unless we need to make it faster and figure a way to join them), because the inflows don't require any flow information.
- I have no idea if this is correct!!! The tests are passing, but they were also passing for known failing intermediary states. In the last version, I compare the **generated LP files as well, and they are slightly different**, but as far as I can see, only on the inflow value for an error of ~1e-12, i.e., it appears to be a normal numerical precision difference.
- Locally, the PR head of #1015 takes around **230 seconds and this is taking around 5 seconds, compared only with `@time`**, but I would prefer that you test in your normal workflow.
- The benchmarks failed with exceeded time for main. I don't remember which version of the file it uses, but I assumed it was the newest. Maybe I'm wrong, or maybe main is just much worse, or GitHub's runners are much slower. Anyway, up to discussion.


datejada (commented):
Hi @abelsiqueira, thanks for the PR. I tested it and we get the same .lp file for 100rp comparing this PR and PR #1015. The SQL queries can e followed, so that is good, since they follow the same type of structure as the others in the code. I also like that the inflows aggregation is not an expression, it is better as a coefficients (maybe the same is useful for other aggregations, like the availability and demand profile aggregation).

I ran the benchmark locally with 100rp and the times to create the model are:
- PR #1015 1167s
- This PR 28.6s

My main comment is that in these changes the JuMP.add_to_expression! is lost in the code, and I think we still can gain more with change to (or am I not seeing/understanding something here)

Finally, since the inflow profiles are calculated in the coefficients, it is better to clean the function and delete the commented parts that are not used anymore (even some TODOs might be obsolete).

Thanks again, great work here!

Diego

src/model-preparation.jl
```diff
@@ -356,95 +357,284 @@ This function is only used internally in the model.
...
+                        continue
+                    end
+                    # Note to future. Using `mergewith!` did not work because the
+                    # `combine` function is only applied for clashing entries, i.e., the weight is not applied uniformly to all entries.
+                    # It passed for most cases, since `weight = 1` in most cases.
+                    for (var_idx, var_coef) in group_flows_accumulation
+                        if !haskey(flows_per_period_workspace[period], var_idx)
+                            flows_per_period_workspace[period][var_idx] = 0.0
+                        end
+                        flows_per_period_workspace[period][var_idx] += var_coef * weight
```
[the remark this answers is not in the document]
datejada: And same in line 223 in the rp aggregation. I think we also there can get rid of the += using the JuMP function (as recommended by Oscar and Joaquim)

src/model-preparation.jl
```diff
@@ -3,6 +3,7 @@ export create_sets, prepare_profiles_structure
 
 """
     add_expression_terms_rep_period_constraints!(df_cons,
+using Base: Workqueues
```
abelsiqueira, line 6:
```suggestion
```

src/model-preparation.jl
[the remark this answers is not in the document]
abelsiqueira: `flows_per_period_workspace` is not a JuMP expression, it is a dictionary per period of the variable index to the variable coefficient, see line 418.

The logic is to aggregate in dictionaries before summing if over. I.e., instead of a workspace of JuMP expressions, the workspace is of pairs `(idx, coef)`. Then, down in line 572, we aggregate these dictionaries with `mergewith` and on lines 578 we use the variables in a sum over this dictionary which creates the expression.
This removes the need for `JuMP.add_to_expression!`. In the rp case, there were other motivations to make the switch (aggregation could be sum or sum(unique), which returns in difference aggregation logic which doesn't work in all corner cases with the expressions. But it was also slightly faster.

benchmark/benchmarks.jl
```diff
@@ -31,8 +31,8 @@ energy_problem = EnergyProblem(connection)
 
 SUITE["energy_problem"]["create_model"] = @benchmarkable begin
     create_model!($energy_problem)
-end samples = 3 evals = 1 seconds = 86400
-create_model!(energy_problem)
+end samples = 2 evals = 1 seconds = 86400
```
abelsiqueira, line 34:
```suggestion
end samples = 3 evals = 1 seconds = 86400
```
Since it's not running, might as well use the default.

src/model-preparation.jl
```diff
@@ -356,95 +357,284 @@ This function is only used internally in the model.
...
+
+    # The flow_per_period_workspace holds the list of flows that will be aggregated in a given period
+    maximum_num_periods = only(
+        row[1] for row in DuckDB.query(connection, "SELECT MAX(period) FROM rep_periods_mapping")
+    )::Int32
+    flows_per_period_workspace = [Dict{Int,Float64}() for _ in 1:maximum_num_periods]
+    # inflows_per_period = zeros(maximum_num_periods)
+    # TODO: The computation of inflows_profile_aggregation is done separately using DuckDB alone at the end of this file.
+    # The code to perform the computation together with the computation of the expressions is commented out (search inflows).
+    # It was not working, and my last guess is that it related to missing flows.
```
abelsiqueira, line 422:
```suggestion
```

src/model-preparation.jl
```diff
@@ -356,95 +357,284 @@ This function is only used internally in the model.
...
+
+    for case in cases
+        from_or_to = case.asset_match
+        # update_inflows = case.expr_key == :incoming
```
abelsiqueira, line 426:
```suggestion
```

src/model-preparation.jl
```diff
@@ -356,95 +357,284 @@ This function is only used internally in the model.
...
+                AND cons.year = profiles.commission_year
+                AND profile_type = 'inflows'
+            GROUP BY cons.asset, cons.year;
+            FROM t_groups
+            ",
+        )
+            empty!.(flows_per_period_workspace)
+            # inflows_per_period .= 0.0
+
+            # Problem! The flow might be not present but the asset should still be computed
```
abelsiqueira, line 490:
```suggestion
```

src/model-preparation.jl
```diff
@@ -356,95 +357,284 @@ This function is only used internally in the model.
...
+                ARRAY_AGG(COALESCE(var.index, []) ORDER BY var.rep_period) AS var_indices,
+                ARRAY_AGG(COALESCE(var.time_block_start, []) ORDER BY var.rep_period) AS var_time_block_start_vec,
+                ARRAY_AGG(COALESCE(var.time_block_end, []) ORDER BY var.rep_period) AS var_time_block_end_vec,
+                ARRAY_AGG(COALESCE(var.efficiency, []) ORDER BY var.rep_period) AS var_efficiencies,
+                ARRAY_AGG(var.rep_period ORDER BY var.rep_period) AS var_rep_periods,
+                ARRAY_AGG(rpdata.num_timesteps ORDER BY var.rep_period) AS num_timesteps,
+                ARRAY_AGG(asset_milestone.storage_inflows ORDER BY var.rep_period) AS storage_inflows,
+                ARRAY_AGG(COALESCE(rpmap.periods, []) ORDER BY var.rep_period) AS var_periods,
+                ARRAY_AGG(COALESCE(rpmap.weights, []) ORDER BY var.rep_period) AS var_weights,
+                ANY_VALUE(profiles.profile_name) AS profile_name,
```
abelsiqueira, line 465:
```suggestion
```

src/model-preparation.jl
```diff
@@ -356,95 +357,284 @@ This function is only used internally in the model.
...
+            LEFT JOIN rep_periods_data AS rpdata
+                ON rpdata.year = cons.year
+                AND rpdata.rep_period = var.rep_period
+            LEFT JOIN asset_milestone
+                ON asset_milestone.asset = cons.asset
+                AND asset_milestone.milestone_year = cons.year
+            LEFT OUTER JOIN assets_profiles AS profiles
+                ON cons.asset = profiles.asset
+                AND cons.year = profiles.commission_year
+                AND profile_type = 'inflows'
```
abelsiqueira, line 482:
```suggestion
```

src/model-preparation.jl
```diff
@@ -356,95 +357,284 @@ This function is only used internally in the model.
...
+                    cons.expressions[case.expr_key][cons_idx] = sum(
+                        coefficient * flow.container[var_idx] for
+                        (var_idx, coefficient) in workspace_aggregation
+                    )
+                end
+
+                # if update_inflows
+                #     cons.coefficients[:inflows_profile_aggregation][cons_idx] +=
+                #         sum(inflows_per_period[period_block])
+                # end
```
abelsiqueira, line 587:
```suggestion
```

abelsiqueira:
Hi Diego, I've replied to your `JuMP.add_expression` comments, but in summary, we don't need them anymore because we don't have a workspace of expressions anymore.

On doing the same aggregation for everyone (or the opposite, not doing for any of them), it's a good question that should be investigate. Can you create an issue for that?

I removed the other comments that I could see from this PR page and reduced the DuckDB query to remove the LEFT JOIN on the profiles. Hopefully the code still works.

This should address all comments, but if it doesn't, feel free to pick it up while I'm back on the multi-year refactor. But if you do need me, feel free to ping me.

src/model-preparation.jl
```diff
@@ -356,95 +357,284 @@ This function is only used internally in the model.
...
-
-            sub_df_flows = filter(
-                [:from, :year, :rep_period] =>
-                    (from, y, rp) ->
-                        (from, y, rp) == (row_cons.asset, row_map.year, row_map.rep_period),
-                flow.indices;
-                view = true,
+        grouped_var_table_name = "t_grouped_$(flow.table_name)_match_on_$(from_or_to)"
+        if !_check_if_table_exists(connection, grouped_var_table_name)
+            DuckDB.query(
```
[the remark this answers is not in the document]
datejada: Thanks! The answer to the question is, no, it is not possible. The RP version will always call for the timeframe constraints. So, the check and an error message should be the way to go here.

datejada:
> Hi Diego, I've replied to your `JuMP.add_expression` comments, but in summary, we don't need them anymore because we don't have a workspace of expressions anymore.
> 
> On doing the same aggregation for everyone (or the opposite, not doing for any of them), it's a good question that should be investigate. Can you create an issue for that?
> 
> I removed the other comments that I could see from this PR page and reduced the DuckDB query to remove the LEFT JOIN on the profiles. Hopefully the code still works.
> 
> This should address all comments, but if it doesn't, feel free to pick it up while I'm back on the multi-year refactor. But if you do need me, feel free to ping me.

Thank you for the reply and explanation. My question also arose when I saw this message in the Julia console using this code (for hourly 36rp):

´´´julia
┌ Warning: The addition operator has been used on JuMP expressions a large number of times. This warning is safe to ignore but may indicate that model generation is slower than necessary. For performance reasons, you should not add expressions in a loop. Instead of x += y, use add_to_expression!(x,y) to modify x in place. If y is a single variable, you may also use add_to_expression!(x, coef, y) for x += coef*y.
´´´

So, if it is not this part, then I wonder where else we have the `+=` such that JuMP complains 🤔  Any ideas?

To reproduce the warning, use the files in the benchmark and before creating the model, use the following command to change all the partitions to 1 hours:

```julia
DuckDB.query(connection, "UPDATE assets_rep_periods_partitions SET partition = 1")
DuckDB.query(connection, "UPDATE flows_rep_periods_partitions SET partition = 1")
```

Thanks!

abelsiqueira:
No ideas, but indeed there is something somewhere. I'm probably not touching this until Monday, though.

src/model-preparation.jl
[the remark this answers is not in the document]
datejada: Thanks! it makes sense 😃

src/model-preparation.jl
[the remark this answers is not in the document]
datejada: @abelsiqueira Hi, I change the code here to have the error check, but it looks a bit ugly, so, feel free to suggest some changes. Thanks!

datejada:
> No ideas, but indeed there is something somewhere. I'm probably not touching this until Monday, though.

great! thanks, since it seems it is happening in another part of the code, then we can address it in another issue and merge this one first.

datejada (approved):
@abelsiqueira, the PR is approved, feel free to merge it once you have looked at the error message in the code.

datejada:
@abelsiqueira, I forgot to mention: I am running the versions' benchmarks locally to get a full picture of improvements. That is why I am not benchmarking this PR anymore and returning the files to the 100rp benchmark.

abelsiqueira:
Hi @datejada, I've updated your error message a bit. I have not found any reason for the expression warning yet. I looked into the `+=` and `sum` in that file, and they don't appear to be dealing with expressions (we do create expressions summing a large number of Julia variables, though, could that be the same warning?).
I wrote a message on slack, in case someone has a suggestions on how to easily find out where the warning originates?

abelsiqueira:
I will merge this now. Thanks for the help.

PR #1018, opened 2025-02-14
Change sum(coef * var) to JuMP.add_to_expression

abelsiqueira:
@datejada identified the warning for summation generating expressions, but I could not figure out where. @matbesancon helped figure out that `sum(coef * var)` also falls into that situation.
A more visually appealing solution would be changing that to `dot(coeffs, vars)`, but we would need to create the vectors for that, so we resort to `JuMP.add_to_expression!`.
Hopefully the benchmark should reveal some speed up. 

## Related issues

Related to #1014, missing warning.

## Checklist

- [x] Tests are passing

### DIFF
--- a/src/model-preparation.jl
+++ b/src/model-preparation.jl
@@ -225,9 +225,11 @@ function add_expression_terms_rep_period_constraints!(
                 end
                 if length(workspace_agg) > 0
                     # Step 1.2.2.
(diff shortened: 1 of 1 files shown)


abelsiqueira:
Great success! Thanks @matbesancon! And great job finding the issue @datejada!

datejada (approved):
Awesome! Thanks, that is an extra speedup ⏩

PR #1023, opened 2025-02-17
Remove graph and other internal structures and general cleaning

### LINKED ISSUE #942: Change graph data access and constraint creation to use the indices tables

abelsiqueira:
Waiting for #1004 

- Remove graph from remaining places and delete all mentions to it
- Update various docstrings
- Remove older code and comments

## Related issues

Closes #942 

## Checklist

- [x] Tests are passing


datejada (approved):
Wow! It's so satisfying to clean up the code. Thanks!

PR #1032, opened 2025-02-19
Always saving the solution when using run scenario

datejada:
This pull request includes changes in the `run_scenario` function. The changes focus on improving the names in the timing and always saving the solution when using this function. Since this function is a wrapper of the individual steps, keeping the solution should be part of the steps while exporting should remain optional.

Also there is improvements in the docstring of the funtion.

## Related issues

There is no related issue.

## Checklist

- [x] Tests are passing

### DIFF
--- a/src/run-scenario.jl
+++ b/src/run-scenario.jl
@@ -31,10 +31,11 @@ function run_scenario(
 
     @timeit to "create_model!" create_model!(energy_problem; write_lp_file, enable_names)
 
-    @timeit to "solve and store solution" solve_model!(energy_problem, optimizer; parameters)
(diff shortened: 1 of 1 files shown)


datejada:
Cbc is given an error when computing the duals 😢 

I will try to see if there is a workaround 🤷

abelsiqueira:
I would rather just disable Cbc definitely

datejada:
I agree, it gives too many headaches; first the tests in Mac, and now this :/ 

Thanks! I will include those changes here

abelsiqueira (approved):
Thanks!

PR #1035, opened 2025-02-20
Remove DataFrames completely

abelsiqueira:
Now that DuckDB is being used almost everywhere, we can finish the job and remove DataFrames.
This mostly means:

- TulipaVariables (and friends) field `indices` is now a DuckDB thing. To loop over it we don't need to use `for row in eachrow(indices)`, just `for row in indices`.
- Some things are harder to do, since they **must** use SQL now. Namely, subsets.
- There should be some memory and speed improvement.

It should also be possible to do the following improvements, which can be considered "nice to have":

- Don't even store `indices`, instead store `connection` and when it is desirable to loop, call the query to create the looping object. I'm not sure this is actually necessary or useful, because I believe the DuckDB `indices` will never be store materialised. So in practice it should be the same efficiency/memory use.
- Make loops look nicer by creating an API to loop over the rows and the container/expressions at the same time. For instance, instead of doing
  ```
  [... for (row, expr1, expr2) in zip(cons.indices, cons.expressions[:expr1], cons.expressions[:expr2])]
  ```
  we can create a function to wrap that:
  ```
  [... for (row, expr1, expr2) in loop_over(cons, with_expr[:expr1, :expr2])
  ```
  Some thought is necessary to ensure (some) type stability.

## Related issues

Related to #701 

## Checklist

- [x] Tests are passing


src/constraints/storage.jl
```diff
@@ -44,19 +44,17 @@ function add_storage_constraints!(connection, model, variables, expressions, con
...
+                            last_index = only([
+                                row[1] for row in DuckDB.query(
+                                    connection,
+                                    "SELECT
+                                        MAX(index)
+                                    FROM cons_$table_name
+                                    WHERE asset = '$(row.asset)' AND year = $(row.year) AND rep_period = $(row.rep_period)
+                                    ",
+                                )
+                            ])::Int
```
datejada, line 57: This comment is for my understanding: `::Int` is here to ensure type stability since we know in advance that the result will be an Int, but Julia can't know it in advance because it comes from a SQL query. Is that correct?

datejada (approved):
Thanks for the PR. The code looks cleaner, and deleting dependencies is good 😄 The code was also initially review on the Tulipa working day. I left a comment for my learning porpuse, so that it is why it is approved now.

src/constraints/storage.jl
abelsiqueira: Exactly. The draft PR #1055 will add more of these.

Issue #1039, opened 2025-02-21
Put the defaults in a JSON file


clizbe:
- [x] #992
- [ ] Fix Tulipa so it uses the file (untangle OBZ pipeline)
- [x] Tell the user how to overwrite the defaults

Start with the defaults file Diego has drafted here in function get_default_values:
https://github.com/TulipaEnergy/Tulipa-OBZ-CaseStudy/blob/main/utils/functions.jl

Also consider:
- #461

datejada:
One option is to have a JSON file that has the schema, description of the parameters, recommended units, and defaults. Issue #992 will also help with #1038 since it can be reused to print it in the docs.

clizbe:
Yes I looked up how one of the solvers stored their defaults, and it was something like this. Although I think theirs were just in a markdown file.

clizbe:
Closed by #1054.
Still need to make features explicit so it doesn't just infer if data is present/missing, but that continues with #996

PR #1054, opened 2025-02-26
Create model schema from a JSON file

### LINKED ISSUE #992: Store schema in a JSON file
### Description

This will help add helpful information, such as a description of the parameter and suggested units, to the schema file.

- [ ] Create a JSON file with the schema in file [input-schemas.jl](https://github.com/TulipaEnergy/TulipaEnergyModel.jl/blob/main/src/input-schemas.jl) 
- [ ] Change the code to read the schemas of the input files
- [ ] Use the schema file to add to the documentation a nice table with it's information

### Validation and testing

Add a test to read the schema file

### Motivation

We can use the file add the info in the docs and use it in other languages to generate the Tulipa Tables/files with the correct schema.

### Target audience

_No response_

### Can you help?

_No response_

datejada:
This pull request includes the main changes to read the schema from a JSON file, which also includes a description, units, default, and constraints of the parameters in the model. The main changes are:

Configuration Management:

* [`.pre-commit-config.yaml`](diffhunk://#diff-63a9c44a44acf85fea213a857769990937107cf072831e1a26808cfde9d096b9L36-L41): Removed the Prettier hook to streamline the pre-commit configuration.

Code Enhancements:

* [`src/TulipaEnergyModel.jl`](diffhunk://#diff-b865087258345b8a7313140b34fffef927393317cfa1a336c0f2b01e38d7f862R10): Added the `JSON` module import to extend the functionality of the script.

Schema Handling:

* [`src/input-schemas.jl`](diffhunk://#diff-d4fa2bdc071734808bbba177dbceb6c24f9a89219632e2a1f7d45f1860472380L1-R49): Refactored the schema definitions to read from a JSON file instead of hardcoding them in the script. This change simplifies the maintenance and updates of schema definitions.

Documentation updates:

* [`docs/src/50-schemas.md`](diffhunk://#diff-a28611b655f56239d20252cec35a63df8927c81cc6771b61fd68ebb1a52a4b90L5-R5): Updated the documentation to inform users that schemas can now be found in the `input-schemas.json` file and provided an example code snippet for generating Markdown documentation from the JSON schema. [[1]](diffhunk://#diff-a28611b655f56239d20252cec35a63df8927c81cc6771b61fd68ebb1a52a4b90L5-R5) [[2]](diffhunk://#diff-a28611b655f56239d20252cec35a63df8927c81cc6771b61fd68ebb1a52a4b90L14-R65)

## Related issues

Closes #992, #1039 

## Checklist

- [x] Tests are passing


datejada:
@suvayu and @abelsiqueira these are the changes to have the schema (and more) in a JSON file. I have updated the documentation to create the parameter section from this file, including the units of measurement, description, constraints, and default values. Your comments and suggestions are more than welcome 😉

clizbe:
This is great! But now that we want the true schema inside the model, we should probably (after merging this) do another update that makes this JSON whatever format we want the user to send to TEM. I guess that's the User Format and then we use pre-processing to make the internal format. Let's discuss on Monday. :)

docs/src/50-schemas.md
```diff
@@ -2,7 +2,7 @@
 
 The optimization model parameters with the input data must follow the schema below for each table. To create these tables we currently use CSV files that follow this same schema and then convert them into tables using TulipaIO, as shown in the basic example of the [Tutorials](@ref basic-example) section.
 
-The schemas can be accessed at any time after loading the package by typing `TulipaEnergyModel.schema_per_table_name` in the Julia console. Here is the complete list of model parameters in the schemas per table (or CSV file):
+The schemas can be found in the `input-schemas.json` or can be accessed at any time after loading the package by typing `TulipaEnergyModel.schema_per_table_name` in the Julia console. Here is the complete list of model parameters in the schemas per table (or CSV file):
```
clizbe, line 5: We should link to the file.

clizbe, line 5: Also, does this command produce a proper output now? When I tried it, it only gave me a statement of "structure with 20 rows" or something lame but not the actual output.

src/input-schemas.jl
```diff
@@ -1,232 +1,44 @@
...
-        # Schema for the flows-rep-periods-partitions.csv file.
-        rep_periods_partition = OrderedDict(
-            :from_asset => "VARCHAR",          # Name of Asset
-            :to_asset => "VARCHAR",            # Name of Asset
-            :year => "INTEGER",
-            :rep_period => "INTEGER",
-            :specification => "VARCHAR",
-            :partition => "VARCHAR",
-        ),
+const schema_per_table_name = OrderedDict(
```
clizbe, line 5: Is there a way of coding this less repetitively? It's okay if this is the best way.

clizbe (changes requested):
Thanks for doing this! I'm happy about this change but have a couple of questions.

src/input-schemas.json
```diff
@@ -0,0 +1,1028 @@
...
+      "default": false,
+      "description": "Whether the unit commitment variables are integer or not.",
+      "type": "BOOLEAN"
+    },
+    "unit_commitment_method": {
+      "UoM": null,
+      "constraints": {
+        "OneOf": [
+          null,
+          "basic"
```
suvayu, line 194: What is the intention here?  Does `null` mean no `unit_commitment`? In that case does TEM still expect this column in the tables?

I think the logically correct way to do this would be something like this:
```
{
  "unit_commitment_integer": {...},
  "unit_commitment_method": {
    ...
    "constraints": {
      "oneOf": ["basic", "something-else"]
    }
  },
  "required": ["unit_commitment_integer"]
}
```
This means, there will always be a `unit_commitment_integer` column, and `unit_commitment_method` will be present only when the previous one is true (which could be checked in code).

I understand if the current structure in TEM does not allow that, in that case this could be a future improvement.

src/input-schemas.json
```diff
@@ -0,0 +1,1028 @@
+{
+  "asset": {
+    "asset": {
+      "UoM": null,
+      "constraints": null,
+      "default": null,
```
suvayu, line 6: Do you think it's possible to drop the keys when the value is `null`?  That would make the schema less verbose.

suvayu (changes requested):
Good first pass!

docs/src/50-schemas.md
```diff
@@ -11,14 +11,56 @@ The schemas can be accessed at any time after loading the package by typing `Tul
...
+using OrderedCollections: OrderedDict
+
+input_schemas = JSON.parsefile("../../src/input-schemas.json"; dicttype = OrderedDict)
+
+let buffer = IOBuffer()
+    for (i,(table_name, fields)) in enumerate(input_schemas)
+        write(buffer, "## Table $i : `$table_name`\n\n")
+        for (field_name, field_info) in fields
+            desc = get(field_info, "description", "No description provided")
+            typ = get(field_info, "type", "Unknown type")
```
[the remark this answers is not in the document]
clizbe: I don't like the look of leading/trailing underscores for any names, but maybe it's something I should get used to if it's a common coding practice?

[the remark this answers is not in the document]
datejada: write the full name (no underscore)

docs/src/50-schemas.md
```diff
@@ -2,7 +2,7 @@
 
 The optimization model parameters with the input data must follow the schema below for each table. To create these tables we currently use CSV files that follow this same schema and then convert them into tables using TulipaIO, as shown in the basic example of the [Tutorials](@ref basic-example) section.
 
-The schemas can be accessed at any time after loading the package by typing `TulipaEnergyModel.schema_per_table_name` in the Julia console. Here is the complete list of model parameters in the schemas per table (or CSV file):
+The schemas can be found in the `input-schemas.json` or can be accessed at any time after loading the package by typing `TulipaEnergyModel.schema_per_table_name` in the Julia console. Here is the complete list of model parameters in the schemas per table (or CSV file):
 
 !!! info "Optional tables/files and their defaults"
```
datejada, line 7: mention that some files use the same schema

src/input-schemas.json
datejada: Thake out the `null` from the JSON and use the get in Julia to handle it. This simplifies the JSON file.

src/input-schemas.jl
abelsiqueira: Something like the following should be possible:

```Julia
const schema_per_table_name = OrderedDict(
    schema_key => OrderedDict(key => value["type"] for (key, value) in schema_content)
    for (schema_key, schema_content) in schema
)
```

src/input-schemas.json
```diff
@@ -0,0 +1,1028 @@
+{
+  "asset": {
+    "asset": {
+      "UoM": null,
```
abelsiqueira, line 4: Is `UoM` a normal JSON schema thing? If not, then the name could be more descriptive.

src/input-schemas.json
```diff
@@ -0,0 +1,1028 @@
...
+      "description": "Capacity of one storage unit.",
+      "type": "DOUBLE"
+    },
+    "consumer_balance_sense": {
+      "UoM": null,
+      "constraints": {
+        "OneOf": [
+          "==",
+          ">=",
+          "=<"
```
abelsiqueira, line 34:
```suggestion
          "<="
```
In Julia, at least.

src/input-schemas.json
```diff
@@ -0,0 +1,1028 @@
...
+      "default": 2000,
+      "description": "Year of investment and operation decisions in the optimization.",
+      "type": "INTEGER"
+    },
+    "min_energy_timeframe_partition": {
+      "UoM": "MWh",
+      "constraints": {
+        "minimum": 0
+      },
+      "deafult": null,
```
abelsiqueira, line 362:
```suggestion
      "default": null,
```

src/input-schemas.json
```diff
@@ -0,0 +1,1028 @@
...
+      "description": "The minimum amount of energy across the timeframe (e.g., a year) that the asset must produce. If the initial value is null, empty, or missing, it will be no limit.",
+      "type": "DOUBLE"
+    },
+    "peak_demand": {
+      "UoM": "MW",
+      "constraints": {
+        "minimum": 0
+      },
+      "deafult": 0,
```
abelsiqueira, line 371:
```suggestion
      "default": 0,
```
I also noticed that there are more, and there could be more before this one, so a replace-all is adequate.

src/input-schemas.json
```diff
@@ -0,0 +1,1028 @@
...
+      "description": "Cost of keeping unit online for one hour or the objective function coefficient on `units_on` variable. e.g., no_load cost or idling cost",
+      "type": "DOUBLE"
+    }
+  },
+  "assets_profiles": {
+    "asset": {
+      "UoM": null,
+      "constraints": null,
```
abelsiqueira, line 397: There might be a way to constrain the value of `asset` to be one of the values from the asset table. I don't actually know if it is possible

src/input-schemas.json
```diff
@@ -0,0 +1,1028 @@
...
+      "default": null,
+      "description": "Name of the asset. Same as the one in the `asset` table.",
+      "type": "VARCHAR"
+    },
+    "partition": {
+      "UoM": null,
+      "constraints": null,
+      "default": 1,
```
abelsiqueira, line 445:
```suggestion
      "default": "1",
```
At least I expect it to have a difference.

src/input-schemas.json
```diff
@@ -0,0 +1,1028 @@
...
+      "UoM": "number",
+      "constraints": null,
+      "default": 0,
+      "description": "Number of existing units in `to_asset` -> `from_asset` direction",
+      "type": "DOUBLE"
+    },
+    "milestone_year": {
+      "UoM": null,
+      "constraints": null,
+      "default": 2000,
```
abelsiqueira, line 625: Should there be a default for `milestone_year`?

abelsiqueira: And `commission_year`, as I see below.

src/input-schemas.json
```diff
@@ -0,0 +1,1028 @@
...
+      "type": "VARCHAR"
+    },
+    "variable_cost": {
+      "UoM": "CUR/MWh",
+      "constraints": null,
+      "default": 0,
+      "description": "Variable cost for the flow.",
+      "type": "DOUBLE"
+    }
+  },
```
abelsiqueira, line 726: Got tired around here. Hope the rest of the file is similar enough to the beginning, and similar comments apply.

abelsiqueira (changes requested):
I made some comments. I don't know much about JSON Schemas, so take it with a grain of salt.
Another comment is that since the JSON schema is a particular type of file, it could benefit from a linter (quick googling gave this option: https://github.com/python-jsonschema/check-jsonschema). @suvayu might have a better recommendation. 

I got a bit confused on how you're planning to use this, but since you're already in discussion with @suvayu on this, I will leave the details to you

src/input-schemas.json
suvayu: I think it stands for "unit of measure"

datejada: correct, it is unit of measure, but, I think being explicit is better, so I will change it 😃

src/input-schemas.json
suvayu: Technically it's possible, but we are not going to use them completely as it should (I'll respond to your earlier "how to use" question separately), so this is something we could do explicitly in code for now.

src/input-schemas.json
suvayu: Why string?

src/input-schemas.json
datejada: ok! I will check if it is possible. Thanks!

src/input-schemas.json
datejada: I prefer to have them, trying to think of users who really don't want to run a multiyear case and the year is meaningless (e.g., the Norse case or any other of the examples we have in the tests—except the multiyear).

suvayu:
> I got a bit confused on how you're planning to use this, but since you're already in discussion with @suvayu on this, I will leave the details to you

So while discussing this during before starting on the change, I realised that doing this "properly" might be difficult because I didn't know any tooling within Julia to do this kind of schema based validation.  So we essentially simplified the schema into a map of mini schemas.  Using it then becomes simpler, because you don't do complex validation anymore, it's just for one column.

E.g. you will see it doesn't define the columns properly, something like this would be the proper way to do it.
```json
{
   "name": string,
   "type": object,
   "properties": [
      {
        "name": "type",
        "type": string,
        "oneOf": ["INTEGER", "DOUBLE", "VARCHAR", "BOOL"]
      },
      {
        "name": "constraints",
        "type": object
      },
      {
         "name": "UoM",
         "type": string
      },
      {
         "name": "default"
      }
    ]
    "required": ["type"]
}
```
But then you need libraries that generate these schemas, and/or that can then validate data using these later (e.g. `pydantic`, or `marshmallow` in Python).

So we simplified our approach to whatever we can use effectively.  So it is now a slightly more easily extendable version of the `OrderedDict` already in `input-schemas.jl`.  It adds a few additional fields, like units, default, description, etc.  And the use in code is unchanged, you get the same `OrderedDict` as before, but in the docs you can have richer more informative lists/tables.

I hope that answers your doubt.

abelsiqueira (approved):
Thanks for the extra info, @suvayu.
LGTM

docs/src/50-schemas.md
datejada: As we discussed during the meeting, this will be added once the the file exists in the main branch.

datejada:
Hi @suvayu and @clizbe, After giving it some more thought, I decided to adopt different schemas for the `assets_profiles` and `assets_timeframe_profiles`. While they share the same columns, their constraints are different. This distinction wasn't relevant before we introduced the JSON file, which is why we previously reused the schema. Now that it is relevant, having separate schemas for each file allows us to establish a direct 1:1 relationship between the table definitions and their schemas. This change will streamline the code, making it cleaner and less repetitive.

Please let me know if you agree or if you have any additional comments or suggestions. 

Thank you!

src/input-schemas.json
```diff
@@ -0,0 +1,820 @@
...
+      "type": "VARCHAR"
+    },
+    "capacity": {
+      "constraints": {
+        "minimum": 0
+      },
+      "default": 0,
+      "description": "Capacity for one unit of the asset (is therefore multiplied by number of existing plus number of invested assets).",
+      "type": "DOUBLE",
+      "unit of measure": "MW"
```
suvayu, line 14: Keys should not have spaces. Use `unit_of_measure`. Otherwise everything looks good.

suvayu (changes requested):

docs/src/50-schemas.md
```diff
@@ -11,14 +11,56 @@ The schemas can be accessed at any time after loading the package by typing `Tul
...
+
+TABLE_NAME
+=========
+
+PARAMETER_NAME
+
+  •  Description: Lorem ipsum
+  •  Type: SQL type of the parameter
+  •  Default: a value or "No default"
+  •  Unit of measure: a value or "-"
```
clizbe, line 25: Can we use "No unit" for consistency or does that cause problems?

clizbe:
@datejada WOW that was a good decision because the new implementation is gorgeous! 😍

suvayu (approved):

clizbe (approved):
Looks GREAT!
Sorry for the slow review.

PR #1055, opened 2025-02-26
Add type assertions in add_expression and attach_expression functions

abelsiqueira:
**Add type assertions in add_expression and attach_expression functions**

This should help make it a bit faster by making all three functions more type stable. All three functions are similarly typed now.


clizbe (approved):
Looks good to me!

PR #1068, opened 2025-03-04
Add simple investment

### LINKED ISSUE #1064: Add simple method for multi-year investments
### Description

Add simple method for multi-year investments

### Sub issues
- [x] Remove some asset data (as if they are moved to simple), and check if the current data (for compact) works. If not, fix the current code. 
~- [ ] Separate compact and simple files, rename current to only compact~
- [x] Define new input files for assets and profiles (e.g., asset-milestone-simple-investment, assets-profiles-simple-investment)
- [x] Add/modify investment and decom variables
- [x] Add capacity outgoing upper limit constraints
- [x] Add/modify objective
- [x] Make sure `none` investment method is included in `simple`
- [x] Add/modify all capacity constraints (including incoming, outgoing for storage binary)
- [x] Add/modify storage constraints
- [x] Do flows
- [x] Do storage energy

### Separate issue
- [x] #1123
...

gnawin:
- Add new files `asset-milestone-simple-investment` and `asset-profiles-simple-investment`.
- Create expressions for simple method:
  - `available_asset_units_simple_investment`
  - `available_energy_units_simple_investment`
  - `available_energy_capacity_simple_investment`
- Change all capacity constraints, including incoming and outgoing. Note now investment method 'none' and 'simple' both use the simple method formulation, so the constraint names have suffix `_simple_investment`
- Modify storage constraints, where storage expressions are used.
- Modify objective function because now compact and simple have to be calculated separately.

## Related issues

Closes #1064 

## Checklist

- [ ] Tests are passing

### DIFF
--- a/src/constraints/ramping-and-unit-commitment.jl
+++ b/src/constraints/ramping-and-unit-commitment.jl
@@ -251,6 +251,7 @@ function _append_ramping_data_to_indices(connection, table_name)
         FROM cons_$table_name AS cons
         LEFT JOIN asset
             ON cons.asset = asset.asset
+        -- Do we need asset_commission below?
         LEFT JOIN asset_commission
             ON cons.asset = asset_commission.asset
             AND cons.year = asset_commission.commission_year
--- a/src/data-preparation.jl
+++ b/src/data-preparation.jl
@@ -60,6 +60,48 @@ blocks, and are replaced by columns `time_block_start` and `time_block_end`.
 
 """
 function create_unrolled_partition_tables!(connection)
+    # Create a temporary table because the functions in this file is created when asset_both must be present
+    # Now asset_both can be missing, but the functions are still used in the code
+    # The required information should be from the union of asset_both and asset_milestone_simple_investment
+    # This is a temporary solution until this function is further refactored
+    DBInterface.execute(
+        connection,
+        "CREATE OR REPLACE TABLE tmp_asset_both AS (
+            SELECT
+                asset_both.asset,
+                asset_both.milestone_year,
+                asset_both.commission_year
+            FROM asset_both
+
+            UNION
+
+            SELECT
+                simple.asset,
+                simple.milestone_year,
+                simple.milestone_year,
+            FROM asset_milestone_simple_investment AS simple
+        )
+        ",
+    )
+
+    # Create a temporary table, same as above
+    DBInterface.execute(
+        connection,
+        "CREATE OR REPLACE TABLE tmp_asset_commission AS (
+            SELECT
+                asset_commission.asset,
+                asset_commission.commission_year
+            FROM asset_commission
+
+            UNION
+
+            SELECT
+                simple.asset,
+                simple.milestone_year
+            FROM asset_milestone_simple_investment AS simple
+        )
+        ",
+    )
     DBInterface.execute(
         connection,
         "CREATE OR REPLACE TABLE t_explicit_assets_rep_periods_partitions AS
@@ -72,9 +114,9 @@ function create_unrolled_partition_tables!(connection)
             rep_periods_data.num_timesteps,
         FROM asset
         CROSS JOIN rep_periods_data
-        LEFT JOIN asset_commission
-            ON asset.asset = asset_commission.asset
-            AND rep_periods_data.year = asset_commission.commission_year
+        LEFT JOIN tmp_asset_commission
+            ON asset.asset = tmp_asset_commission.asset
+            AND rep_periods_data.year = tmp_asset_commission.commission_year
         LEFT JOIN assets_rep_periods_partitions as arpp
             ON asset.asset = arpp.asset
             AND rep_periods_data.year = arpp.year
@@ -83,7 +125,7 @@ function create_unrolled_partition_tables!(connection)
             SELECT
                 asset,
                 milestone_year,
-            FROM asset_both
+            FROM tmp_asset_both
             GROUP BY
                 asset, milestone_year
         ) AS t
@@ -207,15 +249,15 @@ function create_unrolled_partition_tables!(connection)
         "CREATE OR REPLACE TABLE t_explicit_assets_timeframe_partitions AS
         SELECT
             asset.asset,
-            asset_commission.commission_year AS year,
+            tmp_asset_commission.commission_year AS year,
             COALESCE(atp.specification, 'uniform') AS specification,
             COALESCE(atp.partition, '1') AS partition,
         FROM asset AS asset
-        LEFT JOIN asset_commission
-            ON asset.asset = asset_commission.asset
+        LEFT JOIN tmp_asset_commission
+            ON asset.asset = tmp_asset_commission.asset
         LEFT JOIN assets_timeframe_partitions AS atp
             ON asset.asset = atp.asset
-            AND asset_commission.commission_year = atp.year
+            AND tmp_asset_commission.commission_year = atp.year
         WHERE
             asset.is_seasonal
         ",
@@ -399,10 +441,10 @@ function create_lowest_resolution_table!(connection)
         @timeit to "append $table_name rows" for row in DuckDB.query(
             connection,
             "SELECT merged.* FROM $merged_table AS merged
-            LEFT JOIN asset_both
-                ON asset_both.asset = merged.asset
-                AND asset_both.milestone_year = merged.year
-                AND asset_both.commission_year = merged.year
+            LEFT JOIN tmp_asset_both
+                ON tmp_asset_both.asset = merged.asset
+                AND tmp_asset_both.milestone_year = merged.year
+                AND tmp_asset_both.commission_year = merged.year
             ORDER BY
                 merged.asset,
                 merged.year,
@@ -467,10 +509,10 @@ function create_highest_resolution_table!(connection)
                 SELECT DISTINCT asset, year, rep_period, time_block_start
                 FROM $merged_table
             ) AS merged
-            LEFT JOIN asset_both
-                ON asset_both.asset = merged.asset
-                AND asset_both.milestone_year = merged.year
-                AND asset_both.commission_year = merged.year
+            LEFT JOIN tmp_asset_both
+                ON tmp_asset_both.asset = merged.asset
+                AND tmp_asset_both.milestone_year = merged.year
+                AND tmp_asset_both.commission_year = merged.year
             LEFT JOIN rep_periods_data
                 ON merged.year = rep_periods_data.year
                     AND merged.rep_period = rep_periods_data.rep_period


gnawin:
@abelsiqueira I finished the simple investment for one constraint. Could you have a look? The idea is that we iterate and optimize the first constraint, and then the rest will be similar (hopefully). Thank you!

src/expressions/multi-year.jl
```diff
@@ -29,6 +31,36 @@ function create_multi_year_expressions!(connection, model, variables, expression
     #
     # Assumption:
     # - asset_both exists only for (a,my,cy) where technical lifetime was already taken into account
+    #
+    # ------------------------- Simple investmnet -------------------------
+    #
+    # The variable assets_decommission_simple_method is defined for (a, my)
+    # The capacity expression that we need to compute is
+    #
+    #   profile_times_capacity[a, my] = ∑_cy agg(
```
abelsiqueira, line 40: This is not a summation anymore, I think

gnawin: Good catch. It is not anymore.

src/expressions/multi-year.jl
```diff
@@ -179,10 +245,43 @@ function _create_multi_year_expressions_indices!(connection, expressions)
...
+        ARRAY_AGG(DISTINCT var_inv.id) FILTER (
+            WHERE
+                var_inv.milestone_year +
+                    (SELECT
+                        ANY_VALUE(a.technical_lifetime)
+                    FROM asset as a
+                    WHERE a.asset = asset_milestone_simple_investment.asset
+                    )
+              - 1 >= asset_milestone_simple_investment.milestone_year
+            ) AS var_investment_indices,
```
abelsiqueira, line 272: I wanted to test a different approach, but it looks like the query only returns 1 `var_inv` matching this? Is this accurate?

It would be easier to know what to expect before continuing. Can we have a more complicated example to make sure that the query is doing what is expected and to test alternatives?

src/constraints/capacity.jl
```diff
@@ -360,9 +386,49 @@ function _append_capacity_data_to_indices(connection, table_name)
...
+            asset_milestone_simple_investment.investment_limit AS investment_limit,
+        FROM cons_$table_name AS cons
+        LEFT JOIN asset
+            ON cons.asset = asset.asset
+        LEFT JOIN asset_milestone_simple_investment
+            ON cons.asset = asset_milestone_simple_investment.asset
+            AND cons.year = asset_milestone_simple_investment.milestone_year
+        LEFT JOIN expr_available_asset_units_simple_investment AS expr_avail
+            ON cons.asset = expr_avail.asset
+        -- Is it possible to not have milestone_year in this table? It is redundant because year is available in profiles_rep_periods, see the try below
```
abelsiqueira, line 419: I didn't understand this question. The table `expr_available_asset_units_simple_investment` requires `milestone_year`, no? The `cons.year` is also `milestone_year`, right?

src/objective.jl
```diff
@@ -319,6 +350,46 @@ function _create_objective_auxiliary_table(connection, constants)
...
+            COALESCE(
+                lead(asset_milestone_simple_investment.milestone_year) OVER (PARTITION BY asset_milestone_simple_investment.asset ORDER BY asset_milestone_simple_investment.milestone_year) - asset_milestone_simple_investment.milestone_year,
+                1,
+            ) AS years_until_next_milestone_year,
+            operation_discount * years_until_next_milestone_year AS weight_for_operation_discounts,
+        FROM asset_milestone_simple_investment
+        LEFT JOIN asset
+            ON asset.asset = asset_milestone_simple_investment.asset
+        ",
+    )
```
abelsiqueira, line 391: This looks like the only change is that the data from `asset_commission` is already available on `asset_milestone_simple_investment`. It should be correct, if that's the case

abelsiqueira (commented):
@gnawin, thanks for pinging me. I took a first look. I wanted to test the less obvious query to see if it could be improved, but the data is not sufficient to provide an informed decision. 

This query:

```sql
    SELECT 
        asset_simple.asset, 
        asset_simple.milestone_year,
        ARRAY_AGG(var_inv.id) FILTER (
          WHERE
            var_inv.milestone_year +
            (
              SELECT ANY_VALUE(asset.technical_lifetime)
              FROM asset
              WHERE asset.asset = asset_simple.asset
            )
            - 1 >= asset_simple.milestone_year
        ) AS var_inv_indices,
    FROM asset_milestone_simple_investment AS asset_simple 
    LEFT JOIN var_assets_investment AS var_inv 
        ON var_inv.asset = asset_simple.asset 
        AND var_inv.milestone_year <= asset_simple.milestone_year 
    GROUP BY asset_simple.asset, asset_simple.milestone_year
```

returns

```plaintext
 Row │ asset   milestone_year  var_inv_indices
     │ String  Int32           Array…
─────┼──────────────────────────────────────────────────
   1 │ solar             2030  Union{Missing, Int64}[6]
   2 │ ccgt              2030  Union{Missing, Int64}[3]
   3 │ solar             2050  Union{Missing, Int64}[7]
   4 │ ccgt              2050  Union{Missing, Int64}[3]
```

Is this the expected output? If yes, then the problem is that all rows have exactly one matching `var_inv`. So we don't capture any corner cases.
The ideal situation would be having multiple cases here, for instance (if it makes sense), having 0 matches and multiple matches.
Let me know if that makes sense.

Some of the repetition could be generalized, but it would probably still be a bit hard to follow. Let's leave that for the next refactor :wink:

src/constraints/capacity.jl
gnawin: Sorry I put the comment in the wrong line. I meant for the table assets_profiles_simple_investment. Previously there is `commission_year` in the table `asset_profiles`. Now it is gone and the year information is only in the table profiles_rep_periods.  Without having milestone_year explicitly in this new table assets_profiles_simple_investment, I did not find a way to generate the indices correctly 😓

src/expressions/multi-year.jl
gnawin: That is indeed the case for this data. I'll create more corner cases for testing :)

src/objective.jl
gnawin: Yes, it is mostly renaming.

Now all the data from `asset_milestone` and `asset_commission` should be in `asset_milestone_simple_investment`, so I change that. Also since `asset_commission` is no longer needed, that left join has been also removed.

gnawin:
> @gnawin, thanks for pinging me. I took a first look. I wanted to test the less obvious query to see if it could be improved, but the data is not sufficient to provide an informed decision.
> 
> This query:
> 
> ```sql
>     SELECT 
>         asset_simple.asset, 
>         asset_simple.milestone_year,
>         ARRAY_AGG(var_inv.id) FILTER (
>           WHERE
>             var_inv.milestone_year +
>             (
>               SELECT ANY_VALUE(asset.technical_lifetime)
>               FROM asset
>               WHERE asset.asset = asset_simple.asset
>             )
>             - 1 >= asset_simple.milestone_year
>         ) AS var_inv_indices,
>     FROM asset_milestone_simple_investment AS asset_simple 
>     LEFT JOIN var_assets_investment AS var_inv 
>         ON var_inv.asset = asset_simple.asset 
>         AND var_inv.milestone_year <= asset_simple.milestone_year 
>     GROUP BY asset_simple.asset, asset_simple.milestone_year
> ```
> 
> returns
> 
> ```
>  Row │ asset   milestone_year  var_inv_indices
>      │ String  Int32           Array…
> ─────┼──────────────────────────────────────────────────
>    1 │ solar             2030  Union{Missing, Int64}[6]
>    2 │ ccgt              2030  Union{Missing, Int64}[3]
>    3 │ solar             2050  Union{Missing, Int64}[7]
>    4 │ ccgt              2050  Union{Missing, Int64}[3]
> ```
> 
> Is this the expected output? If yes, then the problem is that all rows have exactly one matching `var_inv`. So we don't capture any corner cases. The ideal situation would be having multiple cases here, for instance (if it makes sense), having 0 matches and multiple matches. Let me know if that makes sense.
> 
> Some of the repetition could be generalized, but it would probably still be a bit hard to follow. Let's leave that for the next refactor 😉

Thank you for the check! Yes, it makes sense. I'll create more corner cases and let you know 😄

gnawin:
@abelsiqueira I improved the implementation by the following:

- change `var_asset_investment` (which was overlooked)  because it was only getting information from the old tables. It should get compact investment from the existing tables and the simple investment from the new table. I created them separately and unioned them, to not change the existing structure as much as possible.
- add a few corner cases
  - ccgt is for `none` investment method: both investable and decommissonable are false
  - ocgt is for no investment case: only decommissionable is true
  - solar has two investment variables, but one disappears in the latter milestone year because of technical lifetime; also only use one decommission variable
  - wind has two investment variables and two decommission variables, where technical lifetime does not take effect here.

I checked these cases, and the expressions are correctly generated. Could you have a further look? And possibly improve some queries, since you hinted that before :) Thanks

src/constraints/capacity.jl
gnawin: This line is updated!

src/expressions/multi-year.jl
```diff
@@ -179,10 +245,44 @@ function _create_multi_year_expressions_indices!(connection, expressions)
...
+            ARRAY_AGG(DISTINCT var_dec.id) FILTER (var_dec.id IS NOT NULL) AS var_decommission_indices
+        FROM asset_milestone_simple_investment
+        LEFT JOIN var_assets_decommission_simple_investment AS var_dec
+            ON asset_milestone_simple_investment.asset = var_dec.asset
+            AND asset_milestone_simple_investment.milestone_year >= var_dec.milestone_year
+        LEFT JOIN var_assets_investment AS var_inv
+            ON asset_milestone_simple_investment.asset = var_inv.asset
+            AND asset_milestone_simple_investment.milestone_year >= var_inv.milestone_year
+        GROUP BY asset_milestone_simple_investment.asset, asset_milestone_simple_investment.milestone_year
```
abelsiqueira, line 282: I think this can be replaced with something like

```sql
    SELECT
        asset_simple.asset,
        asset_simple.milestone_year,
        ANY_VALUE(asset_simple.initial_units) AS initial_units,
        -- DISTINCT is required because joining both inv and dec at the same time create a mini-cross join between them.
        ARRAY_AGG(DISTINCT var_inv.id) FILTER (var_inv.id IS NOT NULL) AS var_inv_indices,
        ARRAY_AGG(DISTINCT var_dec.id) FILTER (var_dec.id IS NOT NULL) AS var_dec_indices,
    FROM asset_milestone_simple_investment AS asset_simple
    LEFT JOIN asset
        ON asset.asset = asset_simple.asset
    LEFT JOIN var_assets_investment AS var_inv
        ON var_inv.asset = asset_simple.asset
        AND var_inv.milestone_year <= asset_simple.milestone_year
        AND var_inv.milestone_year + asset.technical_lifetime - 1 >= asset_simple.milestone_year
    LEFT JOIN var_assets_decommission_simple_investment AS var_dec
        ON var_dec.asset = asset_simple.asset
        AND var_dec.milestone_year <= asset_simple.milestone_year
    GROUP BY asset_simple.asset, asset_simple.milestone_year
    ORDER BY asset_simple.asset, asset_simple.milestone_year
```

I.e., the filtering is done when joining. This might also apply to the "compact" version of this table.
It has less complicated parts, which is was I suggest it.

The "DISTINCT" in the aggregation of the ids is necessary because, if I got it, the var_inv and var_dec joins end up creating a mini-cross join for a given key. I.e., since `var_inv` has entries `(id=6, wind, 2030)` and `(id=4, wind, 2050)` and `var_dec` has entries `(id = 3, wind, 2050)` and `(id = 5, wind, 2030)`, when we join with `year <= 2050`, then we create the 4 combinations. I added a simplified version of this comment in my suggestion.

abelsiqueira (changes requested):
Thanks for the update, @gnawin. I've made two code suggestions in the places you changed, I hope they're clear enough.
Given that some of these files won't exist, it might be useful to add them to `src/io.jl` function `create_internal_tables!`. Depending on the code, you might now even need to change the data of the other test inputs.

src/sql/create-variables.sql
```diff
@@ -182,6 +182,60 @@ drop sequence id
...
+    from var_assets_simple_investment
+)
+select
+    nextval('id') as id,
+    asset,
+    milestone_year,
+    investment_integer,
+    capacity,
+    investment_limit
+from unioned;
```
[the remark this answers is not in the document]
gnawin: Thanks! Applied.

src/expressions/multi-year.jl
gnawin: Thanks! Applied.

gnawin:
@abelsiqueira I finished most of the bullet points from #1064, only the separate naming of the compact and simple method has not been done. Before I fix all the tests, I'd like to check with you the following:

- Simple method now always has a suffix, and compact hopefully also but I want to make sure that `asset_both`, `asset_milestone` and `asset_commission` are not used for general purposes. See the change in `data-preparation.jl` and a question in `ramping-and-unit-commitment.jl`. 

- Flow files and names have not been changed, because flow can be both a flow and a transport, separating them will not only change the investment, but also break the flow variables, unless there is a not too complex way to do it? 

- I now have a new file `asset-profiles-simple-investment` that has a column `milestone_year`, which is easy but actually not needed, because this information is available in `rep-periods-data`. Ideally we would check the matching profiles names and (milestone) year in `rep-periods-data` and match the profiles names and asset names in `asset-profiles-simple-investment`, so `milestone_year` is no longer needed.

I also tag you in the corresponding places in the code. Could you have a look before I go and fix all the tests? Thank you!

src/constraints/capacity.jl
```diff
@@ -176,19 +207,19 @@ function add_capacity_constraints!(connection, model, expressions, constraints,
...
         attach_expression!(
             cons,
-            :profile_times_capacity_with_investment_variable,
+            :profile_times_capacity_with_investment_variable_simple_investment,
             [
                 begin
                     availability_agg = _profile_aggregate(
                         profiles.rep_period,
-                        (row.profile_name, row.year, row.rep_period),
+                        (row.avail_profile_name, row.year, row.rep_period),
```
gnawin, line 222: @abelsiqueira here I changed the name and also below get rid of the sum, does it make sense to you?

src/constraints/capacity.jl
```diff
@@ -360,9 +394,49 @@ function _append_capacity_data_to_indices(connection, table_name)
...
+            ON cons.asset = asset.asset
+        LEFT JOIN asset_milestone_simple_investment
+            ON cons.asset = asset_milestone_simple_investment.asset
+            AND cons.year = asset_milestone_simple_investment.milestone_year
+        LEFT JOIN expr_available_asset_units_simple_investment AS expr_avail
+            ON cons.asset = expr_avail.asset
+            AND cons.year = expr_avail.milestone_year
+        LEFT JOIN assets_profiles_simple_investment AS avail_profile
+            ON cons.asset = avail_profile.asset
+        -- Is it possible to not have milestone_year in this table? It is redundant because year is available in profiles_rep_periods, see the try below
```
gnawin, line 430: @abelsiqueira

src/constraints/ramping-and-unit-commitment.jl
```diff
@@ -251,6 +251,7 @@ function _append_ramping_data_to_indices(connection, table_name)
         FROM cons_$table_name AS cons
         LEFT JOIN asset
             ON cons.asset = asset.asset
+        -- Do we need asset_commission below?
```
gnawin, line 254: @abelsiqueira

src/variables/investments.jl
```diff
@@ -53,6 +53,13 @@ function add_investment_variables!(model, variables)
...
         ),
+        (
+            :assets_decommission_simple_investment,
+            row -> (row.asset, row.milestone_year),
+            _ -> 0.0,
+            _ -> Inf,
+            row -> row.investment_integer,
+        ),
         (
             :flows_decommission,
```
gnawin, line 64: @abelsiqueira ideally `flows_decomission` should not have `commission_year`, but as explained in the main post, it seems not straightforward to change the file/table structure of the flows. So `commission_year` is preserved. The idea is that we should make sure in the data that we do not create extra decommission variables. I mean ideally, the all flow part should be rewritten for `simple`, but that might be too much.

src/constraints/capacity.jl
abelsiqueira: Answering your third bullet point here. Might apply to other places.
From what I understand, the `profiles` have a `year` for index, and the `asset_profiles` have another year (one is commission and one is milestone). So if you don't need one in `asset_profiles_simple`, it should be fine because you still have another one to match with `profile`. Whether these years match is another issue, that I can't answer.

abelsiqueira: You say below doesn't work. What do you mean?

src/constraints/capacity.jl
abelsiqueira: How did you get rid of the sum? There is no aggregation anymore?

src/constraints/ramping-and-unit-commitment.jl
abelsiqueira: I created #1117 to look into this

src/data-preparation.jl
```diff
@@ -60,6 +60,48 @@ blocks, and are replaced by columns `time_block_start` and `time_block_end`.
 
 """
 function create_unrolled_partition_tables!(connection)
+    # Create a temporary table because the functions in this file is created when asset_both must be present
```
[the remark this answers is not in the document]
abelsiqueira: I created #1117 to look into this

src/io.jl
```diff
@@ -9,6 +9,10 @@ function create_internal_tables!(connection; skip_validation = false)
 
     # Create tables that are allowed to be missing
     tables_allowed_to_be_missing = [
+        "asset_milestone_simple_investment"
```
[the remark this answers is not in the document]
abelsiqueira: I'm surprised that the tables below can be missing.

src/variables/investments.jl
abelsiqueira: I would definitely not touch it now, because this PR is already too big.

abelsiqueira (changes requested):
@gnawin, thanks for the update.

As I mentioned in our conversation yesterday. Any place that has `year` is a mystery to me. Review my comments with that in mind.

I created #1117 to look into the necessity of the `asset_both` and `asset_milestone` in that part of the code.

The fact that we need a new table with the union of some compact stuff with some simple stuff in more than one place, makes me wonder if we should have an official new table with this union.

These data issues related to year (year = commission or milestone; profiles have one 'year' index on its table and another 'year' in another table) are starting to worry me. I hope we don't need another data refactor, but if we do, it's better to do it before everything is (more) entangled again. I mentioned in the past that "asset_both" is not a good name because it doesn't have any meaning. Now we know that its meaning is related to investment (and decommission) for compact assets. If these insights are present when deciding the data format, it makes it easier to reason joins and such, for instance, whether you need a `milestone_year` for `assets_profiles_simple`.

I haven't reviewed everything again, but I think I answered everything you asked. Ping me again, otherwise.

gnawin:
Closed in favor of #1122

PR #1080, opened 2025-03-10
Test changes to MPS on CI

### LINKED ISSUE #1076: Verify in the CI that the MPS files have not changed
### Description

The scenario tests only verify that the solution after solving remains the same - and not even that for the EU case.
One more robust verification is that the MPS file remains the same (initially we considered the LP file, but it is not a nice file).

### Validation and testing

_No response_

### Motivation

_No response_

### Target audience

_No response_

### Can you help?

_No response_

abelsiqueira:
Stores the current version of the MPS files of the tests and create a workflow that compares them in a PR.

- Create a `model-mps-common.jl` and `model-mps-update.jl` files to update the MPS files.
- Create a `model-mps-compare.jl` to generate new MPS files and compare the existing MPS files to the new ones. 
- Create a workflow `CompareMPS.yml` to run the comparison script and, if the files are different, create a GitHub comment on the PR with the contents of the log. 

## Related issues

Closes #1076 

## Checklist

- [x] Tests are passing


abelsiqueira:
@odow @joaquimg, tagging you in case you have any clues to what could cause this:

- I run `julia --project=. utils/scripts/update-model-mps.jl` to generate the .mps of all tests
- In the tests, I run the same commands as the script to generate the .mps
- I sort and compare them and there is a difference due to numerical errors:
  ```
  ❯ diff <(sort test/tmpdir/Norse.mps) <(sort benchmark/model-mps-folder/Norse.mps )
  40466c40466
  <     rhs       max_energy_over_clustered_year[Midgard_Wind,2030,113:140] 1.1339999999999996e8
  ---
  >     rhs       max_energy_over_clustered_year[Midgard_Wind,2030,113:140] 113400000
  ```

Any clue to what could cause it?

odow:
Hey, sorry I missed this ping.

If you're building the problems on separate computers, I'm not sure there's much we can do here. 

I would just advise a more clever comparison of the files than `==`. 

Maybe something like this:
```julia
function test_mps_equality(file_A, file_B)
    function compare(a::AbstractString, b::AbstractString)
        a_float = tryparse(Float64, a)
        b_float = tryparse(Float64, b)
        @test typeof(a_float) == typeof(b_float)
        if a_float === b_float === nothing
            @test a == b  # Must be a ::String field
        else
            @test a_float isa Float64 && b_float isa Float64
            # If numbers, allow some imprecision between the files due to
            # roundoff
            @test isapprox(a, b; atol = 1e-12)
        end
        return
    end
    function compare(a::AbstractVector, b::AbstractVector)
        @test length(a) == length(b)
        @test all(compare(ai, bi) for (ai, bi) in zip(a, b))
        return
    end
    A = readlines(file_A)
    B = readlines(file_B)
    @test length(A) == length(B)
    for (a, b) in zip(A, B)
        compare(split(a), split(b))
    end
    return
end
```

abelsiqueira:
Thanks, @odow. I was hoping we wouldn't have to parse the MPS files, but it seems to be the simplest way.

> If you're building the problems on separate computers, I'm not sure there's much we can do here.

It was actually the same computer, but different environments, and the code was executed in different pipelines (copy-pasted from a script to runtests.jl). But naturally, the test will run on GitHub's CI, so different "computers" will also be a problem 🥲

datejada (commented):
Thanks! @abelsiqueira, I love this PR 😍 it will help to detect changes in the model constraints so much easier 👓 

So, I left a suggestion below, and I want you to add a section in the developers section of the docs describing the files used in this CI and how to run locally, if you want to test before pushing it to GitHub. That is helpful information to have there.

datejada:
BTW: I believe this CI can be generally applied to other optimization models in Julia (even maybe in different languages). We should boast it on social media (e.g., LinkedIn and the JuMP Forum). What do you think?

abelsiqueira:
@datejada, I've updated with your suggestion and with the developer documentation changes. Let me know what you think

gnawin:
@abelsiqueira Thanks for the explanation! Nah, the PR is fine to me. I know it is even trickier to do the lp file. Since you chatted with @datejada, I can consult him later how to read this property. @datejada, could you explain to me here how to read the MPS file? Or let's check during the next Tulipa day.

datejada:
@gnawin I know what you mean, LP files are more human readable, but as Gurobi describes in link below _“LP files do not preserve column order when read, and they typically don’t preserve the exact numerical values of the coefficients”_ So, that means that for comparing them in CI is more challenging, and might give false negative warnings when the files are generated on different machines.

On the other hand, MPS files have a standard format and schema that makes them more hard to read, but, in the CI will be certain that the change is a real difference in the model.

So, once you get used to MPS file format (see also link below from Gurobi), one can tell where the difference is. But, if you don't want to deal with the MPS files, the CI will tell you the difference and then you can create the .lp files and compare them.

https://docs.gurobi.com/projects/optimizer/en/current/reference/fileformats/modelformats.html

We can try to extend this workflow to LP files, but I am unsure how many false positives will appear when running on different machines. Maybe we can leave it as an option or a convenience function to run it locally.

Anyway, I hope this extra information helps.

abelsiqueira:
In this case, I recommend going ahead with this PR and evaluating the workflow for a while with the following in mind:

When it fails,
- Is this check useful?
- Is this log useful?
- What is missing here that would be available in the LP?

When it works,
- Is it taking too long to pass?

And, naturally, if it has false positives or negatives then a bug report should be created.

datejada (approved):
@abelsiqueira Thanks for the changes, the new section in the dev docs is clear and will help a lot. Thanks!

datejada:
> In this case, I recommend going ahead with this PR and evaluating the workflow for a while with the following in mind:
> 
> When it fails,
> 
> * Is this check useful?
> * Is this log useful?
> * What is missing here that would be available in the LP?
> 
> When it works,
> 
> * Is it taking too long to pass?
> 
> And, naturally, if it has false positives or negatives then a bug report should be created.

I agree with the proposed plan. Thanks. I belive it is useful even as a MPS file, but let's see how it goes when use it 😃

PR #1085, opened 2025-03-11
Use absolute path to include json file

abelsiqueira:
Using `@__DIR__` to define a path is more robust.
Fixes a bug importing Tulipa


datejada (approved):
Thanks! This should solve the issue in the Julia register

PR #1088, opened 2025-03-11
Implement data validation step with basic check for duplicate rows

### LINKED ISSUE #1091: No duplicate rows in input data

abelsiqueira:
Create a data validation function to call various validation functions.
Call this data validation after initialising empty tables and before
creating the internal tables inside the create_internal_tables function.

## Related issues

Starting point for #461
Closes #1091 

## Checklist

- [x] Tests are passing

### DIFF
--- a/src/data-validation.jl
+++ b/src/data-validation.jl
@@ -1,0 +1,80 @@
+export DataValidationException
+
+"""
+    DataValidationException
+
+Exception related to data validation of the Tulipa Energy Model input data.
+"""
+mutable struct DataValidationException <: Exception
+    error_messages::Vector{String}
+end
+
+function Base.showerror(io::IO, ex::DataValidationException)
+    println(io, "DataValidationException: The following issues were found in the data:")
+    for error_message in ex.error_messages
+        println(io, "- " * error_message)
+    end
+end
+
+"""
+    validate_data!(connection)
+
+Raises an error if the data is not valid.
+"""
+function validate_data!(connection)
+    error_messages = String[]
+    @timeit to "no duplicate rows" append!(error_messages, _validate_no_duplicate_rows!(connection))
+
+    if length(error_messages) > 0
+        throw(DataValidationException(error_messages))
+    end
+
+    return
+end
+
+function _validate_no_duplicate_rows!(connection)
+    # It should be possible to add a primary key to the tables below to avoid this validation.
+    # However, where to add this, and how to ensure it was added is not clear.
+    duplicates = String[]
+    for (table, primary_keys) in (
+        ("asset", (:asset,)),
+        ("asset_both", (:asset, :milestone_year, :commission_year)),
+        ("asset_commission", (:asset, :commission_year)),
+        ("asset_milestone", (:asset, :milestone_year)),
+        ("assets_profiles", (:asset, :commission_year, :profile_type)),
+        ("assets_rep_periods_partitions", (:asset, :year, :rep_period)),
+        ("assets_timeframe_partitions", (:asset, :year)),
+        ("assets_timeframe_profiles", (:asset, :commission_year, :profile_type)),
+        ("flow", (:from_asset, :to_asset)),
+        ("flow_both", (:from_asset, :to_asset, :milestone_year, :commission_year)),
+        ("flow_commission", (:from_asset, :to_asset, :commission_year)),
+        ("flow_milestone", (:from_asset, :to_asset, :milestone_year)),
+        ("flows_profiles", (:from_asset, :to_asset, :year, :profile_type)),
+        ("flows_rep_periods_partitions", (:from_asset, :to_asset, :year, :rep_period)),
+        ("group_asset", (:name, :milestone_year)),
+        ("profiles_rep_periods", (:profile_name, :year, :rep_period, :timestep)),
+        ("profiles_timeframe", (:profile_name, :year, :period)),
+        ("rep_periods_data", (:year, :rep_period)),
+        ("rep_periods_mapping", (:year, :period, :rep_period)),
+        ("timeframe_data", (:year, :period)),
+        ("year_data", (:year,)),
+    )
+        append!(duplicates, _validate_no_duplicate_rows!(connection, table, primary_keys))
+    end
+
+    return duplicates
+end
+
+function _validate_no_duplicate_rows!(connection, table, primary_keys)
+    keys_as_string = join(primary_keys, ", ")
+    duplicates = String[]
+    for row in DuckDB.query(
+        connection,
+        "SELECT $keys_as_string, COUNT(*) FROM $table GROUP BY $keys_as_string HAVING COUNT(*) > 1",
+    )
+        values = join(["$k=$(row[k])" for k in primary_keys], ", ")
+        push!(duplicates, "Table $table has duplicate entries for ($values)")
+    end
+
+    return duplicates
+end
--- a/src/io.jl
+++ b/src/io.jl
@@ -5,7 +5,7 @@ export create_internal_tables!, export_solution_to_csv_files
 
 Creates internal tables.
 """
-function create_internal_tables!(connection)
+function create_internal_tables!(connection, skip_validation = false)
 
     # Create tables that are allowed to be missing
     tables_allowed_to_be_missing = [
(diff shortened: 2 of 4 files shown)


abelsiqueira:
@datejada, for the failing data you shared Monday, this is the result:

```Julia
ERROR: DataValidationException: The following issues were found in the data:
- Table asset_commission has duplicate entries for (asset=DE_Onshore_Wind, commission_year=2050)
- Table asset_commission has duplicate entries for (asset=NL_Onshore_Wind, commission_year=2050)
- Table asset_commission has duplicate entries for (asset=DE_Offshore_Wind, commission_year=2050)
- Table asset_commission has duplicate entries for (asset=NL_Solar, commission_year=2050)
- Table asset_commission has duplicate entries for (asset=BE_Onshore_Wind, commission_year=2050)
- Table asset_commission has duplicate entries for (asset=BE_Offshore_Wind, commission_year=2050)
- Table asset_commission has duplicate entries for (asset=BE_Solar, commission_year=2050)
- Table asset_commission has duplicate entries for (asset=NL_Offshore_Wind, commission_year=2050)
- Table asset_commission has duplicate entries for (asset=DE_Solar, commission_year=2050)
- Table asset_milestone has duplicate entries for (asset=DE_Onshore_Wind, milestone_year=2050)
- Table asset_milestone has duplicate entries for (asset=NL_Onshore_Wind, milestone_year=2050)
- Table asset_milestone has duplicate entries for (asset=DE_Offshore_Wind, milestone_year=2050)
- Table asset_milestone has duplicate entries for (asset=NL_Solar, milestone_year=2050)
- Table asset_milestone has duplicate entries for (asset=BE_Onshore_Wind, milestone_year=2050)
- Table asset_milestone has duplicate entries for (asset=BE_Offshore_Wind, milestone_year=2050)
- Table asset_milestone has duplicate entries for (asset=BE_Solar, milestone_year=2050)
- Table asset_milestone has duplicate entries for (asset=NL_Offshore_Wind, milestone_year=2050)
- Table asset_milestone has duplicate entries for (asset=DE_Solar, milestone_year=2050)
- Table assets_rep_periods_partitions has duplicate entries for (asset=BE_Onshore_Wind, year=2050, rep_period=1)
- Table assets_rep_periods_partitions has duplicate entries for (asset=NL_Solar, year=2050, rep_period=1)
- Table assets_rep_periods_partitions has duplicate entries for (asset=DE_Onshore_Wind, year=2050, rep_period=1)
- Table assets_rep_periods_partitions has duplicate entries for (asset=NL_Onshore_Wind, year=2050, rep_period=1)
- Table assets_rep_periods_partitions has duplicate entries for (asset=BE_Offshore_Wind, year=2050, rep_period=1)
- Table assets_rep_periods_partitions has duplicate entries for (asset=DE_Offshore_Wind, year=2050, rep_period=1)
- Table assets_rep_periods_partitions has duplicate entries for (asset=NL_Offshore_Wind, year=2050, rep_period=1)
- Table assets_rep_periods_partitions has duplicate entries for (asset=BE_Solar, year=2050, rep_period=1)
- Table assets_rep_periods_partitions has duplicate entries for (asset=DE_Solar, year=2050, rep_period=1)
```

src/data-validation.jl
```diff
@@ -0,0 +1,80 @@
...
+function validate_data!(connection)
+    error_messages = String[]
+    @timeit to "no duplicate rows" append!(error_messages, _validate_no_duplicate_rows!(connection))
+
+    if length(error_messages) > 0
+        throw(DataValidationException(error_messages))
+    end
+
+    return
+end
```
datejada, line 33: So, for my understanding, in this function, we can start adding more validation functions, correct?
And second, any validation function must/(should?) return a text (i.e., String) type with a meaningful message to the user, correct?

src/data-validation.jl
```diff
@@ -0,0 +1,80 @@
...
+
+function _validate_no_duplicate_rows!(connection)
+    # It should be possible to add a primary key to the tables below to avoid this validation.
+    # However, where to add this, and how to ensure it was added is not clear.
```
datejada, line 37: The other day, I wondered about primary keys (PK) in our tables in DuckDB. The bare minimum data that the user should provide for each table is PKs. Any other value can be filled in with the defaults, but PKs are necessary.

Anyway, it is more a thought to discuss later, but I like that the implementation here starts giving the information of which are the PK in our tables.

src/data-validation.jl
abelsiqueira: 1. Yes
2. In this implementation, yes. There are other ways to handle this. For instance, we can fail faster by making every function throw an exception directly, but then if you have multiple issues, you would need to run it multiple times. So I prefer aggregating the error messages. Also, there might be some tweaks, but let's have some validation functions first then think about them.

src/io.jl
```diff
@@ -5,7 +5,7 @@ export create_internal_tables!, export_solution_to_csv_files
 
 Creates internal tables.
 """
-function create_internal_tables!(connection)
+function create_internal_tables!(connection, skip_validation = false)
```
datejada, line 8: Should it be a keyword? I mean something like this:

```suggestion
function create_internal_tables!(connection; skip_validation = false)
```

src/data-validation.jl
abelsiqueira: The pipeline could be creating the empty SQL table with PKs (and foreign keys) and loading the CSV into them. But I'm not sure how robust that would be for different sources.
An alternative could be using `ALTER TABLE` to add the PKs and FKs, but I don't know how to check that they already exist, and where to do that alteration.

src/data-validation.jl
```diff
@@ -0,0 +1,80 @@
...
+
+    return duplicates
+end
+
+function _validate_no_duplicate_rows!(connection, table, primary_keys)
+    keys_as_string = join(primary_keys, ", ")
+    duplicates = String[]
+    for row in DuckDB.query(
+        connection,
+        "SELECT $keys_as_string, COUNT(*) FROM $table GROUP BY $keys_as_string HAVING COUNT(*) > 1",
```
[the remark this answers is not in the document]
abelsiqueira: I actually didn't know anything about `COUNT(1)` (where did you get it?). After searching a bit, it looks like that is **not** the case, at least in DuckDB. They should return the same thing, as far as I see. `COUNT(*)` will count how many rows, and `COUNT(1)` will count 1 however many times there are rows. The difference would be `COUNT(col)` where `col` is some column. If some values of `col` are `null`, these are not counted.

MWE:

```sql
create table mydb (id integer, value real);
insert into mydb values (1, 1.0), (2, null), (null, 3.0), (null, null);
select count(*), count(1), count(id) from mydb;
```

Output:

```text
┌──────────────┬──────────┬───────────┐
│ count_star() │ count(1) │ count(id) │
│    int64     │  int64   │   int64   │
├──────────────┼──────────┼───────────┤
│            4 │        4 │         2 │
└──────────────┴──────────┴───────────┘
```

The table has a row with `null` only, and still the `count(*)` and `count(1)` return the same counts.

src/io.jl
abelsiqueira: Yes, good catch

test/test-data-validation.jl
```diff
@@ -0,0 +1,40 @@
+const TEM = TulipaEnergyModel
+
+@testset "Test DataValidationException print" begin
```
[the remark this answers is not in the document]
abelsiqueira: Me too. I hope this remains possible for all validation steps.

abelsiqueira:
Thanks for the review. I think I replied to all, let me know otherwise.

src/data-validation.jl
datejada: Ok! It is not a urgent thing, so we can just start thinking about it. Thanks!

src/data-validation.jl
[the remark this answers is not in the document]
datejada: Awesome example! Thanks, I read about the COUNT(1) in a Medium article: https://medium.com/@vijaygadhave2014/when-to-use-count-vs-count-1-in-sql-queries-5d049529355c

So, no changes in the code here, it was more for learning different options when using SQL. Thanks!

datejada (approved):
Great! Thanks for all the replies to my comments.

PR #1089, opened 2025-03-11
Create connection inside benchmark run because tables exist

abelsiqueira:
The benchmark run is failing due to `var_flow` existing.
This happened after we changed to SQL in files, I believe, because we change
from "create or replace" to "create".

## Related issues

Bugfix

## Checklist

- [x] Tests are passing

### DIFF
--- a/benchmark/benchmarks.jl
+++ b/benchmark/benchmarks.jl
@@ -15,17 +15,23 @@ end
 
 const OUTPUT_FOLDER_BM = mktempdir()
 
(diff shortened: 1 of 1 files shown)


abelsiqueira:
This also makes me think that `create_model!` might not be properly evaluating the whole thing, for instance, not recreating the expression tables.

datejada (approved):
Thanks! Independently of the bug, It is better to benchmark it as in this PR. LGTM.

PR #1092, opened 2025-03-12
Add setup steps to benchmark to properly run from scratch

abelsiqueira:
@datejada, similar to #1089.
Probably, if we don't benchmark from scratch, then we are skipping some table creation.

## Related issues

None.

## Checklist

- [x] I am following the [contributing guidelines](https://github.com/TulipaEnergy/TulipaEnergyModel.jl/blob/main/docs/src/90-contributing.md)
- [x] Tests are passing
- [x] Lint workflow is passing
- [x] Docs were updated and workflow is passing

PR #1093, opened 2025-03-12
Add instructions to run comparative benchmarks locally

abelsiqueira:
- **Add instructions to run comparative benchmarks locally**

Blocked until #1092 is merged 

## Related issues

There is no related issue.

## Checklist

- [x] Tests are passing


docs/src/91-developer.md
```diff
@@ -424,6 +424,65 @@ If you want to manually run the benchmarks, you can do the following:
   results = run(SUITE, verbose=true)
   ```
 
+### Manually running the benchmark across versions
+
+We use the package [AirspeedVelocity.jl](https://github.com/MilesCranmer/AirspeedVelocity.jl) to run the benchmarks in the CI, but it can also be used to compare explicitly named versions manually.
+
+1. Run the following to install AirspeedVelocity's commands to your Julia `bin` folder (`~/.julia/bin` on MacOS and Linux)
```
datejada, line 431: It took me a bit to find it in Windows, so adding a comment for Windows' users

```suggestion
1. Run the following to install AirspeedVelocity's commands to your Julia `bin` folder (`~/.julia/bin` on MacOS and Linux). On Windows, if you are using the default Julia installation, search for `C:/Users/` then the folder of your Windows user and then `.julia/bin`
```

docs/src/91-developer.md
```diff
@@ -424,6 +424,65 @@ If you want to manually run the benchmarks, you can do the following:
...
+
+   ```bash
+   julia -e 'using Pkg; Pkg.activate(temp=true); Pkg.add("AirspeedVelocity")'
```
datejada, line 434: More for curiosity: Why do we need the `temp=true`?

docs/src/91-developer.md
```diff
@@ -424,6 +424,65 @@ If you want to manually run the benchmarks, you can do the following:
...
+   ```
+
+1. Check that `benchpkg` was installed:
+
+   ```bash
+   benchpkg --version
+   ```
+
+   If if can't be found, then it is possible that your Julia `bin` folder is not in the `PATH`. After fixing this, try again.
```
datejada, line 443: Nice! it acctually happened to me 😝

docs/src/91-developer.md
```diff
@@ -424,6 +424,65 @@ If you want to manually run the benchmarks, you can do the following:
...
+
+1. Then, run `benchpkg` with `--rev` to list the versions to be tested and `--bench-on` to indicate with script to use (if necessary). For instance:
+
+   ```bash
+   benchpkg TulipaEnergyModel --rev=v0.12.0,main --bench-on=main
+   ```
+
+   It is possible to use `dirty` as a `rev` value, to get the local modification. Alternatively, list tags or branches to compare.
```
datejada, line 451: I got the following error when  using `dirty`:
```bash
Error: You must specify a `path` when using the `dirty` revision.
```
So, actually, you need to type it like this to work:

```bash
benchpkg TulipaEnergyModel --path="C:\github-repos\TulipaEnergyModel.jl" --rev=dirty,main --bench-on=main
```

Then it worked. So, Maybe we can add something like this in the text:

```suggestion
   It is possible to use `dirty` as a `rev` value, to get the local modification. If So, you must also pass the `path` to your local package files. For instance:

```bash
benchpkg TulipaEnergyModel --path="PATH TO YOUR LOCAL REPO" --rev=dirty,main --bench-on=main
```

Alternatively, list tags or branches to compare.
```

File: docs/src/91-developer.md [hunk 5]
```diff
@@ -424,6 +424,65 @@ If you want to manually run the benchmarks, you can do the following:
...
+   benchpkg --version
+   ```
+
+   If if can't be found, then it is possible that your Julia `bin` folder is not in the `PATH`. After fixing this, try again.
+
+1. Then, run `benchpkg` with `--rev` to list the versions to be tested and `--bench-on` to indicate with script to use (if necessary). For instance:
+
+   ```bash
+   benchpkg TulipaEnergyModel --rev=v0.12.0,main --bench-on=main
+   ```
```
datejada, line 449: I tested what happens with older versions, and as in the CI benchmarks, the versions must have same internal structure(?) and the benchmark files must also be the same (?)...

Anyway, my point is that I tried `--rev=v0.10.4` and it failed. As expected. So, maybe me should add a note or warning that versions you are comparing must be compatible for benchmarking otherwise the benchmark will fail.

I am unsure how to write to clarify it, so please add it as you think it is more understandable.

docs/src/91-developer.md
abelsiqueira: We don't need to explicitly install `AirspeedVelocity` anywhere, so we create a temp env to not pollute any project or global env.

docs/src/91-developer.md
abelsiqueira: I hope it was easy to fix, because on Windows I have no idea

docs/src/91-developer.md
abelsiqueira: True, let me write something

docs/src/91-developer.md
abelsiqueira: True, let me write something a bit different. An easier way is to just move to the TulipaEnergyModel repo and remove the `TulipaEnergyModel` argument, which is what I was trying before.

abelsiqueira:
Thanks for the review! I've updated with your suggestions.

docs/src/91-developer.md
datejada: Yes, I am used to it as a native Windows user 😛, but there is plenty of information about how to add a folder in the `PATH`

datejada (approved):
Thanks for the changes! LGTM ;)

PR #1099, opened 2025-03-17
Validate schema's oneOf constraints

### LINKED ISSUE #1098: Check that all assets have a type associated (i.e., consumer, producer, conversion, storage, hub)

abelsiqueira:
Add a validation step to check the schema's oneOf constraints.
I.e., loop over each table and their columns and find the ones
that have constraints of the type oneOf, and check if the
corresponding table in the connection have the correct specification.

Closes #1098

### DIFF
--- a/src/data-validation.jl
+++ b/src/data-validation.jl
@@ -24,6 +24,10 @@ Raises an error if the data is not valid.
 function validate_data!(connection)
     error_messages = String[]
     @timeit to "no duplicate rows" append!(error_messages, _validate_no_duplicate_rows!(connection))
+    @timeit to "valid schema's oneOf constraints" append!(
+        error_messages,
+        _validate_schema_one_of_constraints!(connection),
+    )
 
     if length(error_messages) > 0
         throw(DataValidationException(error_messages))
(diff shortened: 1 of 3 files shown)


src/data-validation.jl
```diff
@@ -78,3 +82,24 @@ function _validate_no_duplicate_rows!(connection, table, primary_keys)
 
     return duplicates
 end
+
+function _validate_schema_one_of_constraints!(connection)
+    error_messages = String[]
+    for (table_name, table) in TulipaEnergyModel.schema, (col, attr) in table
```
datejada, line 88: By looking at this part `TulipaEnergyModel.schema`, I wonder if it is then better to be explicit in the constant definition in this part of the code:

https://github.com/TulipaEnergy/TulipaEnergyModel.jl/blob/d1cf45f0c5379a95e6320eabdc84810ad8c9d6b6/src/input-schemas.jl#L3

And then say something like:

```julia
const schema = JSON.parsefile(joinpath(@__DIR__, "input-schemas.json"); dicttype = OrderedDict);
```

abelsiqueira: That's a good idea. I've updated now

abelsiqueira:
Thanks for the review. I've updated it now.

datejada (approved):
Awesome! Thanks for the changes

PR #1102, opened 2025-03-17
Validate that only transport flows can be investable

### LINKED ISSUE #1101: Check that the flows that are defined as investable are also defined as transport

abelsiqueira:
Check that only transport flows are investable 

## Related issues

Closes #1101

## Checklist

- [x] Tests are passing

### DIFF
--- a/test/test-data-validation.jl
+++ b/test/test-data-validation.jl
@@ -84,3 +84,22 @@ end
         ]
     end
 end
+
+@testset "Check only transport flows can be investable" begin
+    connection = DBInterface.connect(DuckDB.DB)
+    _read_csv_folder(connection, joinpath(@__DIR__, "inputs", "Tiny"))
+    # Create all four combinations of is_transport and investable
+    # First set ccgt and ocgt to transport = TRUE
+    DuckDB.query(
+        connection,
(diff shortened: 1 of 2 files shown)


test/test-data-validation.jl
```diff
@@ -84,3 +84,22 @@ end
         ]
     end
 end
+
+@testset "Check only transport flows can be investable" begin
+    connection = DBInterface.connect(DuckDB.DB)
+    _read_csv_folder(connection, joinpath(@__DIR__, "inputs", "Tiny"))
```
datejada, line 90: I was wondering why we need to read the data for a case study instead of creating the tables from scratch. But I think it makes sense since the Tiny case study reads fast, and you need to create at least two tables to do this validation. Is my interpretation correct? @abelsiqueira

datejada (approved):
Hi @abelsiqueira, thanks for the PR; I have more of a curiosity rather than a code question, see the comment I left in the code. So, approving the PR, and feel free to merge after you reply to the comment.

abelsiqueira:
Hi @datejada, thanks for the review. I've updated the test to be more robust. I'll re-request the review so you can let me know what you think.

datejada (approved):
Thank you! New standard indeed 😃 
And I learned that this function `register_data_frame` is also available in DuckDB 😉

PR #1107, opened 2025-03-18
Validate group consistency in asset and group_asset tables

### LINKED ISSUE #1105: Validate consistency between group field in asset and group table
Check if a group in that groups-data file has no members (e.g., assets) in the assets-data file and if an asset in the assets-data file has defined a group that it is not defined in the groups-data file.

abelsiqueira:
- **Make validation loop more maintainable**
- **Validate that the group foreign key has valid values**
- **Validate that any group listed in group_asset has matching assets**

## Related issues

Closes #1105

## Checklist

- [x] Tests are passing

### DIFF
--- a/src/data-validation.jl
+++ b/src/data-validation.jl
@@ -23,15 +23,14 @@ Raises an error if the data is not valid.
 """
 function validate_data!(connection)
     error_messages = String[]
-    @timeit to "no duplicate rows" append!(error_messages, _validate_no_duplicate_rows!(connection))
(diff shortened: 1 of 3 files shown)


datejada (approved):
Thanks! It reads better like this in the data validation, and I like the new tests. It is good to have a PK check while we decide whether to define the PK in DuckDB.

Finally, I love the change to have `_tiny_fixture()` it was bothering me a bit visually so many times the same connection to the Tiny. With the new change, it will be easy to switch the case study to test (if we want to do it)

PR #1119, opened 2025-03-25
Update CompareMPS to not make multiple comments and link to docs

abelsiqueira:
The CompareMPS workflow should find older comments in the PR to replace, instead of creating new comments every time. (This is already done by the Benchmark workflow, I just missed it).
Add the link to the docs to the PR comment.

## Related issues

There is no related issue.

## Checklist

- [x] Tests are passing


datejada (approved):
@abelsiqueira thanks for the PR, I saw the behaviour of the MPS, and it didn't bother me at first, but I for a longer PR with more test that might failed, it is better this way.

PR #1122, opened 2025-03-28
Add simple method multi-year

### LINKED ISSUE #1064: Add simple method for multi-year investments

gnawin:
Add simple method for multi-year investments

This PR aims to modify the existing data structure as little as possible. The first try to implement simple method was in #1068. The idea was to make a substantial difference with the compact method straight from the data structure, so it is much clearer. It was almost done, but I realized that that was too disruptive to the existing data structure, only for the gain of a bit more clarity. This current implementation balances clarity and consistency with the existing data (only use existing files, no new files!). 

A few points of attention:

- Make it explicit in the names which constraints use compact method, and which use simple method. Same for expressions of available capacity. This applies to capacity constraints and upper bounds for `units_on`.
- Without explicit names, it is simple method. This applies to storage and flow constraints. They used to use compact method, now they use simple.
- None method (pure operation case) also uses simple method formulation. It is a choice.

## Related issues

Closes #1064 

## Checklist

- [ ] Tests are passing


src/constraints/capacity.jl
```diff
@@ -360,7 +395,53 @@ function _append_capacity_data_to_indices(connection, table_name)
...
+            AND cons.year = assets_profiles.commission_year
             AND assets_profiles.profile_type = 'availability'
+        LEFT OUTER JOIN assets_profiles AS avail_profile
+            ON cons.asset = avail_profile.asset
+            AND expr_avail.commission_year = avail_profile.commission_year
+            AND avail_profile.profile_type = 'availability'
+        WHERE asset.investment_method in ('simple', 'none')
         GROUP BY cons.id
         ORDER BY cons.id
         ",
```
abelsiqueira, line 447: Since there is only 1 here, you shouldn't need to `GROUP BY` - which also means you don't need `ANY_VALUE`.

src/constraints/capacity.jl
```diff
@@ -198,8 +226,9 @@ function add_capacity_constraints!(connection, model, expressions, constraints,
                         availability_agg *
                         row.capacity *
                         (
-                            row.avail_initial_units * is_charging +
-                            sum(expr_avail[avail_id] for avail_id in row.avail_indices)
+                            row.avail_initial_units * is_charging + sum(
+                                expr_avail_simple_method[avail_id] for avail_id in row.avail_indices
+                            )
```
abelsiqueira, line 231: Isn't it just a single `id`? You shouldn't need the `sum` and the name should be `avail_id`.

src/constraints/capacity.jl
```diff
@@ -360,7 +395,53 @@ function _append_capacity_data_to_indices(connection, table_name)
...
+    return DuckDB.query(
+        connection,
+        "SELECT
+            ANY_VALUE(cons.id) AS id,
+            ANY_VALUE(cons.asset) AS asset,
+            ANY_VALUE(cons.year) AS year,
+            ANY_VALUE(cons.rep_period) AS rep_period,
+            ANY_VALUE(cons.time_block_start) AS time_block_start,
+            ANY_VALUE(cons.time_block_end) AS time_block_end,
+            ANY_VALUE(expr_avail.id) AS avail_indices,
```
abelsiqueira, line 419: Since it's a single one, better to keep the name singular.

src/constraints/transport.jl
```diff
@@ -109,14 +109,14 @@ function _append_transport_data_to_indices(connection)
             ANY_VALUE(cons.time_block_end) AS time_block_end,
             ANY_VALUE(cons.var_flow_id) AS var_flow_id,
             ANY_VALUE(flow.capacity) AS capacity,
-            ARRAY_AGG(expr_avail.id) AS avail_indices,
-            ARRAY_AGG(expr_avail.commission_year) AS avail_commission_years,
+            ANY_VALUE(expr_avail.id) AS avail_indices,
```
abelsiqueira, line 112: If there is no aggregation, there should be no need to `GROUP BY` and this can be singular as well.

abelsiqueira (changes requested):
Thanks, @gnawin. I've left some comments.

src/constraints/capacity.jl
gnawin: Indeed, I removed the `sum` at first, then there was an error that now I realized was due to the `GROUP BY`. Thanks!

gnawin:
@abelsiqueira thanks a lot, I addressed your comments (also changed storage and unit-commitment, where I believe the same comments apply, but let me know if this is fine).

In addition, I added the data validation for simple/none:
- validate we must have all milestone year data
- validate we can not have more data for commission year (which is for the compact method)
This happens for both asset and flow, so 4 checks in total. I wrapped everything in one validation function. On one hand, it is cleaner. On the other hand, to run this validation, we need to have data for both asset and flow (which can be seen in the test). In principle these checks can be put in different functions, but I chose the clean approach. Let me know what you think :).

I also test the 4 conditions using both fake data and Tiny data.

test/test-data-validation.jl
```diff
@@ -218,3 +218,126 @@ end
...
+            :to_asset => ["B", "B", "B", "B"],
+            :milestone_year => [1, 2, 1, missing],
+            :commission_year => [1, 2, 1, missing],
+        )
+        connection = DBInterface.connect(DuckDB.DB)
+        DuckDB.register_data_frame(connection, year_data, "year_data")
+        DuckDB.register_data_frame(connection, asset, "asset")
+        DuckDB.register_data_frame(connection, asset_both, "asset_both")
+        DuckDB.register_data_frame(connection, flow, "flow")
+        DuckDB.register_data_frame(connection, flow_both, "flow_both")
```
abelsiqueira, line 248: This is not testing what you want it to test.
Here the problem is `asset_both` having missing `milestone_year` and `commission_year`, independent of whatever is around.
So you have to build a valid `asset_both` that fails something else.

For this test (consistency between asset_both and simple investment), what you want is

- A test where `asset_both` has only one row for simple asset `a`, and that row is `my != cy`
- A test where `asset_both` has more than one row for simple asset `a`
- A test where `asset_both` has no rows for simple asset `a` (maybe covered by the next paragraph)
- Does `compact` methods require two rows? If so, then
  - a test where `asset_both` has only one row for compact asset `a`
  - a test where `asset_both` has no rows for compact asset `a` (maybe covered by the next paragraph)

---

I started rambling on about other validation tests below, I will leave as to maybe clarify, but if it does not, then please ignore.

There are other missing validation tests, which are not strictly related to simple investment. These are closer to what you were trying to test here, but they should be a different validation (you don't need to implement these here, but if you agree, we should create issues for them (if there isn't one))

- There are combinations of `a` in `asset` and `cy` in `year_data` that are not present in `asset_commission`
- Similar for `my` in `year_data` and `asset_milestone`
- There is at least one row for each asset in asset_both.

Also, the years in these tables are primary keys, so they cannot be `null`, so that, in theory, doesn't need to be checked, as long as we eventually check it via the whole schema validation. #1129 (defaults) has some intersection here.
In the same vain, if we declare the years as foreign keys of `year_data` and validate the schema, then some checks will be unnecessary as well.

cc. @datejada to check if we are in the same page

gnawin: Thanks @abelsiqueira! I think I did not fully get what you meant, but I made a further commit (https://github.com/TulipaEnergy/TulipaEnergyModel.jl/pull/1122/commits/721fb25c69a4608281dfa9116b1246270e7ab25b) to modify tests based on the bullet points **without** changing my test functions. In my test functions, I check whether there is data or not in `asset_both` for the milestone years (from `year_data`). Because for `simple` to work, you must have data for all milestone years (i.e., you cannot miss one), which is the check NULL part in the function. It seems to me that you suggested not checking `year_data`, but I did not fully grasp it, so I did not change it yet. Keep that in mind when reading my comments/latest commit.

I assume by _one row_ you meant _one row per milestone year_? With this assumption, I now have 3 tests covering the first 3 bullet points.

`compact` method requires one row per milestone year (milestone year equal to commission year) so any (non-) investment decisions will be taken into account. It will not break the code, if you have only have this row, or if you do not have this row but have at least one row per milestone year (commission can be anything). However, these two possibilities are not meant for compact method, the first one should use simple/none method, and the second is wrong.

gnawin: @abelsiqueira I updated the data validation after the chat today. Please let me know if they now reflect what we discussed and the error messages are informative enough, then I can further improve if necessary. Thanks!

abelsiqueira (changes requested):
Hi @gnawin, I made some suggestions on the validation. I still get a bit confused, but I think the new queries should make it clearer if we are understanding each other, especially if they end up making the tests change.

I also changed the error message to try to make it a bit more focused on the error from the beginning of the message ("Unexpected (a,my,cy)" vs "Missing row").

I hope this is cleaner, but it's definitely converging. Thanks!

gnawin:
@abelsiqueira thank you very much for the comments. I think the queries that you wrote make things much clearer. I clearly had misunderstandings of your previous comments, and our brief talk the other day did not help much 😄 . 

The confusion from my side are two-fold. On the one hand, I had the impressions that you suggested not use `year_data` to check if all milestone years are covered, because you mentioned that this check would be a more general check not tied to this PR, at least that's how I understood 😄 . On the other hand, I followed the bullet points here https://github.com/TulipaEnergy/TulipaEnergyModel.jl/pull/1122#discussion_r2022411016, but in a too strict way I think (that's why I used count etc.), also because I did not completely follow your logic.

Now given your queries, things are much clearer to me. So we have two validations
1. we check if there is data where milestone year != commission year, and point out the erroring rows
2. we check if all milestone years are covered, and point out the missing milestone years.

I think we converged earlier already but diverged a bit, because here https://github.com/TulipaEnergy/TulipaEnergyModel.jl/pull/1122/commits/082231348c0c471f3f583fa410d0207f7cad4e7b I tried to make the first validation, and here https://github.com/TulipaEnergy/TulipaEnergyModel.jl/pull/1122/commits/082231348c0c471f3f583fa410d0207f7cad4e7b I tried to make the second validation (but not fully correct compare to this current query).

Another note is about this part https://github.com/TulipaEnergy/TulipaEnergyModel.jl/pull/1122/commits/082231348c0c471f3f583fa410d0207f7cad4e7b, you mentioned the possibility to use `asset_milestone`. I think is is possible but only given that `asset_milestone` has no missing rows. This may introduce another validation which I think should be not addressed in this PR.  

Let me know if I can improve anything further. Thanks!

abelsiqueira:
> On the one hand, I had the impression that you suggested not use `year_data` to check if all milestone years are covered, because you mentioned that this check would be a more general check not tied to this PR, at least that's how I understood 😄 . 

These are two things:
- Not using `year_data` would be cleaner to avoid the cross join. But it depends on `asset_milestone` being an appropriate substitute (i.e., do we expect `asset_milestone` to have rows for each combination of asset and milestone year?
- There are more general checks that are not related to simple. For instance, "every (asset, milestone_year) combination needs to have at least one row in asset_both" is a valid condition, but it is not related to simple.

> On the other hand, I followed the bullet points here [#1122 (comment)](https://github.com/TulipaEnergy/TulipaEnergyModel.jl/pull/1122#discussion_r2022411016), but in a too strict way I think (that's why I used count etc.), also because I did not completely follow your logic.

No problem, that was confusing indeed.

> Now given your queries, things are much clearer to me. So we have two validations
> 
> 1. we check if there is data where milestone year != commission year, and point out the erroring rows
> 2. we check if all milestone years are covered, and point out the missing milestone years.
> 

Yes!

> I think we converged earlier already but diverged a bit, because in my initial try here [0822313](https://github.com/TulipaEnergy/TulipaEnergyModel.jl/commit/082231348c0c471f3f583fa410d0207f7cad4e7b) (line 219 - 230) I tried to make the first validation, and here [0822313](https://github.com/TulipaEnergy/TulipaEnergyModel.jl/commit/082231348c0c471f3f583fa410d0207f7cad4e7b) (line 197 - 210) I tried to make the second validation. However, they were not fully correct compare to the current ones.
> 

They do appear to be close to the solution, so your first instinct was good, but my complaint at that time was that the tests were not testing what you needed to test. I hadn't even checked the query yet.

> Another note is about this part [0822313](https://github.com/TulipaEnergy/TulipaEnergyModel.jl/commit/082231348c0c471f3f583fa410d0207f7cad4e7b), you mentioned the possibility to use `asset_milestone`. I think is is possible but only given that `asset_milestone` has no missing rows. This may introduce another validation which I think should be not addressed in this PR.
> 
> Let me know if I can improve anything further. Thanks!

Agreed. Let me make that suggestions as a review

src/data-validation.jl
```diff
@@ -187,3 +188,109 @@ function _validate_group_consistency!(connection)
...
+            WHERE asset.investment_method in ('simple', 'none')
+                AND year_data.is_milestone
+        )
+        SELECT asset_simple.asset, asset_simple.milestone_year, asset_simple.investment_method
+        FROM asset_simple
+        LEFT JOIN asset_both
+            ON asset_simple.asset = asset_both.asset
+            AND asset_simple.milestone_year = asset_both.milestone_year
+            AND asset_simple.milestone_year = asset_both.commission_year
+        WHERE asset_both.commission_year IS NULL
```
abelsiqueira, line 260:
```suggestion
        "SELECT asset_milestone.asset, asset_milestone.milestone_year, asset.investment_method
        FROM asset_milestone
        LEFT JOIN asset
            ON asset_milestone.asset = asset.asset
        LEFT JOIN asset_both
            ON asset_milestone.asset = asset_both.asset
            AND asset_milestone.milestone_year = asset_both.milestone_year
            AND asset_milestone.milestone_year = asset_both.commission_year
        WHERE asset_both.commission_year IS NULL
            AND asset.investment_method in ('simple', 'none')
```

Following our discussion, this avoids using `year_data` and trusts that `asset_milestone` is defined for every `asset` and `milestone_year`.

src/data-validation.jl
```diff
@@ -187,3 +188,109 @@ function _validate_group_consistency!(connection)
...
+        )
+        SELECT flow_simple.from_asset, flow_simple.to_asset, flow_simple.milestone_year,
+        FROM flow_simple
+        LEFT JOIN flow_both
+            ON flow_simple.from_asset = flow_both.from_asset
+            AND flow_simple.to_asset = flow_both.to_asset
+            AND flow_simple.milestone_year = flow_both.milestone_year
+            AND flow_simple.milestone_year = flow_both.commission_year
+        WHERE flow_both.commission_year IS NULL
+        ",
```
abelsiqueira, line 287: Similar change here

abelsiqueira (approved):
Thanks @gnawin, I left a comment on the code for the milestone_year change.
It is optional, because we are already doing what we need to do without the assumption on milestone_year, but if that assumption is fine, the change makes the code simpler.
So I'm pre-approving.

PS. I haven't rechecked everything, only looked at the validation.

src/data-validation.jl
gnawin: I agree that this is fine to assume that, because I guess in one way or another we would validation that in the future. So it simplifies the code here.

gnawin:
@abelsiqueira thank you! It's fine because I did not changed the rest.

Since this is a relatively big PR, @datejada would you also have a look to see if you have any suggestions/spot anything strange? Thanks.

datejada (approved):
@gnawin thanks for the PR, I rebased it just to check that all was working after the merge of the previous PR. I don't have more comments. Good work here 😄

PR #1124, opened 2025-03-28
Add option to use JuMP direct model

### LINKED ISSUE #1031: Add option to use direct model in the JuMP model creation
### Description

This option can be used to create a `direct_model` instead of `model` to reduce memory allocation (and potentially solution time), see docs here:

https://jump.dev/JuMP.jl/stable/api/JuMP/#direct_model

example:
```julia
model = direct_model(HiGHS.Optimizer())
```

### Validation and testing

- [ ] Use the benchmark case study with and without the option and compare the memory allocations. Test using HiGHS, Gurobi, and Xpress.
- [ ] Add a test that uses the option
- [ ] Add a docs section (in the how-to?) explaining the use of this option and the (enable_names) to speedup the model (something like `performance tips`?). 

### Motivation

Have options to reduce the memory allocations of the model

### Target audience

Developers

### Can you help?

Always ;)

clizbe:
## Related issues

Closes #1031

## Checklist

- [x] Tests are passing

### DIFF
--- a/docs/src/20-how-to-use.md
+++ b/docs/src/20-how-to-use.md
@@ -60,7 +60,7 @@ To change this, you can give the functions [`run_scenario`](@ref) or [`create_mo
 
 Here is an example running the Tiny case using the [GLPK](https://github.com/jump-dev/GLPK.jl) optimizer:
 
-```julia @example change-optimizer
+```julia
 using DuckDB, TulipaIO, TulipaEnergyModel, GLPK
 
 input_dir = "../../test/inputs/Tiny" # you path will be different
@@ -77,7 +77,7 @@ In both cases above, the `GLPK` optimizer uses its default parameters, which you
 To change any optimizer parameters, you can pass a dictionary to the `optimizer_parameters` keyword argument.
 The example below changes the maximum allowed runtime for GLPK to 1 second, which will probably cause it to fail to converge in time.
 
-```julia @example change-optimizer
+```julia
 # change the optimizer parameters
 parameter_dict = Dict("tm_lim" => 1) # list optimizer parameters as comma-separated parameter=>value pairs
 energy_problem = run_scenario(connection; optimizer = GLPK.Optimizer, optimizer_parameters = parameter_dict)
--- a/docs/src/20-how-to-use.md
+++ b/docs/src/20-how-to-use.md
@@ -60,34 +60,49 @@ To change this, you can give the functions [`run_scenario`](@ref) or [`create_mo
 
 Here is an example running the Tiny case using the [GLPK](https://github.com/jump-dev/GLPK.jl) optimizer:
 
-```@example change-optimizer
+```julia @example change-optimizer
 using DuckDB, TulipaIO, TulipaEnergyModel, GLPK
 
-input_dir = "../../test/inputs/Tiny" # hide
-# input_dir should be the path to Tiny as a string (something like "test/inputs/Tiny")
+input_dir = "../../test/inputs/Tiny" # you path will be different
 connection = DBInterface.connect(DuckDB.DB)
 read_csv_folder(connection, input_dir; schemas = TulipaEnergyModel.schema_per_table_name)
 energy_problem = run_scenario(connection; optimizer = GLPK.Optimizer)
-#OR model = create_model!(energy_problem; optimizer = GLPK.Optimizer)
+#OR create_model!(energy_problem; optimizer = GLPK.Optimizer)
 ```
 
 !!! info
     Notice that you need to add the GLPK package and run `using GLPK` before running `GLPK.Optimizer`.
 
 In both cases above, the `GLPK` optimizer uses its default parameters, which you can query using [`default_parameters`](@ref).
-To change any optimizer parameters, you can pass a dictionary to the `parameters` keyword argument.
+To change any optimizer parameters, you can pass a dictionary to the `optimizer_parameters` keyword argument.
 The example below changes the maximum allowed runtime for GLPK to 1 second, which will probably cause it to fail to converge in time.
 
-```@example
-using DuckDB, TulipaIO, TulipaEnergyModel, GLPK
-
-input_dir = "../../test/inputs/Tiny" # your input directory will be different
-connection = DBInterface.connect(DuckDB.DB)
-read_csv_folder(connection, input_dir; schemas = TulipaEnergyModel.schema_per_table_name)
-
-parameter_dict = Dict("tm_lim" => 1) # list as comma-separated parameter=>value pairs
-energy_problem = run_scenario(connection; optimizer = GLPK.Optimizer, parameters = parameter_dict)
+```julia @example change-optimizer
+# change the optimizer parameters
+parameter_dict = Dict("tm_lim" => 1) # list optimizer parameters as comma-separated parameter=>value pairs
+energy_problem = run_scenario(connection; optimizer = GLPK.Optimizer, optimizer_parameters = parameter_dict)
+#OR create_model!(energy_problem; optimizer = GLPK.Optimizer, optimizer_parameters = parameter_dict)
 energy_problem.termination_status
+```
+
+If `direct_model = false` you can change the optimizer and parameters after creating the model (but before solving it) using the JuMP commands demonstrated below.
+For more information on `direct_model`, see [Speed improvements in the model creation](@ref need-for-speed).
+
+```julia @example change-optimizer
+# create the model and solve with the default optimizer and optimizer parameters
+energy_problem = EnergyProblem(connection)
+create_model!(energy_problem)
+solution1 = solve_model(energy_problem)
+
+# change the solver and parameters and resolve:
+parameter_dict = Dict("tm_lim" => 1) # list optimizer parameters as comma-separated parameter=>value pairs
+
+JuMP.set_optimizer(energy_problem.model, GLPK.Optimizer) # change the optimizer
+for (k, v) in optimizer_parameters
+    JuMP.set_attribute(energy_problem.model, k, v) # change the optimizer_parameters
+end
+
+solution2 = solve_model(energy_problem) # solve the model with new optimizer & optimizer_parameters
 ```
 
 For the complete list of parameters, check your chosen optimizer.
@@ -105,9 +120,9 @@ Use `energy_problem.model` for the model argument. For instance:
 
 ```julia
 if energy_problem.termination_status == INFEASIBLE
- compute_conflict!(energy_problem.model)
- iis_model, reference_map = copy_conflict(energy_problem.model)
- print(iis_model)
+  compute_conflict!(energy_problem.model)
+  iis_model, reference_map = copy_conflict(energy_problem.model)
+  print(iis_model)
 end
 ```
 
@@ -129,15 +144,17 @@ For more information, see the JuMP documentation for [Disable string names](http
 
 ### Create a direct model
 
-If you want to reduce memory usage, consider using JuMP's [`direct_model`](https://jump.dev/JuMP.jl/stable/api/JuMP/#direct_model). This restricts certain actions after model creation (see JuMP docs), for instance you can no longer set the optimizer and parameters in [`solve_model`](@ref).
+If you want to reduce memory usage, consider using `direct_model = true`. This restricts certain actions after model creation, such as changing the optimizer.
 
 ```julia
-# Create a direct model
-create_model!(energy_problem; direct_model = true)
+# Create direct model with run_scenario
+run_scenario(connection; direct_model = true)
 
-# Create a direct model while specifying a solver and attributes
-create_model!(energy_problem; direct_model = true, optimizer = HiGHS.Optimizer, parameters = Dict("presolve" => "off", MOI.Silent() => true))
+# OR while using create_model!
+create_model!(energy_problem; direct_model = true)
 ```
+
+For more information, see the JuMP documentation for [`direct_model`](https://jump.dev/JuMP.jl/stable/api/JuMP/#direct_model).
 
 ## Storage specific setups
 
--- a/docs/src/20-how-to-use.md
+++ b/docs/src/20-how-to-use.md
@@ -122,6 +122,8 @@ end
 
 ## [Speed improvements in the model creation](@id need-for-speed)
 
+### Disable names of variables and constraints
+
 If you want to speed-up model creation, consider disabling the naming of variables and constraints. Of course, removing the names will make debugging difficult (or impossible) - so enable/disable naming as needed for your analysis.
 
 ```julia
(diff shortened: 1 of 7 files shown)


clizbe:
Not sure if this is the correct way of benchmarking this. 🤔
More likely I should run a case for speed with and without this feature... ?

datejada:
> Not sure if this is the correct way of benchmarking this. 🤔 More likely I should run a case for speed with and without this feature... ?

Nah, that is related to JuMP. We are enabling the option in Tulipa, but whether it actually works or not is a JuMP issue and should be reported there.

datejada (commented):
@clizbe thanks for the update in the docs, I love "need for speed" as the id of the docs section, great reference to a great game 😄 

I have one last request 🙈 would you mind adding the option to the `run_scenario` function? as it is the enable names one. So, users can activate it when using `run_scenario` too; otherwise to use this option, you will need go run step by step the functions inside `run_scenario`.

Please let me know if you agree, or I am missing something here.

test/test-options.jl
```diff
@@ -17,6 +17,13 @@
     )
 end
 
+@testset "Test enable_names & direct_model usage" begin
+    connection = DBInterface.connect(DuckDB.DB)
+    _read_csv_folder(connection, joinpath(INPUT_FOLDER, "Tiny"))
+    energy_problem = TulipaEnergyModel.run_scenario(connection;)
+    TulipaEnergyModel.create_model!(energy_problem; enable_names = false, direct_model = true)
```
datejada, line 24: if you add the option to the `run_scenario` function, then the test is much simple, since you just need to call the `run_scenario` function with both keywords 😉

clizbe:
@datejada I'm adding it to run_scenario now. But it's a bit weird because of the different places in the arguments the user specifies the solver and parameters. And they can't change the solver in solve_model so I don't think I should pass those arguments to that function...

clizbe:
If this looks fine to you I can update the docs. But it seems janky to me because now this one little option has added two if-statements in different places.

datejada:
@clizbe I see what you mean, and I agree that it feels too much 🤔 for just an option.

So, @abelsiqueira, we want to enable the option to use the JuMP direct model since Oscar and Joaquim recommend it for reducing memory usage and improving speed. But it looks like making it available as it is right now would require a bit too much change in the code. Would you happen to have any ideas or suggestions to make it cleaner or straightforward?

Thanks!

abelsiqueira:
Hi both, here's what I identified:

- `optimizers_with_attributes` is just a way to set the parameters on model creation (https://jump.dev/JuMP.jl/stable/manual/models/#solver_options), so it is equivalent to what we do in `solve_model`/`run_scenario`, i.e., create a solver and set attributes. We should not have both approaches.
- The use of `direct_model` means that the solver cannot be changed. So `model` (the object) creation is attached to the solver specification, although the attributes can be changed.
- What changes for EnergyProblem

I see two ways of solving this (they can be combined, but ignore this for now):

1. Always pass the solver (and parameters?) to `create_model`, and use it for both `direct_model` and `Model`. `run_scenario` only needs the boolean flag `direct_model` because the `optimizer` and `parameters` are already there. Then remove the `optimizer` argument from `solve_model`, because you have already passed the `optimizer` to `create_model`. These arguments should be passable to `create_model!(energy_problem, ...)`, so probably the EnergyProblem API doesn't need to change much.
2. Separate the "model object" creation from the model creation, and attach the solver at that point. Then, pass `model` as an argument to `create_model` and `solve_model`, and never pass the `optimizer`. Then we will need a new function `create_model_object`, but `create_model` gets smaller. In this case, `run_scenario` simply calls the new function before calling `create_model`. Maybe the EnergyProblem API can be improved, not clear yet.

Neither option should require the conditionals in `run_scenario`. 

The `test-pipeline.jl` will need to be updated.

Let me know if this is not clear enough, writing walls of text never feels clear.

datejada:
@abelsiqueira thanks, I think both suggestions make sense since they avoid the conditional in the run_scenario function. I don't have a strong opinion, so I leave it to @clizbe to choose the one she feels is easier or makes more sense to her. (Since you have been changing the code, you might have a strong preference.)

clizbe:
I think I follow, but doesn't this remove the user's capability of creating the model once and solving it multiple times with different solvers and parameters?

datejada:
> Oh, but I guess only if they use run_scenario. If they run steps individually they can pass the optimiser argument to solve_model?

Yeap, run_scenario is just a wrapper, so the user can always run the steps individually and make changes when doing multiple runs (although I think most of time you will run with the same solver parameters...anyway, there are some cases where you might want to change the solver options...)

abelsiqueira:
> I think I follow, but doesn't this remove the user's capability of creating the model once and solving it multiple times with different solvers and parameters?

- Using `direct_model` constrains the solver to the model at the JuMP level (nothing we can do about it).
- It is always possible to use `set_attribute` to set the parameters. Either manually/explicitly on the `model`, or inside one of our functions (like `solve_model` does now)
- In both implementation strategies, we should still call `set_attribute` with the default parameters/parameters passed by the user.
- We want to tie the implicit calls to `set_attribute` (e.g., to apply the default parameters) with the model creation. Because we want to make sure that (i) we give defaults if the user gives nothing; (ii) we don't override the user parameters, whenever the user gives them.
  - This means that if the model is created at an earlier stage, then we also have to change the `parameters` in `solve_model`, so we don't override the user.

> Oh, but I guess only if they use run_scenario. If they run steps individually they can pass the optimiser argument to solve_model?

`run_scenario` allows the `optimizer` and `parameters` to be passed, so all the comments above still apply here. By the end of `run_scenario`, they have the `energy_problem`, so all restrictions above apply.

@clizbe, one idea is to look into `test/test-pipeline.jl` and think like the user in the cases (what if I want to use direct_model? What if I want to change the parameters?) for `run_scenario`, for explicit EnergyProblem, and for the low-level (no EP) cases.

clizbe:
Okay I thiiiink I implemented Abel's first idea correctly.
Tests passed, but then I had to deconflict a rebase (because of my friendly-docs) so we'll see.

clizbe:
Also I standardized the arguments as "optimizer" and "parameters" - maybe it should be "optimizer_parameters?"
And I moved 'optimizer' to a keyword argument everywhere. So this will probably change OBZ.

test/test-options.jl
datejada, line 24: @clizbe unless I am missing something, I think this can be simplified like this and still get the goal of the test:

```suggestion
    energy_problem = TulipaEnergyModel.run_scenario(connection; enable_names = false, direct_model = true)
```

abelsiqueira (changes requested):
Thanks for the update. I've made some comments, in particular because `create_model` is now the starting point for the solver and its parameters, but `solve_model` still changes both these things.
For instance, I think `run_scenario(connection, direct_model=true)` fails at the moment because the `optimizer` is being set twice.
I think I left comments in the relevant places, let me know if a piece is missing.

abelsiqueira:
Forgot to mention that changing to optimizer_parameters is indeed better.

src/solve-model.jl
```diff
@@ -43,12 +45,12 @@ using [`read_parameters_from_file`](@ref).
...
 parameters = Dict{String,Any}("presolve" => "on", "time_limit" => 60.0, "output_flag" => true)
-solve_model(model, HiGHS.Optimizer; parameters = parameters)
+solve_model(model; optimizer = HiGHS.Optimizer, parameters = parameters)
 ```
 """
 function solve_model(
-    model::JuMP.Model,
-    optimizer = HiGHS.Optimizer;
+    model::JuMP.Model;
+    optimizer = HiGHS.Optimizer,
```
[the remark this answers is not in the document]
clizbe: I thought I handled this by leaving the optimizer+parameters argument out of the call to solve_model in run_scenario.
So when using run_scenario it sets the optimizer in create_model every time - but if they go step-by-step they have the option of specifying in solve-model as long as it's not a direct_model, which is in the docs.

But maybe that's not what it actually does or it's not the right way to do it?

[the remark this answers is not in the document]
clizbe: Ah I get it - using run_scenario, when it calls solve_model(energy_problem) it'll pass the default even if they changed it in create_model. And like you said, it passes the solver even if they don't specify it. :/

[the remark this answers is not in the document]
clizbe: I looked into using multiple dispatch on solve_model! to have the best of all worlds. But then there's more duplication for solve_model! and solve_model, unless we remove solve_model from user control.

Anyway it seems like a lot of hassle for one little switch. If we say create_model controls setting the solver, the user can still change the solver & parameters with the JuMP commands as long as they're not using a direct_model.

clizbe:
Alrighty, how about this?

Also I think our tests are bit weird that we have to set an input_dir and load a connection for every single thing. :/

test/test-options.jl
```diff
@@ -17,6 +17,47 @@
     )
 end
 
+@testset "Test run_scenario arguments" begin
+    connection = DBInterface.connect(DuckDB.DB)
+    _read_csv_folder(connection, joinpath(INPUT_FOLDER, "Tiny"))
```
abelsiqueira, line 22:
```suggestion
    connection = _tiny_fixture()
```
This is new-ish, we are only using it in the test-data-validation.jl at the moment

test/test-options.jl
```diff
@@ -17,6 +17,47 @@
...
+
+    @test JuMP.get_attribute(ep.model, "msg_lev") == GLPK.GLP_MSG_OFF
+    @test JuMP.get_attribute(ep.model, "tol_int") == 1e-3
+    @test JuMP.constraint_by_name(ep.model, "consumer_balance[demand,2030,1,1:1]") === nothing
+    @test JuMP.mode(ep.model) == JuMP.ModelMode(2)
+end
+
+@testset "Test create_model! arguments" begin
+    connection = DBInterface.connect(DuckDB.DB)
+    _read_csv_folder(connection, joinpath(INPUT_FOLDER, "Tiny"))
```
abelsiqueira, line 44:
```suggestion
    connection = _tiny_fixture()
```

docs/src/20-how-to-use.md
```diff
@@ -52,51 +52,57 @@ To save the solution to CSV files, you can use [`export_solution_to_csv_files`](
...
+By default, the model is solved using the [HiGHS](https://github.com/jump-dev/HiGHS.jl) optimizer (or solver).
+To change this, you can give the functions [`run_scenario`](@ref) or [`create_model!`](@ref) a different optimizer.
 
 !!! warning
     HiGHS is the only open source solver that we recommend. GLPK and Cbc are not (fully) tested for Tulipa.
 
 Here is an example running the Tiny case using the [GLPK](https://github.com/jump-dev/GLPK.jl) optimizer:
 
-```@example change-optimizer
+```julia @example change-optimizer
```
abelsiqueira, line 63: You don't want to execute the commands here anymore? If so, remove `@example change-optimizer`, otherwise, remove `julia`.

docs/src/20-how-to-use.md
```diff
@@ -52,51 +52,57 @@ To save the solution to CSV files, you can use [`export_solution_to_csv_files`](
...
-connection = DBInterface.connect(DuckDB.DB)
-read_csv_folder(connection, input_dir; schemas = TulipaEnergyModel.schema_per_table_name)
-energy_problem = run_scenario(connection, optimizer = GLPK.Optimizer, parameters = parameters)
-energy_problem.termination_status
+JuMP.set_optimizer(energy_problem.model, GLPK.Optimizer) # change the optimizer
+for (k, v) in optimizer_parameters
+    JuMP.set_attribute(energy_problem.model, k, v) # change the optimizer_parameters
+end
+
+solution2 = solve_model(energy_problem) # solve the model with new optimizer & optimizer_parameters
```
abelsiqueira, line 105:
```suggestion
solve_model(energy_problem) # solve the model with new optimizer & optimizer_parameters
```
There is no solution output anymore.

docs/src/20-how-to-use.md
```diff
@@ -52,51 +52,57 @@ To save the solution to CSV files, you can use [`export_solution_to_csv_files`](
...
-In both cases, the `GLPK` optimizer uses its default parameters,
-which you can query using [`default_parameters`](@ref).
-To change any optimizer parameters, you can pass a dictionary using the keyword argument `parameters`.
-The example below changes the maximum allowed runtime for GLPK to 1 second,
-which will probably cause it to fail to converge in time.
+```julia @example change-optimizer
+# create the model and solve with the default optimizer and optimizer parameters
+energy_problem = EnergyProblem(connection)
+create_model!(energy_problem)
+solution1 = solve_model(energy_problem)
```
abelsiqueira, line 95:
```suggestion
solve_model(energy_problem)
```
There is no solution output anymore.

docs/src/20-how-to-use.md
```diff
@@ -52,51 +52,57 @@ To save the solution to CSV files, you can use [`export_solution_to_csv_files`](
...
+    Notice that you need to add the GLPK package and run `using GLPK` before running `GLPK.Optimizer`.
 
-```@example change-optimizer
-using GLPK
+In both cases above, the `GLPK` optimizer uses its default parameters, which you can query using [`default_parameters`](@ref).
+To change any optimizer parameters, you can pass a dictionary to the `optimizer_parameters` keyword argument.
+The example below changes the maximum allowed runtime for GLPK to 1 second, which will probably cause it to fail to converge in time.
 
-solution = solve_model!(energy_problem, GLPK.Optimizer)
+```julia @example change-optimizer
```
abelsiqueira, line 80: Same here

abelsiqueira (approved):
Thanks, @clizbe. I made some small suggestions only, so I am pre-approving.
If you accept the suggestions, please check that the tests continue passing, because the linter might fail due to formatting.

docs/src/20-how-to-use.md
clizbe: Aww I wanted it to both execute the commands and show as julia format.

docs/src/20-how-to-use.md
clizbe, line 80:
```suggestion
```julia
```

docs/src/20-how-to-use.md
abelsiqueira: The `@example` directive is Documenter, it will execute and translate into two blocks, the code with `julia`, and the output with a plain text. So it will have both if you just delete `julia`. E.g.: https://tulipaenergy.github.io/TulipaEnergyModel.jl/stable/20-how-to-use/#Changing-the-solver-(optimizer)-and-specifying-parameters

datejada (approved):
Great work here! thanks, merge it when you are ready 😄

clizbe:
@datejada I learned a lot! :D

PR #1130, opened 2025-03-31
Create function to populate defaults

### LINKED ISSUE #1129: Create a function that populates tables in a connection with defaults from the schema
Create a function that populates tables in a connection with defaults from the schema.
(Was "Create a function that reads those files and fills the optional data with the defaults defined in the JSON file").

abelsiqueira:
WIP

## Related issues

Closes #1129

## Checklist

- [ ] Tests are passing

### DIFF
--- a/test/data-simplest.jl
+++ b/test/data-simplest.jl
@@ -32,18 +32,10 @@ const simplest_data = Dict(
     "asset_commission" => DataFrame(
         :asset => ["some_producer", "some_consumer"],
         :commission_year => [2030, 2030],
-        :fixed_cost => [1.0, 2.0],
-        :fixed_cost_storage_energy => [1.0, 2.0],
-        :investment_cost => [1.0, 2.0],
-        :investment_cost_storage_energy => [1.0, 2.0],
-        :investment_limit => [1.0, 2.0],
-        :investment_limit_storage_energy => [1.0, 2.0],
     ),
     "asset_milestone" => DataFrame(
         :asset => ["some_producer", "some_consumer"],
         :milestone_year => [2030, 2030],
-        :min_energy_timeframe_partition => [1.0, 1.0],
-        :max_energy_timeframe_partition => [2.0, 2.0],
     ),
 
     # Basic flow data
@@ -58,9 +50,6 @@ const simplest_data = Dict(
         :from_asset => ["some_producer"],
         :to_asset => ["some_consumer"],
         :commission_year => [2030],
-        :fixed_cost => [1.0],
-        :investment_cost => [1.0],
-        :investment_limit => [1.0],
     ),
     "flow_milestone" => DataFrame(
         :from_asset => ["some_producer"],
@@ -69,33 +58,10 @@ const simplest_data = Dict(
     ),
 
     # Basic time information
-    "year_data" => DataFrame(:year => [2030], :length => [8760], :is_milestone => [true]),
-    "rep_periods_data" =>
-        DataFrame(:year => [2030, 2030], :rep_period => [1, 2], :num_timesteps => [24, 24]),
-    "timeframe_data" => DataFrame(:year => 2030, :period => 1:365, :num_timesteps => 24),
+    "year_data" => DataFrame(:year => [2030]),
+    "rep_periods_data" => DataFrame(:year => [2030, 2030], :rep_period => [1, 2]),
+    "timeframe_data" => DataFrame(:year => 2030, :period => 1:365),
     "rep_periods_mapping" =>
         DataFrame(:year => 2030, :period => 1:365, :rep_period => mod1.(1:365, 2)),
-
-    # Profiles
-    "assets_profiles" => DataFrame(
-        "asset" => ["some_producer"],
-        "commission_year" => [2030],
-        "profile_name" => ["some_producer_profile"],
-        "profile_type" => ["availability"],
-    ),
-    "flows_profiles" => DataFrame(
-        "from_asset" => ["some_producer"],
-        "to_asset" => ["some_consumer"],
-        "year" => [2030],
-        "profile_name" => ["some_flow_profile"],
-        "profile_type" => ["availability"],
-    ),
-    "profiles_rep_periods" => DataFrame(
-        "profile_name" => "some_producer_profile",
-        "year" => 2030,
-        "rep_period" => [fill(1, 24); fill(2, 24)],
-        "timestep" => [1:24; 1:24],
-        "value" => ones(48),
-    ),
 )
 end
--- a/test/data-simplest.jl
+++ b/test/data-simplest.jl
@@ -1,0 +1,101 @@
+# Repository of data to be used for the tests
+# All data should be NamedTuples of
+
+function _create_connection_from_dict(data::Dict{String,DataFrame})
+    connection = DBInterface.connect(DuckDB.DB)
+
+    for (table_name::String, table::DataFrame) in data
+        # Check that these `table_name` exist in the schema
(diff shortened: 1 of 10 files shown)


test/data-simplest.jl
```diff
@@ -0,0 +1,120 @@
...
+        "profile_type" => ["availability"],
+    ),
+    "profiles_rep_periods" => DataFrame(
+        "profile_name" => "some_producer_profile",
+        "year" => 2030, # TODO: Should be required
+        "rep_period" => [fill(1, 24); fill(2, 24)], # TODO: should be required
+        "timestep" => [1:24; 1:24], # TODO: should be required
+        "value" => ones(48),
+    ),
+)
```
abelsiqueira, line 119: @datejada @gnawin I'm creating the simplest data that has no defaults. Can you check these? Notice I added some "TODO" to check whether the listed data should have defaults or not. 
I am assuming that this data is for TEM after clustering, so rep_period things are included as required.
For some, I guess it's convenient for the user to have them friendly included, but for TEM I don't think that should be the case (like the years).

datejada, line 119: @abelsiqueira this is what I have in mind. My criterion is: which are the primary keys of each table? in other words, which values are unique for a set of columns? The ones that are not PKs should be fill in with the defaults. Let me know what do you think 😄 

```suggestion
const simplest_data = Dict(
    # Basic asset data
    "asset" => DataFrame(
        :asset => ["some_producer", "some_consumer"],
        :type => ["producer", "demand"],
    ),
    "asset_both" => DataFrame(
        :asset => ["some_producer", "some_consumer"],
        :commission_year => [2030, 2030],
        :milestone_year => [2030, 2030],
    ),
    "asset_commission" => DataFrame(
        :asset => ["some_producer", "some_consumer"],
        :commission_year => [2030, 2030],
    ),
    "asset_milestone" => DataFrame(
        :asset => ["some_producer", "some_consumer"],
        :milestone_year => [2030, 2030],
    ),

    # Basic flow data
    "flow" => DataFrame(
        :from_asset => ["some_producer"],
        :to_asset => ["some_consumer"],
    ),
    "flow_both" => DataFrame(
        :from_asset => ["some_producer"],
        :to_asset => ["some_consumer"],
        :commission_year => [2030],
        :milestone_year => [2030],
    ),
    "flow_commission" => DataFrame(
        :from_asset => ["some_producer"],
        :to_asset => ["some_consumer"],
        :commission_year => [2030],
    ),
    "flow_milestone" => DataFrame(
        :from_asset => ["some_producer"],
        :to_asset => ["some_consumer"],
        :milestone_year => [2030],
    ),

    # Basic time information
    "year_data" => DataFrame(
        :year => [2030],
    ),
    "rep_periods_data" => DataFrame(
        :year => [2030],
        :rep_period => [1],
    ),
    "timeframe_data" => DataFrame(
        :year => [2030],
        :period => [1],
    ),
    "rep_periods_mapping" => DataFrame(
        :year => [2030],
        :period => [1],
        :rep_period => [1],
    ),

    # Profiles
    "assets_profiles" => DataFrame(
        "asset" => ["some_producer"],
        "commission_year" => [2030],
    ),
    "flows_profiles" => DataFrame(
        "from_asset" => ["some_producer"],
        "to_asset" => ["some_consumer"],
        "year" => [2030],
        "profile_name" => ["some_flow_profile"],
    ),
    "profiles_rep_periods" => DataFrame(
        "profile_name" => "some_producer_profile",
        "year" => 2030,
        "rep_period" => [1],
        "timestep" => [1],
    ),
)
```

gnawin:
Hi @abelsiqueira  @datejada  I'm not entirely following the discussion. From what I see ihere, we start from the `simplest_data` (with some _keys_/columns, called primary keys?) and then populate defaults for other _keys_/columns (called foreign keys?). 

What is not clear to me is: what is the border between the two? Do we expect the model can run after `simplest_data`? 

Also the defaults can be the default option we recommend, but also can be the default values for the keys so that model can run. Which is the correct interpretation?

test/data-simplest.jl
abelsiqueira: Thanks for the reply, @datejada. The primary keys for sure cannot be ignored, but there can be other columns that are necessary. The main issue:

- What values should be used as default for the columns that don't have a default yet (*_cost in asset_commission, value in profiles_rep_periods, etc.)?

I am expecting to solve the problem that is generated with this "simplest-data", so maybe to have it make sense, other fields are required?

For the "profiles_rep_periods", if I don't need the value, then do I need the file? It can be part of the "tables_allowed_to_be_missing", otherwise.

abelsiqueira:
> Hi @abelsiqueira @datejada I'm not entirely following the discussion. From what I see ihere, we start from the `simplest_data` (with some _keys_/columns, called primary keys?) and then populate defaults for other _keys_/columns (called foreign keys?).

Some terms:
- "primary keys" are the columns that form the "key" for the table. I.e., if it were a dictionary, it would be from `(primary_key1, ..., primary_keyN)` to the rest of the row. This means that they are unique, or distinct.
- "foreign keys" are the columns whose values are takes from some other table. For instance, `asset.milestone_year` is a foreign key because its values come from `year_data.year`.

Back to this PR. We start with some data (in this case `simplest_data`), and then use this function to complete the table with the other tables that are required for TulipaEnergyModel.

> What is not clear to me is: what is the border between the two? Do we expect the model can run after `simplest_data`?
> 

This is a great question. My expectation is that it should (after populating with the defaults), but @datejada's suggestion makes me think we are not aligned.
That said, the objective of the data is to have a minimum set of columns that **all** users need, independent of whatever application.

> Also the defaults can be the default option we recommend, but also can be the default values for the keys so that model can run. Which is the correct interpretation?

These are default values that have no meaning to you as user, or whose meaning are implicit. For instance, if you don't provide a "min_energy", then we use 0, because that's the implicit physical limitation. If you don't provide "investable", then it defaults to `false`, because if the user is not concerned about an asset being investable, then it shouldn't be.

It is NOT a default for missing values, it is for missing columns, so it is not about coalescing null values, but instead, preparing the data for TEM.

The alternative would be to check existence of these things in the code directly. For instance, `if has_column("investable"), add_investment_variable!(...)`.

@gnawin, are these the types of defaults you were concerned about?

test/data-simplest.jl
datejada: @abelsiqueira Thanks! Regarding your questions:

1. We should add the defaults that are missing in the JSON file. The ones you where suggesting, are good as an example. I would add those to the JSON file.
2. "profiles_rep_periods", "assets_profiles", "flows_profiles" can be added to the "tables_allowed_to_be_missing", since if there are not values there we assume a default of 1. So, they can be empty.

With that, it should be enough to run the simplest data + defaults + allow it to be missing without problems.

Let me know if something else I am missing 😄 we can have a call if needed

gnawin:
@abelsiqueira thanks for explaining. 

>My expectation is that it should (after populating with the defaults), but @datejada's suggestion makes me think we are not aligned.
That said, the objective of the data is to have a minimum set of columns that all users need, independent of whatever application.

Here it suggests that the model should run (after populating with the defaults, i.e., missing columns as you mentioned.) Whether we include the bracket of not is important, which will answer my question regarding the borders. If we look at them as a whole, then I don't have a strong opinion on what to include in the `simplest_data` and what will be populated afterwards. I guess my point is just that we should align on this.

datejada (approved):
Thanks! It looks good to me for TEM. So, I am approving to test how I can incorporate this function into the current OBZ workflow 😄

PR #1137, opened 2025-04-04
Add feature self loss storage

### LINKED ISSUE #1097: Add self-discharge parameter for the storage balance constraint
### Description

Self-discharge is a feature that helps model storage losses. This is relevant for seasonal storage. For example, an aquifer thermal energy storage (ATES) to store warm/cold water underground (seasonal storage) has dissipation losses over time, dependent on the water temperature and storage duration.

The storage balance equation will be something like this:

`storage(t) = (1 - SF) * storage(t-1) + EFF_IN * charging(t) - discharge(t) / EFF_OUT`

`SF`: self-discharge parameter

The default value of this parameter is 0.0

### Validation and testing

Modify the Norse case study to have two storage assets with this parameter set to values different to 0.0, and check the constraints in the LP (or MPS) file.

### Motivation

This is motivated by the case study in the NextGenOpt project to model the UU campus energy system.

### Target audience

Develepers

### Can you help?

Always!

gnawin:
## Related issues

Closes #1097 

## Checklist

- [x] Tests are passing


src/constraints/storage.jl
```diff
@@ -60,8 +60,11 @@ function add_storage_constraints!(connection, model, variables, expressions, con
                         @constraint(
                             model,
                             var_storage_level[row.id] ==
-                            previous_level + profile_agg * row.storage_inflows + incoming_flow -
-                            outgoing_flow,
+                            (
+                                1 - row.storage_loss_from_stored_energy
+                            )^length(row.time_block_start:row.time_block_end) * previous_level +
```
abelsiqueira, line 65: Add to validation list:
- `storage_loss_from_stored_energy` between 0 and (strictly?) less than 1

abelsiqueira, line 65: I suggest pre-computing `(1 - r)^length` before the `@constraint`. Something like this:

```julia
computed_storage_loss_coef = 1.0
if row.storage_loss_from_stored_energy > 0.0
    duration = row.time_block_end - row.time_block_start + 1
    computed_storage_loss_coef = (1 - row.storage_loss_from_stored_energy)^duration
end
```

src/constraints/storage.jl
```diff
@@ -172,7 +175,10 @@ function add_storage_constraints!(connection, model, variables, expressions, con
                         @constraint(
                             model,
                             var_storage_level_over_clustered_year.container[row.id] ==
-                            previous_level + inflows_agg + incoming_flow - outgoing_flow,
+                            (1 - row.storage_loss_from_stored_energy)^row.duration_period_block *
```
abelsiqueira, line 178: Same here

src/constraints/storage.jl
```diff
@@ -243,12 +249,45 @@ function add_storage_constraints!(connection, model, variables, expressions, con
...
+                con.asset,
+                con.year,
+                con.period_block_start,
+                SUM(mapping.num_timesteps) AS duration_period_block
+            FROM cons_balance_storage_over_clustered_year AS con
+            LEFT JOIN timeframe_data AS mapping
+                ON mapping.year = con.year
+                AND mapping.period BETWEEN con.period_block_start AND con.period_block_end
+            GROUP BY con.asset, con.year, con.period_block_start
+            ORDER BY con.asset, con.year, con.period_block_start
```
abelsiqueira, line 270:
```suggestion
                cons.asset,
                cons.year,
                cons.period_block_start,
                SUM(mapping.num_timesteps) AS duration_period_block
            FROM cons_balance_storage_over_clustered_year AS cons
            LEFT JOIN timeframe_data AS mapping
                ON mapping.year = cons.year
                AND mapping.period BETWEEN cons.period_block_start AND cons.period_block_end
            GROUP BY cons.asset, cons.year, cons.period_block_start
```

Changed `con` to `cons` for consistency with other parts (like below), and remove the ordering because this will be joined with another table, so it isn't needed.

abelsiqueira (approved):
Thanks, @gnawin. I've made a few suggestions that shouldn't change much the resulting code, so I'm pre-approving.

src/constraints/storage.jl
gnawin: Added to #461

datejada (approved):
@gnawin Thanks for the PR; it includes the changes we commented on yesterday in person, so I approve it.

Issue #1168, opened 2025-04-24
Enforce transport flows have free signs


gnawin:
### Description

Transport flows should have free signs, and in principle, any flows can be transport flows.

Currently in `flows.jl`, we have 

```
lower_bound(row) =
        if row.from_asset_type in ("producer", "conversion", "storage") ||
           row.to_asset_type in ("conversion", "storage")
            0.0
        else
            -Inf
        end
```

Instead we should have 
```
lower_bound(row) =
        if row.is_transport
            -Inf
        else
            0.0
        end
```
This means we only determine the signs of the flows based on whether or not they are transport flows, instead of looking at the connecting assets.

### Related Issues
- #1162

datejada:
@gnawin it depends on what we want to allow. The current implementation doesn't allow that charging or discharging flows of storage to being negative. But, in the proposal, if someone defines ‘is_transport=true’ in a flow that is defined for charging an storage asset, then it means that we allow that flow to change and instead of charging it can also discharge using that flow (which will be strange and maybe wrong by looking at the storage balance constraints).

So, that will be my counterargument to your proposal. But, there might be some cases where we do want that, or we validate that you don't do that in your inputs.

What do you think? Am I missing something? Is it related to the new constraint we commented that the sum of outputs must be greater or equal to zero? Because then it might be covered and we can simplify as you are proposing.

gnawin:
After discussion with @datejada, we conclude that this issue should be fixed together with #1159. Fixing both will allow more possible definitions of the transport flows, while ensuring everything is correct.

PR #1172, opened 2025-04-25
Add low bound to the sum of flows for transport flows

### LINKED ISSUE #1159: Lower bound for the capacity constraints
#### Problem
Consider an example for a producer connected to two other assets. In Tulipa, we would connect them directly, 

```mermaid
flowchart LR
    A["Producer (Production)"] -- "Flow 1" --> C["Another asset"]
    A["Producer (Production)"] -- "Flow 2" --> D["Another asset"]
```
For the case when the producer just sends flows to other assets (note the arrows of the flows), we have
$Flow1 \ge 0$
$Flow2 \ge 0$
This is fine as we cover this.

However, Flow 1 or Flow 2 can be transport flows, then their signs become free (note the arrows of the flows).
```mermaid
flowchart LR
    A["Producer (Production)"] <-- "Flow 1" --> C["Another asset"]
    A["Producer (Production)"] <-- "Flow 2" --> D["Another asset"]
```

In non-Tulipa way, we are essentially modeling this case:
```mermaid
flowchart LR
    A["Producer"] -- "Production" --> B["Node"]
    B <-- "Flow 1" --> C["Another asset"]
    B <-- "Flow 2" --> D["Another asset"]
``` 
The node balance is:
$Production = Flow 1 + Flow 2$

And both sides have to be non-negative:   
$Production \ge 0 $
$Flow 1 + Flow 2 \ge 0$

#### Proposal
Currently in Tulipa, we do not have this enforcement $Flow 1 + Flow 2 \ge 0$ yet. This lower bound should apply to producers, storage, and conversion, if at least one of the flows is transport flows
- sum of outgoing flows for producer
- sum of incoming flows, sum of outgoing flows for storage
- sum of incoming flows, sum of outgoing flows for conversion

#### Implementation
Add lower bound to the capacity constraints, under the condition that at least one of the flows is transport flows. We may need a tag somewhere for the incoming/outgoing flow expression, and if that tag is true, we apply the lower bound.

### Can you help?

Yes

gnawin:
#### Model
- Enforce free signs only for transport flows, the rest flows will have a low bound of 0. 
- Add a lower bound of 0 for the sum of outgoing flows, if one of them is a transport flow, for producers, conversions, and storages. Exclude assets with UC because the bound is already applied by constraint `min_output_flow_with_unit_commitment`.
- Add a lower bound of 0 for the sum of incoming flows, if one of them is a transport flow, for conversions, and storages. Since UC does not have bounds for incoming flows, so assets with UC are not excluded.
#### Documentation
- Update concepts.
- Update mathematical formulation.

## Related issues

Closes #1159, #1168 

## Checklist

- [x] Tests are passing


gnawin:
For some reason, codecov complains. However, when I run `generate_coverage()` locally, there is no difference between main and this branch - all % match. Codecov says there is less coverage for `ramping-and-unit-commitment.jl`, but it is 100%.

Here is the report for main:
main

Here is the report for this branch:
branch

docs/src/30-concepts.md
```diff
@@ -52,12 +52,23 @@ By implementing this approach, we can reduce the number of variables and constra
 
 This example of a `phs` and a `wind` asset is useful for illustrating the advantages of this modeling approach and will be reused in the following sections. However, please keep in mind that there are other applications of hybrid configurations, such as battery-solar, hydro-solar, and electrolyzer-wind.
 
+### Additional explanation on transport flows
+
+In _TulipaEnergyModel.jl_, flows can be both input/output flows and transport flows (i.e., where you model a transmission asset). In the above illustrative example, `wind-balance` is an energy output flow, representing the production from `wind` directly sent to `balance`. In this case, the flow has a direction from `wind` to `balance`. However, in principle, any flow can be a transport flow and by definition in _TulipaEnergyModel.jl_, transport flows are bidirectional.
```
datejada, line 57: Some changes in the writing:

```suggestion
In _TulipaEnergyModel.jl_, flows can be both input/output flows and transport flows (i.e., where you model a transmission asset). In the above illustrative example, `wind-balance` is an energy output flow, representing the production from `wind` directly sent to `balance`. In this case, the flow has a direction from `wind` to `balance`. However, in principle, any flow can be a transport flow, and by definition, in _TulipaEnergyModel.jl_, transport flows are bidirectional.

Let's zoom in on the `phs-wind-balance` triangle and see what happens in the figure below. To the left, all three flows have transformed into transport flows, and they are now bidirectional.. Flows can go towards `wind`, which can be counter-intuitive: what does it mean to have a producer receiving energy? Translating into a standard method may help the thinking, we are essentially modeling the case on the right. The producer cannot receive energy, but transport flows can pass through an extra node with a unidirectional flow from the producer. We reduce this unnecessary node, but the modeling problem is not changed.
```

docs/src/30-concepts.md
```diff
@@ -52,12 +52,23 @@ By implementing this approach, we can reduce the number of variables and constra
...
+
+Let's zoom in to the `phs-wind-balance` triangle and see what happens in the below figure. On the left, all three flows become transport flows, and they are now bidiretional. Flows can go towards `wind`, which can be counter-intuitive: what does it mean to have a producer receiving energy? Translating into a standard method may help the thinking, we are essentially modeling the case on the right. Producer cannot receive energy, but transport flows can pass through an extra node with a unidirectional flow from the producer. We reduce this unnecessary node, but the modeling problem is not changed.
```
datejada, line 59: some minor changes:

```suggestion
In _TulipaEnergyModel.jl_, flows can be both input/output flows and transport flows (i.e., where you model a transmission asset). In the above illustrative example, `wind-balance` is an energy output flow, representing the production from `wind` directly sent to `balance`. In this case, the flow has a direction from `wind` to `balance`. However, in principle, any flow can be a transport flow, and by definition, in _TulipaEnergyModel.jl_, transport flows are bidirectional.

Let's zoom in on the `phs-wind-balance` triangle and see what happens in the figure below. To the left, all three flows have transformed into transport flows, and they are now bidirectional.. Flows can go towards `wind`, which can be counter-intuitive: what does it mean to have a producer receiving energy? Translating into a standard method may help the thinking, we are essentially modeling the case on the right. The producer cannot receive energy, but transport flows can pass through an extra node with a unidirectional flow from the producer. We reduce this unnecessary node, but the modeling problem is not changed.
```

docs/src/30-concepts.md
```diff
@@ -52,12 +52,23 @@ By implementing this approach, we can reduce the number of variables and constra
...
+!!! warning "Be careful with the definition of flows"
+    By having transport flows, we now model a different problem, where the charging/discharging of `phs` can be either `wind-phs-balance` or `balance-phs-wind`.
+
 ## [Flexible Time Resolution](@id flex-time-res)
 
 One of the core features of _TulipaEnergyModel.jl_ is that it can handle different time resolutions on the assets and the flows. Typically, the time resolution in an energy model is hourly, like in the following figure where we have a 6-hour energy system:
 
 ![Hourly Time Resolution](./figs/variable-time-resolution-1.png)
-
+]
```
datejada, line 71: do we need this `]`?
```suggestion
```

src/constraints/transport.jl
```diff
@@ -89,6 +90,48 @@ function add_transport_constraints!(
...
+                    model,
+                    outgoing_flow ≥ 0,
+                    base_name = "$cons_name[$(row.asset),$(row.year),$(row.rep_period),$(row.time_block_start):$(row.time_block_end)]"
+                ) for (row, outgoing_flow) in
+                zip(constraints[table_name].indices, constraints[table_name].expressions[:outgoing])
+            ],
+        )
+    end
+
+    # - Minmum input flows limit if any of the flows is transport flow
```
datejada, line 114:
```suggestion
    # - Minimum input flows limit if any of the flows is transport flow
```

src/constraints/transport.jl
```diff
@@ -89,6 +90,48 @@ function add_transport_constraints!(
         )
     end
 
+    # - Minmum output flows limit if any of the flows is transport flow
```
datejada, line 93:
```suggestion
    # - Minimum output flows limit if any of the flows is transport flow
```

src/variables/flows.jl
```diff
@@ -35,13 +34,11 @@ function _create_flow_table(connection)
...
+            flow.is_transport,
         FROM var_flow
-        LEFT JOIN asset AS from_asset
-            ON var_flow.from_asset = from_asset.asset
-        LEFT JOIN asset AS to_asset
-            ON var_flow.to_asset = to_asset.asset
+        LEFT JOIN flow
+            ON flow.from_asset = var_flow.from_asset
+            AND flow.to_asset = var_flow.to_asset
         ORDER BY var_flow.id
```
datejada, line 42: Are these two equivalent? I see that you need a flow for `is_transport`, but have you double-checked that you get the same resulting table? With different inputs?

datejada (commented):
@gnawin Thanks for the changes, they look good. Minor comments on the text and two main comments:

1. One for now: I see that you are not testing the new constraints, aren't you? I mean, you don't change any data in the case studies that are actually generating these constraints. I am saying this because I was expecting changes in the MPS files of one of the case studies (Norse, perhaps?). Any particular reason that I might be missing? otherwise I would suggest testing the new constraints by modifying the Norse.
2. long-term, so not in this PR (for future record): some constraints might be just bounds, if the asset only has one input or output. So, if we identify those cases beforehand, we can avoid creating constraints that are actually bounds. As I said, for the future 😃

gnawin:
Closed in favor of #1178

src/variables/flows.jl
gnawin: These are not equivalent. The removed part is simply not needed anymore as now we do not check the types of assets.

docs/src/30-concepts.md
gnawin: Thank you. I will address these in #1178.

PR #1178, opened 2025-04-28
Add low bound to the sum of flows for transport flows

### LINKED ISSUE #1159: Lower bound for the capacity constraints

gnawin:
#### Model
- Enforce free signs only for transport flows, the rest flows will have a low bound of 0. 
- Add a lower bound of 0 for the sum of outgoing flows, if one of them is a transport flow, for producers, conversions, and storages. Exclude assets with UC because the bound is already applied by constraint `min_output_flow_with_unit_commitment`.
- Add a lower bound of 0 for the sum of incoming flows, if one of them is a transport flow, for conversions, and storages. Since UC does not have bounds for incoming flows, so assets with UC are not excluded.
#### Documentation
- Update concepts.
- Update mathematical formulation.
## Related issues

Closes #1159, #1168 

## Checklist

- [ ] Tests are passing


gnawin:
@datejada The MPS file has changed, but somehow the change was not detected. I'll update the MPS file once you have a look at the Github actions. Thanks.

datejada (commented):
@gnawin Thanks for the PR, one last thing, I think we commented on it before but I forgot the reason. Can we make the codecov pass? or there is something why we are not testing those cases anymore.

gnawin:
@datejada the Github code coverage give the same results in the original PR (https://github.com/TulipaEnergy/TulipaEnergyModel.jl/pull/1172#issuecomment-2835150310), while locally it was okay. I have only changed the documentation afterwards.

Now when I open code coverage, it shows one line is missing for `flows.jl`:
image
As you can see that line is already missed (for unknown reasons) in main.

I also ran **local** code coverage, here are the results:
**Main**
image
**This branch**
image

So nothing is changing in `flows.jl`, and that line is actually covered (as it is the lower bound of non-transport flow variables).

datejada (approved):
Thank you for the explanation of the codecov. No more comments from y side  😃

PR #1179, opened 2025-04-28
Add parameters for dc-opf feature

### LINKED ISSUE #1177: Define new parameters
Baesd on https://github.com/TulipaEnergy/TulipaEnergyModel.jl/issues/989#issuecomment-2790181008, we need to define:
 
- $p^{\text{power system base}}$: This parameter can go in the ModelParameters structure and is defined in the "model-parameters.toml" (or .toml file with the model parameters). It is similar to the discount rate and discount year. The default value should be 100MVA (the typical value for power systems).
- $p^{\text{reactance}}_{f,y}$: This parameter should go in `flow-milestone.csv` since it only depends on the milestone year like the variable cost.
- New method for DC-OPF

gnawin:
- Add `dc_opf` and `reactance` to `flow-milestone.csv`
- Add `power_system_base` to `model-parameters.jl`

## Related issues

Closes #1177 

## Checklist

- [x] Tests are passing


gnawin:
@datejada could you look at the description (including the default) of `reactance`? I'm not a electrical engineer so not sure if it is correct. Thanks!

datejada (commented):
@gnawin, thanks for the PR. I like the latest change, the DC-OPF should depend on the year to be more flexible in the modelling options (e.g., 2040 with DC-OPF and then 2050 without it...for the same transport flow).

The descriptions are fine, we will give you the title of electrical engineer `honoris causa` 😃 

My main comment is that you added the model parameter that is in the tests, but didn't test it here:

https://github.com/TulipaEnergy/TulipaEnergyModel.jl/blob/main/test/test-model-parameters.jl

Could you add it? This is to check that we read the parameter correctly.

Thanks

datejada:
Another comment, maybe for discussion first and then maybe (or not) create another issue:

Adding the columns to all the case studies is annoying, but we now have the function to add the defaults. So, we can simplify the input CSV of the case studies and have only the data we need for the case studies. In the tests, we can use the function to populate the defaults to run the tests.

That will avoid the hassle of changing the CSVS every time.

What do you think? @gnawin @abelsiqueira

gnawin:
> Another comment, maybe for discussion first and then maybe (or not) create another issue:
> 
> Adding the columns to all the case studies is annoying, but we now have the function to add the defaults. So, we can simplify the input CSV of the case studies and have only the data we need for the case studies. In the tests, we can use the function to populate the defaults to run the tests.
> 
> That will avoid the hassle of changing the CSVS every time.
> 
> What do you think? @gnawin @abelsiqueira

I love this idea :)

gnawin:
@datejada thanks you for the heads-up! I added the test. `power_system_base` has a default values, so it is treat the same as `discount_rate`, which are now tested in `Basic usage`.

datejada (approved):
Great! It looks good. I love that we can do small PRs ❤️ it is easier to check and review. Thanks!

datejada:
> > Another comment, maybe for discussion first and then maybe (or not) create another issue:
> > Adding the columns to all the case studies is annoying, but we now have the function to add the defaults. So, we can simplify the input CSV of the case studies and have only the data we need for the case studies. In the tests, we can use the function to populate the defaults to run the tests.
> > That will avoid the hassle of changing the CSVS every time.
> > What do you think? @gnawin @abelsiqueira
> 
> I love this idea :)

Let's comment it with @abelsiqueira when he comes back from holiday

PR #1183, opened 2025-04-29
Add new tables to calculate the lowest resolution for flows that have a flows relationship

### LINKED ISSUE #1182: Create a table with the lowest resolution of both flows in the flows relationship table
### Description

We need to create a table with the lowest resolution of both flows, e.g., `max(resolution of flow 1, resolution flow 2)`, in the flows_relationships table for later use in the constraints.

datejada:
This pull request refactors the `create_merged_tables!` function to simplify its implementation, introduces a new table for flow relationships, and expands test coverage for data preparation.

### Refactoring and Simplification:
* Replaced inline SQL queries in `create_merged_tables!` with an external SQL file (`create-merged-tables.sql`) for better maintainability and readability. (`src/data-preparation.jl`, [[1]](diffhunk://#diff-da8a343a3066315da2c08dd913f31ea303b0d8fc2f40ef42efc055107cac889eL377-R377) [[2]](diffhunk://#diff-da8a343a3066315da2c08dd913f31ea303b0d8fc2f40ef42efc055107cac889eR386-R392) [[3]](diffhunk://#diff-a9f68e30d0a2783c2b4beec133bb8e5d0c53b05830e65026492592255d7a2434R1-R97)

### New Feature: Flow Relationships:
* Added a new temporary table, `merged_flows_relationship`, to handle flow relationships, including creating and populating a `flows_relationship` column in the `flows_relationships` table. (`src/sql/create-merged-tables.sql`, [src/sql/create-merged-tables.sqlR1-R97](diffhunk://#diff-a9f68e30d0a2783c2b4beec133bb8e5d0c53b05830e65026492592255d7a2434R1-R97))
* Updated the `create_lowest_resolution_table!` function to include the new `merged_flows_relationship` table in its processing. (`src/data-preparation.jl`, [src/data-preparation.jlL497-R451](diffhunk://#diff-da8a343a3066315da2c08dd913f31ea303b0d8fc2f40ef42efc055107cac889eL497-R451))

### Expanded Test Coverage:
* Introduced a comprehensive test suite in `test/test-data-preparation.jl` to validate the correctness of the merged tables, including the new `merged_flows_relationship` table, and the lowest and highest resolution tables. (`test/test-data-preparation.jl`, [test/test-data-preparation.jlR1-R234](diffhunk://#diff-521b7acd498e563458453413cd81c9122e2360dbb4517b6a972cd71a4d5243e7R1-R234))
* Added a helper function `_test_rows_exist` to verify the presence of specific rows in test tables. (`test/utils.jl`, [test/utils.jlR20-R33](diffhunk://#diff-3483c521f73b08fdb6b00f014614cc0d69f87ea7b098a24aff838cbdc812704dR20-R33))

## Related issues

Closes #1182

## Checklist

- [x] Tests are passing

### DIFF
--- a/src/sql/create-merged-tables.sql
+++ b/src/sql/create-merged-tables.sql
@@ -61,22 +61,15 @@ from
     merged_all_flows
 ;
 
--- merged table for flows relationships
--- 1. create a new column to store the flows relationship
-alter table flows_relationships
-    add flows_relationship STRING
-;
--- 2. add the values to the new flows relationship column
-update flows_relationships
-set flows_relationship = CONCAT(flow_1_from_asset, '_',
-                                flow_1_to_asset, '_',
-                                flow_2_from_asset, '_',
-                                flow_2_to_asset)
-;
--- 3. get the resolution of the flows that are in the relationship
+-- merged table for flows relationships:
+-- 1. Define the asset as the combination of the two flows (i.e., CONCAT)
+-- 2. Get the resolution of the flows that are in the relationship (i.e., JOIN)
 create or replace temp table merged_flows_relationship as
     select distinct
-        fr.flows_relationship AS asset,
+        CONCAT(fr.flow_1_from_asset, '_',
+               fr.flow_1_to_asset, '_',
+               fr.flow_2_from_asset, '_',
+               fr.flow_2_to_asset) as asset,
         ftrrp.year,
         ftrrp.rep_period,
         ftrrp.time_block_start,
--- a/src/sql/create-merged-tables.sql
+++ b/src/sql/create-merged-tables.sql
@@ -1,0 +1,115 @@
+-- incoming flows
+create or replace temp table merged_in_flows as
+select
+    distinct to_asset as asset,
+    year,
+    rep_period,
+    time_block_start,
+    time_block_end
+from
+    flow_time_resolution_rep_period
+;
+
+-- outgoing flows
+create or replace temp table merged_out_flows as
+select
+    distinct from_asset as asset,
+    year,
+    rep_period,
+    time_block_start,
+    time_block_end
+from
+    flow_time_resolution_rep_period
+;
+
+-- union of all assets and outgoing flows
+create or replace temp table merged_assets_and_out_flows as
+select
+    distinct asset,
+    year,
+    rep_period,
+    time_block_start,
+    time_block_end
+from
+    asset_time_resolution_rep_period
+union
+from
+    merged_out_flows
+;
+
+-- union of all incoming and outgoing flows
+create or replace temp table merged_all_flows as
+from
+    merged_in_flows
+union
+from
+    merged_out_flows
+;
+
+-- union of all assets, and incoming and outgoing flows
+create or replace temp table merged_all as
+select
+    distinct asset,
+    year,
+    rep_period,
+    time_block_start,
+    time_block_end
+from
+    asset_time_resolution_rep_period
+union
+from
+    merged_all_flows
+;
+
+-- merged table for flows relationships
+create or replace temp table merged_flows_relationship as with matched_flows as (
+    select
+        fr.flow_1_from_asset,
+        fr.flow_1_to_asset,
+        fr.flow_2_from_asset,
+        fr.flow_2_to_asset,
+        ftrrp.year,
+        ftrrp.rep_period,
+        ftrrp.time_block_start,
+        ftrrp.time_block_end,
+    from
+        flow_time_resolution_rep_period as ftrrp
+        join flows_relationships as fr on (
+            (
+                ftrrp.from_asset = fr.flow_1_from_asset
+                and ftrrp.to_asset = fr.flow_1_to_asset
+            )
+            or (
+                ftrrp.from_asset = fr.flow_2_from_asset
+                and ftrrp.to_asset = fr.flow_2_to_asset
+            )
+        )
+        and ftrrp.year = fr.milestone_year
+)
+select
+    distinct case
+        -- if flow_1_to_asset matches flow_2
+        when matched_flows.flow_1_to_asset = matched_flows.flow_2_from_asset
+        or matched_flows.flow_1_to_asset = matched_flows.flow_2_to_asset
+        then matched_flows.flow_1_to_asset
+        -- if flow_1_from_asset matches flow_2
+        when matched_flows.flow_1_from_asset = matched_flows.flow_2_from_asset
+        or matched_flows.flow_1_from_asset = matched_flows.flow_2_to_asset
+        then matched_flows.flow_1_from_asset
+        -- if flow_2_to_asset matches flow_1
+        when matched_flows.flow_2_to_asset = matched_flows.flow_1_from_asset
+        or matched_flows.flow_2_to_asset = matched_flows.flow_1_to_asset
+        then matched_flows.flow_2_to_asset
+        -- if flow_2_from_asset matches flow_1
+        when matched_flows.flow_2_from_asset = matched_flows.flow_1_from_asset
+        or matched_flows.flow_2_from_asset = matched_flows.flow_1_to_asset
+        then matched_flows.flow_2_from_asset
+        else null
+    end as asset,
+    year,
+    rep_period,
+    time_block_start,
+    time_block_end
+from
+    matched_flows
+;
--- a/src/sql/create-merged-tables.sql
+++ b/src/sql/create-merged-tables.sql
@@ -62,12 +62,21 @@ from
 ;
 
 -- merged table for flows relationships
-create or replace temp table merged_flows_relationship as with matched_flows as (
-    select
-        fr.flow_1_from_asset,
-        fr.flow_1_to_asset,
-        fr.flow_2_from_asset,
-        fr.flow_2_to_asset,
+-- 1. create a new column to store the flows relationship
+alter table flows_relationships
+    add flows_relationship STRING
+;
+-- 2. add the values to the new flows relationship column
+update flows_relationships
+set flows_relationship = CONCAT(flow_1_from_asset, '_',
+                                flow_1_to_asset, '_',
+                                flow_2_from_asset, '_',
+                                flow_2_to_asset)
+;
+-- 3. get the resolution of the flows that are in the relationship
+create or replace temp table merged_flows_relationship as
+    select distinct
+        fr.flows_relationship AS asset,
         ftrrp.year,
         ftrrp.rep_period,
         ftrrp.time_block_start,
@@ -85,31 +94,4 @@ create or replace temp table merged_flows_relationship as with matched_flows as 
             )
         )
         and ftrrp.year = fr.milestone_year
-)
-select
-    distinct case
-        -- if flow_1_to_asset matches flow_2
-        when matched_flows.flow_1_to_asset = matched_flows.flow_2_from_asset
-        or matched_flows.flow_1_to_asset = matched_flows.flow_2_to_asset
-        then matched_flows.flow_1_to_asset
-        -- if flow_1_from_asset matches flow_2
-        when matched_flows.flow_1_from_asset = matched_flows.flow_2_from_asset
-        or matched_flows.flow_1_from_asset = matched_flows.flow_2_to_asset
-        then matched_flows.flow_1_from_asset
-        -- if flow_2_to_asset matches flow_1
-        when matched_flows.flow_2_to_asset = matched_flows.flow_1_from_asset
-        or matched_flows.flow_2_to_asset = matched_flows.flow_1_to_asset
-        then matched_flows.flow_2_to_asset
-        -- if flow_2_from_asset matches flow_1
-        when matched_flows.flow_2_from_asset = matched_flows.flow_1_from_asset
-        or matched_flows.flow_2_from_asset = matched_flows.flow_1_to_asset
-        then matched_flows.flow_2_from_asset
-        else null
-    end as asset,
-    year,
-    rep_period,
-    time_block_start,
-    time_block_end
-from
-    matched_flows
 ;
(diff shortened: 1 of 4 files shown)


datejada:
@suvayu Inspired by our last conversation, my next step in the new feature is to be able to test the main pieces of the code without doing a full integration test. The flexible temporal resolution is key for Tulipa, so I have created a separate test to test independently the three main functions that calculate these tables. I have also added my new flexible temporal resolution function and tested it for the mock-up data.

Please let me know your comments and suggestions.

@gnawin, also, from the developers' maintainability perspective, please let me know if the tests make sense or if you would like to add/delete something else.

test/test-data-preparation.jl
```diff
@@ -0,0 +1,234 @@
...
+                   ('asset', 'output_1', 'input_1', 'asset', 2025),
+                   ('asset', 'output_1', 'asset', 'output_2', 2025),
+        ",
+    )
+    DBInterface.execute(
+        connection,
+        "INSERT INTO rep_periods_data
+            VALUES (5, 1, 1.0, 2025)
+        ",
+    )
```
gnawin, line 92: I can follow what you are doing here, i.e., first create an empty table and insert values into that. However, it seems a bit cumbersome to do this way (at least to me). For the same purpose, I did it differently in `data-validation.jl`, followed the example of made by Abel: first create a `df`, and then put the `df` directly into the `connection` using `register_data_frame`, see example [here](https://github.com/TulipaEnergy/TulipaEnergyModel.jl/blob/94b7c96fa4b27cb13beaf01fc89dce4b05e92f62/test/test-data-validation.jl#L41). This is easier because we don't need to create an empty table first, and if we all use the same way to create example data, it can be more consistent.

test/test-data-preparation.jl
```diff
@@ -0,0 +1,234 @@
...
+             flow_1_to_asset STRING,
+             flow_2_from_asset STRING,
+             flow_2_to_asset STRING,
+             milestone_year INTEGER)
+        ",
+    )
+
+    DBInterface.execute(
+        connection,
+        "CREATE TABLE rep_periods_data
```
gnawin, line 52: Do you need this table `rep_periods_data` (because I don't see it directly in this PR)?

test/test-data-preparation.jl
```diff
@@ -0,0 +1,234 @@
...
+
+        # Verify the results in the merged tables
+        merged_in_flows = TulipaIO.get_table(connection, "merged_in_flows")
+        rows_to_test = [
+            ("asset", 2025, 1, 1, 1),
+            ("asset", 2025, 1, 2, 5),
+            ("asset", 2025, 1, 1, 2),
+            ("asset", 2025, 1, 3, 5),
+        ]
+        @test _test_rows_exist(rows_to_test, merged_in_flows) |> all
```
gnawin, line 105: This function tests if the row exists or not. As I understood, we should check whether or not they have a exact match.
Maybe something like this:
```
expected_data = DataFrame(
:asset => repeat(["asset"], 4),
:year =>repeat([2025], 4),
:rep_period =>repeat([1], 4)
:time_block_start => [1, 2, 1, 3],
:time_block_end => [1, 5, 2, 5],
)
sorted_expected_data = sort(expected_data, [:asset, :year, :rep_period, :time_block_start, :time_block_end]) 
sorted_actual_data = sort(merged_in_flows, [:asset, :year, :rep_period, :time_block_start, :time_block_end]) 
@test sorted_expected_data == sorted_actual_data
```

test/test-data-preparation.jl
datejada: you need it to test the `create_highest_resolution_table!`, which is not part of my PR, but as @suvayu says, if you can add/fix something in your PR, just do it 😝

test/test-data-preparation.jl
```diff
@@ -0,0 +1,234 @@
...
+            ("input_1_asset_input_2_asset", 2025, 1, 1, 2),
+            ("input_1_asset_input_2_asset", 2025, 1, 3, 5),
+        ]
+        @test _test_rows_exist(rows_to_test, merged_flows_relationship) |> all
+    end
+
+    @testset "Test create_lowest_resolution_table!" begin
+        TulipaEnergyModel.create_lowest_resolution_table!(connection)
+
+        # Verify the results in the lowest resolution tables
```
gnawin, line 178: Nice that you test also other `lower_resolution` tables. One comment is still consistency: maybe use dataframes and compare them directly (or in a convenience function).
In addition, it might be a good idea to use the same test data, for `highest_resolution`. Coincidentally for the dc-opf, I'm touching the highest resolution tables, maybe I can test `highest_resolution` after this PR. The only thing I'm not sure for now is whether your test data is good enough for `highest_resolution` or not. I'll check later, but if not, I can create test data for `highest_resolution`.

gnawin: Sorry I just found you also tested `highest_resolution`, forget what I just said.

gnawin (commented):
@datejada thank you! I think you tested nicely, and you cover more than the purpose of MIMO because you tested all `lowest_resolution` and `highest_resolution` tables, great! I have some minor suggestions regarding the testing, to improve some consistency with other tests.

test/test-data-preparation.jl
datejada: The functions create a bunch of rows that I don't want to test (Apriori). For example, it also creates rows for the "assets" called "input_1", "input_2", "output_1", and "output_2"; so I wanted to focus on one asset. Because I know in the end what I want as the lowest/highest resolution for all the cases when I look at the asset. Testing the full output table makes the test less understandable (in my opinion).

I may need to add that to the initial comment on the tests. Also, I can add a test on the total rows created, then I am testing that the dataframe has the right rows but focussing on one asset, because it should be enough to know that the function works as expected.

Anyway, I am pondering on testing the full detail and the minimum to guarantee we are getting what we want.

Let me know your opinion and we can maybe find a halfway solution, unless there is a strong argument to test the whole dataframe.

@suvayu any suggestions?

Thanks!

test/test-data-preparation.jl
datejada: But, the underlying question remains. I would suggest that if you need a specific test. It may be better to add a new one, there might be things you want to test that does not fit in this data for your development.

test/test-data-preparation.jl
gnawin: Sorry I did not fully understand. You first create a df, then use `DuckDB.register_data_frame(connection, flows_relationships, df)` to create the table `flows_relationships`. To me this is about creating a test dataset, but how does the ALTER part play a role here?

datejada:
> @datejada thank you! I think you tested nicely, and you cover more than the purpose of MIMO because you tested all `lowest_resolution` and `highest_resolution` tables, great! I have some minor suggestions regarding the testing, to improve some consistency with other tests.

Thanks! I like the suggestions, let's see if we can make it work by registering the dataframes. So far, I failed when I tried as I commented here: https://github.com/TulipaEnergy/TulipaEnergyModel.jl/pull/1183#discussion_r2071634061

test/test-data-preparation.jl
gnawin: I see. Then maybe do some filtering to focus on the part you want to test, and some comments on those? My point was more: in my opinion, you want to test a exact match between the expected data and the actual data, after filtering in this case. 

The function that uses `any` misleads me because I thought an exact match should be tested, otherwise you may see more rows are generated for the same asset, and the test would still pass.

test/test-data-preparation.jl
datejada: Sorry, I explained myself poorly before. Let me try again:

1. I first tried your proposal using the function "register_data_frame"
2. I can register all the tables using that option, no problem there
3. The problem appeared when I tested the creation of the merged tables. I introduced one SQL query that ALTERS one of the input tables (because I need a unique identifier for each row in the flows relationship table).
4. I investigated the error, and it said that the function "register_data_frame" does not actually create a table; it is a VIEW in the connection. So, when you use that function, you can´t alter the tables you view in the connection; you only use their information.
5. The option to create my test data is as it is in the PR.
6. The ugly part is that Duckdb does not allow creating and adding values in the same query. So, you must make the table first and then add the values. That is why it looks like too much code. (We could wrap it up in an auxiliary function to make the tests shorter).

I hope this time I have explained myself better 😁

test/test-data-preparation.jl
datejada: Okay, I see your point. Let's discuss it with @suvayu (maybe Abel, if he is available next week) and see what the best approach is here. Thanks! At least, this test makes clear how we get the resolutions in the model and what to expect from the input data with flexible temporal resolutions.

test/test-data-preparation.jl
gnawin: Google says if you convert the view to a real table then it works. I tested, indeed this way you can ALTER. :)

```
using DuckDB, DataFrames

con = DBInterface.connect(DuckDB.DB) 
df = DataFrame(a = 1:3, b = ["x", "y", "z"])
DuckDB.register_data_frame(con,  df, "my_df_view")
DuckDB.execute(con, "CREATE TABLE my_df AS SELECT * FROM my_df_view")
DuckDB.execute(con, "ALTER TABLE my_df ADD COLUMN c INT")
```

datejada: It works! Thanks 😄 I tried that but failed using the same name and CREATE OR REPLACE statement, so the trick is using two names, one for the view and the other for the table. Great! I have changed the code to use data frames 😉 Thanks!

datejada:
@gnawin, thanks for the comments and discussion. I have made two main changes:

- Create the mock case study using Dataframes
- The test will be added to the size of the tables to ensure they have the expected number of rows and columns.

The second change is to balance between:
 a) comparing the whole dataframes one by one, or
 b) focusing on one asset + table size as an extra check that all went okay

Approach a) is 100% safe, but it is not easy to understand why you have all the rows. Approach b) is not 100% safe, but it makes the test readable by focusing on only what is happening in one asset in all the cases + as a second check, that the output tables are the correct size (one would expect these checks to be enough).

I am still trying to balance both approaches, but let's discuss with @suvayu and @abelsiqueira the best way to proceed.

abelsiqueira:
Again, I don't have the whole context. If needed, ping me again and I'll review the whole PR.
I think a clean enough way, not too verbose, that tests fully the expected assets, would be 

```julia
expected_df = DataFrame(
... # I think you can pretty much use the same array of tuples
)
computed_df = DuckDB.query("SELECT .. WHERE asset = 'asset' SORT BY ...") |> DataFrame
@test dataframes_are_equal(expected_df, computed_df)
```

If all `computed_df` follow the same `WHERE` and `SORT BY`, you can create another function

```Julia
expected_df = ...
@test expected_df_matches_filtered_table(expected_df, target_table)
```

src/sql/create-merged-tables.sql
```diff
@@ -0,0 +1,97 @@
...
+update flows_relationships
+set flows_relationship = CONCAT(flow_1_from_asset, '_',
+                                flow_1_to_asset, '_',
+                                flow_2_from_asset, '_',
+                                flow_2_to_asset)
+;
+-- 3. get the resolution of the flows that are in the relationship
+create or replace temp table merged_flows_relationship as
+    select distinct
+        fr.flows_relationship AS asset,
```
datejada, line 79: use here CONCAT instead

test/test-data-preparation.jl
```diff
@@ -0,0 +1,241 @@
...
+            ("asset", 2025, 1, 4, 5),
+            ("asset", 2025, 1, 1, 4),
+            ("asset", 2025, 1, 5, 5),
+        ]
+        @test _test_rows_exist(rows_to_test, merged_out_flows) |> all
+        @test DataFrames.nrow(merged_out_flows) == 8
+        @test DataFrames.ncol(merged_out_flows) == 5
+
+        merged_assets_and_out_flows = TulipaIO.get_table(connection, "merged_assets_and_out_flows")
+        rows_to_test = [
```
datejada, line 102: change to a df

test/test-data-preparation.jl
```diff
@@ -0,0 +1,241 @@
...
+            ("asset", 2025, 1, 1, 3),
+            ("asset", 2025, 1, 4, 5),
+            ("asset", 2025, 1, 1, 4),
+            ("asset", 2025, 1, 5, 5),
+            ("asset", 2025, 1, 1, 5),
+        ]
+        @test _test_rows_exist(rows_to_test, merged_assets_and_out_flows) |> all
```
datejada, line 109: compare both dataframes

test/test-data-preparation.jl
datejada: DataFrame(rows, ["asset", "year", "rp", "begin", "end"])

datejada:
@suvayu @abelsiqueira @gnawin I have included all the comments as we discussed. Please let me know if you have further suggestions.

abelsiqueira (approved):
Thanks, LGTM.
💀 ⭐

gnawin (approved):
Thanks. We have discussed this PR in person. For the future, it would be nice to generalize the function for creating highest/lowest resolutions, maybe changing the name from asset to something else is already good enough.

test/test-data-preparation.jl
```diff
@@ -0,0 +1,238 @@
...
+    # Create mock tables for testing using register_data_frame
+    table_rows = [
+        ("input_1", "death_star", 2025, 1, 1, 1),
+        ("input_1", "death_star", 2025, 1, 2, 5),
+        ("input_2", "death_star", 2025, 1, 1, 2),
+        ("input_2", "death_star", 2025, 1, 3, 5),
+        ("death_star", "output_1", 2025, 1, 1, 3),
+        ("death_star", "output_1", 2025, 1, 4, 5),
+        ("death_star", "output_2", 2025, 1, 1, 4),
+        ("death_star", "output_2", 2025, 1, 5, 5),
```
suvayu, line 26: I like the example ;)

suvayu (approved):
LGTM :+1:

PR #1185, opened 2025-04-29
Add electricity angle variables

### LINKED ISSUE #1181: Add variables for dc-opf

gnawin:
Adding electricity angle variables based on asset resolution.

The variable is created if there is a connecting flow (incoming OR outgoing) that is transport and has dc_opf method.

## Related issues

Closes #1181

## Checklist

- [x] Tests are passing

### DIFF
--- a/src/sql/create-variables.sql
+++ b/src/sql/create-variables.sql
@@ -45,6 +45,30 @@ drop sequence id
 create sequence id start 1
 ;
 
+create table var_electricity_angle as
+select
+    nextval('id') as id,
+    atr.asset,
+    atr.year,
+    atr.rep_period,
+    atr.time_block_start,
+    atr.time_block_end,
+from
+    asset_time_resolution_rep_period as atr
+    left join flow on flow.from_asset = atr.asset
+    or flow.to_asset = atr.asset
+where
+    flow.is_transport
+group by
+    atr.asset, atr.year, atr.rep_period, atr.time_block_start, atr.time_block_end
+;
+
+drop sequence id
+;
+
+create sequence id start 1
+;
+
 create table var_is_charging as
 select
     nextval('id') as id,
--- a/src/sql/create-variables.sql
+++ b/src/sql/create-variables.sql
@@ -54,14 +54,25 @@ select
     atr.time_block_start,
     atr.time_block_end,
 from
+    -- The angle resolution is the same as the time resolution of the asset
     asset_time_resolution_rep_period as atr
+    -- We need to check if the asset has any connecting flows that are transport
+    -- We only get the assets that have outgoing flows OR incoming flows
+    -- With the condition that the flows are transport
     left join flow on flow.from_asset = atr.asset
     or flow.to_asset = atr.asset
+    -- After that we need to also check if the flows have dc_opf method
+    -- This is by joining the flow_milestone table
+    -- Here we use AND because we need to match flow_milestone with flow
     left join flow_milestone on flow_milestone.from_asset = flow.from_asset
     and flow_milestone.to_asset = flow.to_asset
+    and flow_milestone.milestone_year = atr.year
 where
     flow.is_transport
     and flow_milestone.dc_opf
+-- We may end up with duplicates because an asset can have both incoming and outgoing flows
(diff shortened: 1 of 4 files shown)


src/sql/create-variables.sql
```diff
@@ -45,6 +45,33 @@ drop sequence id
...
+    and flow_milestone.dc_opf
+group by
+    atr.asset, atr.year, atr.rep_period, atr.time_block_start, atr.time_block_end
+;
+
+drop sequence id
+;
+
+create sequence id start 1
+;
```
datejada, line 73: The joins here are more complex than the ones for the other variables. It is what it is because we discussed it, and this is what we need. But it took me a bit to follow it and remind myself what we did. So, adding a comment at the beginning of the query explaining what we want here will help follow the query more easily. Plus, we need to comment on the criteria for the resolution in the concepts documentation (in a future PR).

src/variables/create.jl
```diff
@@ -11,6 +11,7 @@ function compute_variables_indices(connection)
             :flow,
             :units_on,
             :is_charging,
+            :electricity_angle,
```
datejada, line 14: I am unsure, but I thought the order matters. Maybe not? I mean, your queries are not in the same order as the keys. According to the queries, it should be units_on, electricity_angle, and is_charging.

Would you mind double-checking?

datejada (commented):
@gnawin thanks for the PR. A couple of minor comments. Loving reviewing small PRs 😄

src/sql/create-variables.sql
gnawin: I added more explanations. It appears verbose, but maybe it helps to make it clear to everybody (also to myself). Let me know if you think that's too much.

src/variables/create.jl
gnawin: It actually works also this way. I think because the whole sql table is first read in, so the table `var_electricity_angle` is created in the `connection`, and then the order does not mattter. 
Nevertheless I changed the order so it looks neater.

gnawin:
@datejada Thank you for the review! Making smaller PRs also is easier for me which also enables faster feedback. I also like this. 

I addressed your comments. In the meantime, I found a bug because `year` was not checked, but it is fixed. Hopefully I do not miss anything else now 😄. I'll also add more comments in my other PRs so it's easier to follow.

src/sql/create-variables.sql
```diff
@@ -45,6 +45,44 @@ drop sequence id
...
+    left join flow_milestone on flow_milestone.from_asset = flow.from_asset
+    and flow_milestone.to_asset = flow.to_asset
+    and flow_milestone.milestone_year = atr.year
+where
+    flow.is_transport
+    and flow_milestone.dc_opf
+-- We may end up with duplicates because an asset can have both incoming and outgoing flows
+-- Or it can have multiple flows
+-- GROUP BY is used to remove duplicates
+group by
```
datejada, line 76: wouldn't then better to start with `select distinct`? I am unsure if there is a performance gain, but I see from Abel's code that he uses select distinct to remove duplicates rather than grouping by. Which actually, I am also not sure what happens with the id column when you groupby these columns (I mean, do you get the sum? the avg? the first one?)

datejada (commented):
@gnawin Thanks! I like the changes and the comments. It will help us in the future since it is not a simple calculation/logic to get the resolution of this variable. One last comment, about the groupby. Please check it and then we are done 😄

src/sql/create-variables.sql
gnawin: Now I remember, `select distinct` does not work because of the `id`, so I used `group by`. Because `select distinct` will select all distinct rows, but with a unique id for each row, there will be no duplicates. However, `select` only happens after `group by`, therefore a unique `id` is given to each unique row. 

I just tested with duplicates, indeed this is the case.

gnawin:
@datejada thanks! I just improved the query, so I now actually use an aggregate function `any_value`, because technically I do not need to group by `time_block_end`.

datejada (approved):
Thanks for all the changes. LTGM

PR #1190, opened 2025-05-06
Fix CompareMPS when files differ in length

### LINKED ISSUE #1188: [Bug] MPS workflow false positive result
### Description

While developing PR #1178, we discovered that the MPS workflow gives a false positive. We know that the MPS file must change, and locally, we compare the MPS files, which are different, but the MPS workflow still says there is no problem.

One more thing: We are unsure if it should be able to run on forks or if it is only in branches in the main repo. Maybe add something to clarify this.

One minor last thing, in PR #1183 the old comment of a failed run was not cleared.

@abelsiqueira, I tried to see what was happening, but I was not able to see anything obvious that was failing. Let us know any ideas 😄  

### Package Version

0.15.1

### Julia Version

1.11.5

### Reproduction steps

Generate locally the MPS files for the main branch and the PR changes and compare them. They are different, and the MPS result is there is no changes https://github.com/TulipaEnergy/TulipaEnergyModel.jl/pull/1178#issuecomment-2835355363

### Relevant log output

```shell

```

### Operating System

_No response_

abelsiqueira:
If the files had different length, the script prints errors,
but was not returning errors. This returns the correct
exit flag.

## Related issues

Closes #1188

## Checklist

- [x] Tests are passing

### DIFF
--- a/.github/workflows/CompareMPS.yml
+++ b/.github/workflows/CompareMPS.yml
@@ -28,7 +28,7 @@ jobs:
         with:
           issue-number: ${{ github.event.pull_request.number }}
           comment-author: "github-actions[bot]"
-          body-includes: This was CompareMPS
+          body-includes: CompareMPS report
       - name: Comment on PR with CompareMPS
         if: github.event.pull_request.head.repo.full_name == github.repository
         uses: peter-evans/create-or-update-comment@v4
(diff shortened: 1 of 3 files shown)


abelsiqueira:
@datejada @gnawin, I think this fixes the issue.

gnawin (approved):
Thank you! I'll rebase the #1178, and see it makes a difference.

PR #1191, opened 2025-05-06
Always run find-comment in CompareMPS, whether it fails or not

abelsiqueira:
This ensures that looks for earlier comments even if there was a failure.

## Related issues

Complements #1188

## Checklist

- [x] Tests are passing


abelsiqueira:
@datejada @gnawin maybe this fixes the duplicate comment issue

datejada (approved):
Thanks! I will test it in my next PRs.

PR #1196, opened 2025-05-13
Update EU benchmark case study

### LINKED ISSUE #1193: Update EU case study for the benchmarks
### Description

The current case study for the benchmark is infeasible. Although this is not a problem since the case study is meant to test only the creation of the model, it would be nice to be able to solve it if needed.

datejada:
Update the EU case study used for benchmark, since the current one is infeasible, we want to have a case study here that also solves. It is not that it must, but it is a nice to have

## Related issues

Closes #1193 

## Checklist

- [x] Tests are passing


datejada:
@abelsiqueira I am unsure why it didn't print the results in the comment, but these are the results:

image


Sizewise they the new data are similar (see the results), the difference is that the old case was infeasible and this is feasible.

abelsiqueira (approved):
Thanks!

PR #1197, opened 2025-05-14
Improve the hotfix for the alignment of `units_on` and compact method

gnawin:
Improve the hotfix for the alignment of `units_on` and compact method in #1195.
These constraints are indexed already by `var_units_on` filtered by the investment methods, so it is possible to obtain the index of `units_on` directly in the original indice function.

## Related issues

N/A

## Checklist

- [x] Tests are passing


datejada (approved):
Thank you for the improve, it is also more readable in this form.

PR #1198, opened 2025-05-14
Create highest resolution table for flows and their connecting assets

### LINKED ISSUE #1187: Add highest resolution table between flows and its connecting assets
Need to rebase after #1183 because it refactors the code on generating resolution tables.

gnawin:
The goal is to obtain the highest resolution between a flow and its connecting assets.
Used the same trick as in #1183, by creating a fake asset called (`from_asset`, `_`, `to_asset`).

This is a workaround and note by doing this, I'll need to split this asset later as I need the flow names.

## Related issues

Closes #1187 

## Checklist

- [x] Tests are passing

### DIFF
--- a/src/sql/create-merged-tables.sql
+++ b/src/sql/create-merged-tables.sql
@@ -88,3 +88,36 @@ create or replace temp table merged_flows_relationship as
         )
         and ftrrp.year = fr.milestone_year
 ;
+
+-- merged table for flows and connecting assets:
+-- 1. Define the asset as the flow
+-- 2. Get the resolution of the flow and the connecting asset (i.e., UNION)
+
+create or replace temp table merged_flows_and_connecting_assets as
+select distinct
+    CONCAT(ftrrp.from_asset, '_', ftrrp.to_asset) as asset,
+    ftrrp.year,
+    ftrrp.rep_period,
+    ftrrp.time_block_start,
+    ftrrp.time_block_end
+from
+    flow_time_resolution_rep_period as ftrrp
+
+union
+
+select distinct
+    CONCAT(ftrrp.from_asset, '_', ftrrp.to_asset) as asset,
+    atrrp.year,
+    atrrp.rep_period,
+    atrrp.time_block_start,
+    atrrp.time_block_end
+from
+    asset_time_resolution_rep_period as atrrp
+join
+    flow_time_resolution_rep_period as ftrrp
+    on (
+        atrrp.asset = ftrrp.from_asset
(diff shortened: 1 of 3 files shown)


src/sql/create-merged-tables.sql
```diff
@@ -88,3 +88,35 @@ create or replace temp table merged_flows_relationship as
         )
         and ftrrp.year = fr.milestone_year
 ;
+
+-- merged table for flows and connecting assets:
+-- 1. Define the asset as the flow
+-- 2. Get the resolution of the flow and the connecting asset (i.e., UNION)
+
+create or replace temp table merged_flows_and_connecting_assets as
```
datejada, line 96: Do we need this table for all flows and all assets?
I mean, if there are no DC-OPF constraints, we won't use this table, aren't we?
Please double-check it. If you don't always need it, you can add a where condition at the end for the cases that we actually need this table.

test/test-data-preparation.jl
```diff
@@ -231,8 +255,21 @@
         ]
         expected_table = DataFrame(expected_rows, expected_cols)
         @test t_highest_out_flows == expected_table
+
+        t_highest_flows_and_connecting_assets = TulipaIO.select_tbl(
+            connection,
+            "t_highest_flows_and_connecting_assets";
+            where_ = "asset = 'input_1_death_star' ORDER BY asset, rep_period, time_block_start, time_block_end",
+        )
+        expected_rows = [
```
datejada, line 264: great! I was expecting this same result from your test input data 😄 

I like that we introduced these tests, they will help us a lot in the future. Thanks! @suvayu for the encouragement to test the pieces, not only the integration.

datejada (commented):
@gnawin thanks for the PR. I left one comment, but in general, it is looking good.

src/sql/create-merged-tables.sql
gnawin: Good point. I added a where condition at the end of the each table for the union, to avoid having more overhead then needed.

test/test-data-preparation.jl
gnawin: I also like this, also because even without this test, we are (at least I am) mocking some test cases to test if the results are correct or not. It's like looking at the lp file to ensure the model is correctly generated. Now with this test, I can write the test while developing the code, and let it actually do the "test".

gnawin:
@datejada thank you for the review, I improved this PR addressing your comments.

datejada (approved):
Great! Thanks for the changes, it looks better.

PR #1207, opened 2025-05-19
Run the case studies twice to ensure it doesn't raise an error

### LINKED ISSUE #1205: Make sure that running `run_scenario` twice works
### Description

I want to run `run_scenario` twice without having to worry about cleaning everything up. It should overwrite everything.

### Validation and testing

_No response_

### Motivation

_No response_

### Target audience

_No response_

### Can you help?

_No response_

abelsiqueira:
Ensure table creation is either 'create or replace' or drop the tables
before creating them. This ensures that they are overwritten.
Run the case studies tests once more running populate_with_defaults
before them to test two things:

- That populate_with_defaults does not change the solution
- That running twice does not raise errors because of existing structures.

## Related issues

Closes #1205

## Checklist

- [x] Tests are passing

### DIFF
--- a/src/sql/create-constraints.sql
+++ b/src/sql/create-constraints.sql
@@ -1,4 +1,10 @@
+drop sequence if exists id
+;
+
 create sequence id start 1
+;
+
+drop table if exists cons_balance_conversion
 ;
 
 create table cons_balance_conversion as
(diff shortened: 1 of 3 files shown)


gnawin (approved):
Thank you. LTGM. 

It would be handy if there are instructions for how to make the sql formatter work :).

abelsiqueira:
> Thank you. LTGM.
> 
> It would be handy if there are instructions for how to make the sql formatter work :).

I haven't been able to make it work consistently, unfortunately. It is currently installed on my editor (not VSCode), and I'm not even sure how to disable now :/

Thanks for the review!

PR #1208, opened 2025-05-19
Add test to make sure that populate_with_defaults doesn't drop extra columns

### LINKED ISSUE #1206: Make sure that populate_with_defaults keeps extra columns
### Description

If the input tables have extra columns, they should not be removed. Create a test for it.

### Validation and testing

_No response_

### Motivation

_No response_

### Target audience

_No response_

### Can you help?

_No response_

abelsiqueira:
Simple test to ensure things won't go awry.

## Related issues

Closes #1206

## Checklist

- [x] Tests are passing

### DIFF
--- a/src/constraints/storage.jl
+++ b/src/constraints/storage.jl
@@ -45,16 +45,16 @@ function add_storage_constraints!(connection, model, variables, expressions, con
                             var_storage_level[row.id-1]
                         else
                             # Find last id of this group (there are probably cheaper ways, in case this becomes expensive)
-                            last_id = only([
-                                row[1] for row in DuckDB.query(
+                            last_id = get_single_element_from_query_and_ensure_its_only_one(
+                                DuckDB.query(
                                     connection,
                                     "SELECT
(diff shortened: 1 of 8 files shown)


abelsiqueira:
It's not natural to Julia, it's too specific.
In other parts of the code I have done different things:

- `only([row[1] for row in DuckDB.query(...)])` (in `src/objective.jl` line 13)
- `only([row.<col_name> for ...])` (in `src/data-preparation.jl` line 174)

And it should also be possible with other functions like `first`, although `only` ensures that there is only one response.

What if I create a helper function `get_and_ensure_unique_from_query(x) = only(only(x))`?

gnawin:
@abelsiqueira it is the double `only()` I was wondering. It is to ensure a unique column and then a unique value. So yes a helper function (due to the self-explanatory name) would help. Thanks!

abelsiqueira:
Thanks, @gnawin. I've created the function and used it in places that had other things. I've re-requested your review, please.

PR #1211, opened 2025-05-20
[Review] Fill NULLs and defaults with populate_with_defaults

### LINKED ISSUE #1210: [Bug] consumer_balance_sense is being incorrectly checked (and data-validation is wrong)
### Description

`consumer_balance_sense` is not allowed to be missing, but it is missing, and this the code is using it assuming that missing means `==`:

https://github.com/TulipaEnergy/TulipaEnergyModel.jl/blob/40a3fd9b038aceba81ba60b8b34a7b0101bb7152/src/constraints/consumer.jl#L20-L24

### Package Version

main

### Julia Version

1.11

### Reproduction steps

See link

### Relevant log output

```shell

```

### Operating System

_No response_

abelsiqueira:
`consumer_balance_sense` should not be missing per the `input-schema.json`. The data validation allowed this because `NULL` values were not being checked properly.
The data was fixed using a script that is also included.

## Related issues

Closes #1210

## Checklist

- [x] Tests are passing


datejada (approved):
@abelsiqueira thank you for the PR. I am glad you changed it, I saw this the other day and I was eager to change it too, since we actually want the three cases to be covered (as in this PR). I forgot the create the issue (my bad), but thanks!

utils/scripts/update-a-column.jl
```diff
@@ -0,0 +1,26 @@
...
+
+    connection = DBInterface.connect(DuckDB.DB)
+    _q(s) = DuckDB.query(connection, s)
+
+    _q("CREATE TABLE t AS FROM read_csv('$filename')")
+
+    # Explicit name of the column and its value
+    _q("UPDATE t SET consumer_balance_sense = '=='")
+
+    _q("COPY t TO '$filename' (HEADER, DELIMITER ',')")
```
datejada, line 25: Apart from this PR, I was doing the same the other day, but when you export the table like this, it does it with "_" instead of "-" (because the table in DuckDB is with "_"). Anyway, the function read csv still works if the files are with "_" instead of "-", which makes me wonder if we just make the CSV files with "_" to make it the same as in DuckDB. What do you think?

abelsiqueira: It makes sense to make that change. But also, let's wait until things are calm, to avoid breaking any rebase in the next few weeks. I'll probably change more of the test files today.

Thanks for the review!

datejada: yeah, no rush, I agree. Thanks!

Issue #1156, opened 2025-04-17
Check (or apply schema) after adding the defaults


datejada:
### Description

The function `populate_with_defaults!` ensures the type and schema of the missing columns, but if the user is defining a column that does not match the schema types, then this function does not guarantee that existing columns have the right type. That can create errors when creating the model. So, we should check (or apply) the schema after adding the missing columsn to ensure all the tables have the right schema, before continuing.

### Validation and testing

Add a test, where the user data has not the write type before populating the defaults.

### Motivation

Avoid errors and improve user experience

### Related

- #1171 

### Target audience

Developers

### Can you help?

 @abelsiqueira help!! 😝

PR #1212, opened 2025-05-21
Create Tinier test case study simplifying Tiny

### LINKED ISSUE #1184: Simplify the test case studies input data by using the defaults function
> > Another comment, maybe for discussion first and then maybe (or not) create another issue:
> > Adding the columns to all the case studies is annoying, but we now have the function to add the defaults. So, we can simplify the input CSV of the case studies and have only the data we need for the case studies. In the tests, we can use the function to populate the defaults to run the tests.
> > That will avoid the hassle of changing the CSVS every time.
> > What do you think? @gnawin @abelsiqueira
> 
> I love this idea :)

Let's comment it with @abelsiqueira when he comes back from holiday

_Originally posted by @datejada in https://github.com/TulipaEnergy/TulipaEnergyModel.jl/issues/1179#issuecomment-2839173438_

abelsiqueira:
The Tinier case is just selecting the Tiny columns that matter.
The test case uses `TulipaIO` directly because the schema is not passed.
The script to select the tables is included.

## Related issues

Closes #1184

## Checklist

- [x] Tests are passing


abelsiqueira:
I have a bug updating the MPS files. I'll mark as draft until I figure out why. I might have to mix this with #1156.

abelsiqueira:
This is working now but is blocked by #1213

abelsiqueira:
@datejada, this is one of the solutions discussed in #1184, let me know if you want a different one, e.g., merging with Tiny.

datejada (commented):
@abelsiqueira thank you for the PR. I had a comment on the tests, but a part from that, it looks good to me.

We can close #1184 with this PR, we are not going to do more than what is here. Thanks!

PR #1213, opened 2025-05-21
Make populate_with_defaults enforce schema (correct types and fill NULLs)

### LINKED ISSUE #1156: Check (or apply schema) after adding the defaults

abelsiqueira:
`populate_with_defaults` now creates a table using the schema and copies the data from the old tables enforcing the data types and defaults. Adds two tests, one for the type correction and one for filling NULLs.

## Related issues

Closes #1156

## Checklist

- [x] Tests are passing

### DIFF
--- a/src/data-preparation.jl
+++ b/src/data-preparation.jl
@@ -1,59 +1,122 @@
 export populate_with_defaults!
 
 """
-    _query_to_complement_with_defaults(connection, existing_table, schema_table)
+    create_str, select_str = _sql_arguments_for_defaults(connection, table_name, table_schema)
 
-Creates a query to completement a given table `existing_table` with the defaults stored in `schema_table`.
+Returns the strings to complement the table creation informing the name, type
+and possible default, and the select with coalescing and type casting.
 
-The `existing_table` should be a table in `connection`.
+The `table_schema` should be the complete schema, not only the `type`.
+For instance, `TulipaEnergyModel.schema[table_name]`.
 
-The `schema_table` should be a dictionary from the column to a dictionary of the column properties.
-For instance, from `TulipaEnergyModel.schema[table_name]`.
+For each column of the `table_name` table, the creation string is like
 
-The output is the argument for a 'SELECT' query that selects all existing
-columns from `existing_table`, and selects the default value for all other
-columns given in `, columnschema_table`.
-If a column in `schema_table` does not have a default and is not present in
-`existing_table`, then a `DataValidationException` is raised.
+    COLUMN_NAME COLUMN_TYPE
+    -- For a column with no default type
+
+or
+
+    COLUMN_NAME COLUMN_TYPE DEFAULT COLUMN_DEFAULT
+    -- For a column with default type
+
+and the selection string is like
+
+    COLUMN_NAME::COLUMN_TYPE
+    -- For an existing column with no default type
+
+or
+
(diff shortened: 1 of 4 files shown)


abelsiqueira:
@datejada, I also fixed the schema default for a few other columns

src/data-preparation.jl
```diff
@@ -178,7 +254,7 @@ function create_unrolled_partition_tables!(connection)
             rep_periods_data.year,
             rep_periods_data.rep_period,
             COALESCE(arpp.specification, 'uniform') AS specification,
-            COALESCE(arpp.partition, '1') AS partition,
+            COALESCE(arpp.partition::string, '1') AS partition,
```
datejada, line 257: I am glad with this change, I had to make a hack in the TLC to cover this while the workflow was updated. Thanks :)

src/data-preparation.jl
```diff
@@ -1,59 +1,122 @@
...
+    selection_arguments = String[]
+
+    # First the existing columns
+    for (col_name, col_type, col_default) in existing_columns
+        # If the column is in the schema, then replace the type and possibly add a default
+        if haskey(table_schema, col_name)
+            # Expected columns
+            props = table_schema[col_name]
+            col_type = props["type"]
+            if haskey(props, "default") && !isnothing(props["default"]) # Only update the default if there is something there
```
datejada, line 62: I am curious what happens if the user doesn't provide one of the columns that is required for the table (i.e., the primary keys). What is going to happen then? Maybe in a separate issue, we need to work on enforcing the primary keys (not sure where or how, though)

src/input-schemas.json
```diff
@@ -279,7 +279,7 @@
       "constraints": {
         "minimum": 0
       },
-      "default": 0,
+      "default": null,
```
datejada, line 282: this defaults makes more sense

test/test-defaults.jl
```diff
@@ -77,6 +77,78 @@
...
+                    FROM duckdb_columns()
+                    WHERE table_name = 'assets_rep_periods_partitions'
+                        AND column_name = 'partition'
+                    ",
+                ),
+            )
+        @test type_of_partition_column == "VARCHAR"
+    end
+
+    @testset "Test that populate_with_defaults fills NULL with defaults" begin
```
datejada, line 116: Great! just as expected :) I like the tests clarity (also helps to follow the code before)

datejada (approved):
@abelsiqueira Thanks, great changes, I managed to follow them, I left some comments, but for my understanding, I don't have proposed changes in the code. So, approving.

src/data-preparation.jl
abelsiqueira: Yes, agree. This may be overkill, with the extra populate changes, BUT it is free, so better to have it.

abelsiqueira:
Thanks for the review!

PR #1216, opened 2025-05-21
Validate data for DC OPF

### LINKED ISSUE #1214: Add validations for dc-opf data
- reactance must be > 0 
- dc-opf only applies to non-investable transport flows

gnawin:
## Related issues

Closes #1214 

## Checklist

- [ ] Tests are passing

### DIFF
--- a/src/data-validation.jl
+++ b/src/data-validation.jl
@@ -4,7 +4,7 @@ export DataValidationException
 TulipaIO.FmtSQL.fmt_quote(::Nothing) = "NULL"
 
 """
-    DataValidationException
+	DataValidationException
 
 Exception related to data validation of the Tulipa Energy Model input data.
 """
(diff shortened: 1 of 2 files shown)


abelsiqueira (changes requested):
The many small changes by the formatter are not good.
The changes are replacing **some** spaces by tabs, i.e., the result is a mix of spaces and tabs.  
I'm not sure why they happened or why the pre-commit allowed them. But that's definitely one of the things we don't want. I'm rejecting to make it clear that we need to figure it out, but I don't have any guesses at the moment.

datejada (changes requested):
Same as @abelsiqueira we need to figure out the formatter problem before merging this. But, apart from that, it looks good to me.

abelsiqueira (approved):
LGTM

abelsiqueira:
Fixed the tabs to spaces issue manually.

PR #1248, opened 2025-06-05
Fix bug investment related variables

### LINKED ISSUE #1247: [Bug] Assets with investment method `none` can have investments
### Description

Currently we only check investable for create the variables, so we can have investment variables even if the investment method is set to `none`, here is the snnipet:

https://github.com/TulipaEnergy/TulipaEnergyModel.jl/blob/e0458292c25c9d2ec6e92ade6ecc0ca748f65990/src/sql/create-variables.sql#L245-L247

This needs to be checked also for the storage investment and flow investment.

### Package Version

0.16.0

### Julia Version

1.1

### Reproduction steps

In the multi-year investment case study asests to `none` and the ones investable will have investments.

### Relevant log output

```shell

```

### Operating System

_No response_

gnawin:
**Model changes:**
- Ensure `var_assets_investment` are only created for **simple or compact** method
- Ensure `var_assets_decommission` are only created for **simple or compact** method
- Ensure `var_assets_investment_energy` are only created for **simple** method
- Ensure `var_assets_decommission_energy` are only created for **simple** method

**Test changes**
- `Tiny` case has `investable=true` for `investment_method=none`. It is corrected in this PR. Hence the test change.
- Almost all other cases have such conflicts (except for Multi-year!), but the results do not change. So I updated all MPS files.

## Related issues

Closes #1247 

## Checklist

- [x] Tests are passing

### DIFF
--- a/benchmark/model-mps-folder/Tinier.mps
+++ b/benchmark/model-mps-folder/Tinier.mps
@@ -1443,6 +1443,79 @@ COLUMNS
     flow[(solar,demand),2030,3,24:24] max_output_flows_limit_simple_method[solar,2030,3,24:24] 1
     flow[(solar,demand),2030,3,24:24] consumer_balance[demand,2030,3,24:24] 1
     MARKER    'MARKER'                 'INTORG'
+    assets_investment[2030,ccgt] max_output_flows_limit_simple_method[ccgt,2030,1,1:1] -400
+    assets_investment[2030,ccgt] max_output_flows_limit_simple_method[ccgt,2030,1,2:2] -400
+    assets_investment[2030,ccgt] max_output_flows_limit_simple_method[ccgt,2030,1,3:3] -400
+    assets_investment[2030,ccgt] max_output_flows_limit_simple_method[ccgt,2030,1,4:4] -400
+    assets_investment[2030,ccgt] max_output_flows_limit_simple_method[ccgt,2030,1,5:5] -400
+    assets_investment[2030,ccgt] max_output_flows_limit_simple_method[ccgt,2030,1,6:6] -400
+    assets_investment[2030,ccgt] max_output_flows_limit_simple_method[ccgt,2030,1,7:7] -400
+    assets_investment[2030,ccgt] max_output_flows_limit_simple_method[ccgt,2030,1,8:8] -400
+    assets_investment[2030,ccgt] max_output_flows_limit_simple_method[ccgt,2030,1,9:9] -400
+    assets_investment[2030,ccgt] max_output_flows_limit_simple_method[ccgt,2030,1,10:10] -400
+    assets_investment[2030,ccgt] max_output_flows_limit_simple_method[ccgt,2030,1,11:11] -400
+    assets_investment[2030,ccgt] max_output_flows_limit_simple_method[ccgt,2030,1,12:12] -400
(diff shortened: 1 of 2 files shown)


src/sql/create-variables.sql
```diff
@@ -269,6 +270,7 @@ from
     left join asset on asset.asset = asset_both.asset
 where
     asset_both.decommissionable
+    and asset.investment_method in ('simple', 'compact')
```
datejada, line 273:
```suggestion
```

datejada (commented):
Hi @gnawin Thanks for the PR. I am not sure about having the decommission variable not created when `investment_method` is `none`; because, you can have existing units (initial units) that you want the model to decide to decommission or not, but you don't want to have investment variables. For example, and old CCGT technology, that you don't want to invest more, but you want to know if it is better to decommission or not. Is it clear?

Also, because the `investment_method` should affect only investment variables not decommission variables.

What do you think?

gnawin:
> Hi @gnawin Thanks for the PR. I am not sure about having the decommission variable not created when `investment_method` is `none`; because, you can have existing units (initial units) that you want the model to decide to decommission or not, but you don't want to have investment variables. For example, and old CCGT technology, that you don't want to invest more, but you want to know if it is better to decommission or not. Is it clear?
> 
> Also, because the `investment_method` should affect only investment variables not decommission variables.
> 
> What do you think?

I see your point. 

In my mind, having `investment_method = none` means you would like to keep the initial units and do nothing with it. In that sense `investment_method` is used as a general method for both investment and decommissions. If you want decomissions, you need to set `investment_method != none` and `decommission=true`,  
The reason for this logic is decommission variables are directly related to the investment variables. I.e., the decom variables are different when you have `simple` or `compact`. 

That being said, I can see if it feels a bit weird to call it `investment_method` while it affects both investment and decom variables. Especially in the future when we have retrofits, but that one we still need to discuss, and I think we should come back to this after that.

datejada:
> > Hi @gnawin Thanks for the PR. I am not sure about having the decommission variable not created when `investment_method` is `none`; because, you can have existing units (initial units) that you want the model to decide to decommission or not, but you don't want to have investment variables. For example, and old CCGT technology, that you don't want to invest more, but you want to know if it is better to decommission or not. Is it clear?
> > Also, because the `investment_method` should affect only investment variables not decommission variables.
> > What do you think?
> 
> I see your point.
> 
> In my mind, having `investment_method = none` means you would like to keep the initial units and do nothing with it. In that sense `investment_method` is used as a general method for both investment and decommissions. If you want decomissions, you need to set `investment_method != none` and `decommission=true`, The reason for this logic is decommission variables are directly related to the investment methods. I.e., the decom variables are different when you have `simple` or `compact`.
> 
> That being said, I can see if it feels a bit weird to call it `investment_method` while it affects both investment and decom variables. Especially in the future when we have retrofits, but that one we still need to discuss, and I think we should come back to this after that.

I see, okay, that explanation is clear, then. However, the name `investment_method` is not ideal, as it misleads users (or at least it did to me). Would you mind creating an issue to modify that name?

gnawin:
@datejada  Would you mind having a look again? I only changed `ccgt` to `simple` in Tiny and `Tinier`, so the objective values do not change anymore. 
All MPS files change because of the inconsistency in data, they will be tackled in #1249 so the MPS files will change again with consistent data. And since this is a bug, I will do it separate from this issue.
Thank you!

PR #1254, opened 2025-06-10
Fix to return non-fatal error instead of ErrorException

datejada:
This pull request makes a small change to improve error handling in the `export_solution_to_csv_files` function in `src/io.jl`. Instead of raising an error with `error()`, it now uses the `@error` macro and returns immediately if the `energy_problem` has not been solved.

The problem is that if the problem is infeasible, having an `error()` does not allow the model to be returned for further debugging of the infeasibility. The `@error` leads to a non-fatal error that helps to run IIS afterwards to check the possible problems in the model.

## Related issues

There is no related issue.

## Checklist

- [x] Tests are passing


datejada:
@abelsiqueira This works as I want, and we can keep the error exception 😄 Let me know if you have more comments. Thanks!

src/io.jl
```diff
@@ -64,7 +64,7 @@ Notice that this assumes that the solution has been computed by [`save_solution!
 """
 function export_solution_to_csv_files(output_folder, energy_problem::EnergyProblem)
     if !energy_problem.solved
-        error("The energy_problem has not been solved yet.")
+        return error("The energy_problem has not been solved yet.")
```
abelsiqueira, line 67:
```suggestion
        error("The energy_problem has not been solved yet.")
```

src/run-scenario.jl
```diff
@@ -56,7 +56,7 @@ function run_scenario(
 
     @timeit to "save_solution!" save_solution!(energy_problem)
 
-    if output_folder != ""
+    if output_folder != "" && energy_problem.solved
```
abelsiqueira, line 59: Since you mentioned, how about adding an `@warn` here? Like

```julia
if output_folder != ""
    if energy_problem.solved
        ...
    else
        @warn "..."
    end
end
```

abelsiqueira (approved):
Small suggestions. Otherwise approved

datejada:
@abelsiqueira lesson learned _"don't try to fix stuff that has workarounds, because you might face the rage of the codecov"_ 🤣 

This started as a minor fix, and took a lot of time to make the codecov happy 🤷 when I could have just lived with the fact that if the model is infeasible, then don't define the output folder 🤣 

Anyway, I am unsure if this is the best way to test this. I tried simpler approaches, but since `run_scenario` has several warnings, you need to capture all of them and check that the new one I added is in them. But, It implied to add a package to the test environment. So, maybe there is a better way, but I couldn't figure it out. If you have a better suggestion, please let me know.

I will wait for your feedback before merging, just in case 😄

Issue #749, opened 2024-08-23
Create individual test for the functions that creates the constraints in the model


datejada:
### What and Why

We want to add tests for each function that creates constraints that do not depend on the case studies (as it currently does). For instance, to test the ramping constraints, we have the Norse case with some input data that creates the constraint. But we would like to have a test that, given some inputs, returns the expected constraint (maybe without running the full energy problem).

This approach is similar to the one in SpineOpt (which is a good practice there)

### Possible Drawbacks

- It increases the number of tests
- It might be not easy to set up the first time.

### Related Issues

Blocked by #701

datejada:
@abelsiqueira, feel free to add or change something in the issue.

PR #1272, opened 2025-07-15
Create tests for group constraints

abelsiqueira:
We tested claude-code using the prompt 'Please implement a test for
"group constraints". Look at the test/test-constraints-*.jl files for
inspiration.'

This also includes a notice on the README about using LLM coding assistants.

## Related issues

Related to #749

## Checklist

- [x] Tests are passing

### DIFF
--- a/test/test-constraint-group.jl
+++ b/test/test-constraint-group.jl
@@ -1,0 +1,180 @@
+@testset "Test add_group_constraints!" begin
+    # Setup a temporary DuckDB connection and model
+    connection = DBInterface.connect(DuckDB.DB)
+    model = JuMP.Model()
+
+    # Create mock tables for testing using register_data_frame
+
+    # Asset table - defines which assets belong to which group
+    asset_rows = [
+        ("wind_1", "renewable_group", 100.0, 2025),
+        ("wind_2", "renewable_group", 150.0, 2025),
+        ("solar_1", "renewable_group", 80.0, 2025),
+        ("gas_1", "fossil_group", 200.0, 2025),
+        ("gas_2", "fossil_group", 250.0, 2025),
+    ]
+    asset = DataFrame(asset_rows, [:asset, :group, :capacity, :milestone_year])
+    DuckDB.register_data_frame(connection, asset, "asset")
+
+    # Group asset table - defines the groups
+    group_asset_rows = [("renewable_group", "renewable"), ("fossil_group", "fossil")]
+    group_asset = DataFrame(group_asset_rows, [:name, :type])
+    DuckDB.register_data_frame(connection, group_asset, "group_asset")
+
+    # Variable assets investment table
+    var_assets_investment_rows = [
+        (1, "wind_1", 2025, true, 100.0, 500.0),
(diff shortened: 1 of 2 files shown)


test/test-constraint-group.jl
```diff
@@ -0,0 +1,180 @@
...
+        [(1, "renewable_group", 200.0), (2, "fossil_group", 100.0)]
+    cons_group_min_investment_limit =
+        DataFrame(cons_group_min_investment_limit_rows, [:id, :name, :min_investment_limit])
+    DuckDB.register_data_frame(
+        connection,
+        cons_group_min_investment_limit,
+        "cons_group_min_investment_limit",
+    )
+
+    # Create empty tables for other investment variables that are required
```
suvayu, line 60: Do I understand this correctly that these empty tables are created explicitly because the code being tested expects this, but it's created elsewhere?

Either way, I don't think this is the correct way to do it, because it is essentially an "alternate implementation" of wherever Tulipa creates the tables.  Unless I recall incorrectly, the tables are created from the "schema" files in JSON format.  That should be used here as well, otherwise this test will go out of sync when the schema changes.

**PS:** If there's no generalised function that does the create table from the JSON files, then maybe this is also the opportunity to do a "good samaritan refactor" and add that feature in this unrelated PR.

test/test-constraint-group.jl
```diff
@@ -0,0 +1,180 @@
...
+    var_assets_investment_container = variables[:assets_investment].container
+
+    # Test renewable group max constraint: 100*x1 + 150*x2 + 80*x3 ≤ 1000
+    renewable_max_constraint = model[:investment_group_max_limit][1]
+    observed_con = JuMP.constraint_object(renewable_max_constraint)
+    expected_con = JuMP.@build_constraint(
+        100.0 * var_assets_investment_container[1] +
+        150.0 * var_assets_investment_container[2] +
+        80.0 * var_assets_investment_container[3] ≤ 1000.0
+    )
```
suvayu, line 150: I'm just picking out this one example, but the comments apply to this whole approach.  I don't really like how the example input data is used.  Basically these lines should be "magically consistent" with lines 9-15 above.  One immediate way to rectify this to actually use that dataframe column here.

```julia
# pseudo-code, replace with whatever is appropriate
asset_rows[:capacity] * var_assets_investment_container
```

I wouldn't be happy with the above either, but I could live with it.  I'll elaborate in the main review comment, since these inline comments are hard to find after the PR is merged, and I want my comment to be easily rediscoverable.

suvayu:
Ah, I just saw your comments @datejada, I guess my comments about example input data are pretty much what you say about `add_investment_variables!`

abelsiqueira:
Talked to @datejada and @gnawin this morning. My main concern here is about using LLMs. Whether we want to or not, and if our policies agree. @suvayu's comments in this regard can be incorporated and can probably be made explicit in a llm-focused file (e.g., `CLAUDE.md`). I'm not sure about the label, though.

The submitted code brought attention to other issues/opportunities in the testing and the investment variables creation. @datejada has mentioned an intent to refactor, so I'll leave further development for him.

PR #1291, opened 2025-07-22
Refactor add investment variables

datejada:
The main idea is to give more flexibility and readability to the investment and decommission variable creation by splitting their creation and addition to the model. Two new tests are added for the variable creation.

## Related issues

Closes #1272

## Checklist

- [x] Tests are passing

### DIFF
--- a/test/test-variable-investment.jl
+++ b/test/test-variable-investment.jl
@@ -10,7 +10,7 @@
         milestone_year = [2030, 2030, 2040, 2040],
         investment_integer = [true, false, true, true],
         capacity = [100.0, 50.0, 25.0, 125.0],
-        investment_limit = [1000.0, missing, 500.0, 650.0],  # Should result in [10.0, Inf, 20.0, 5.0]
+        investment_limit = [1000.0, missing, 500.0, 650.0],
     )
     DuckDB.register_data_frame(connection, asset_investment_data, "var_assets_investment")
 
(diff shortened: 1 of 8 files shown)


datejada:
I updated the MPS file because, the order of the rows changed due to the refactor of the function to add investment. I double-checked, and the constraints are the same, only the position changes. I suppose it is due to the way the variables are added to the JuMP model, then, they might be displaced in the MPS file, but let me know what you think @abelsiqueira (maybe there is something else I am missing)

abelsiqueira:
The difference appears to be (semi-manual check) that there are two extra lines: `MARKER 'MARKER'  'INTORG'` and `MARKER 'MARKER' 'INTEND'`. 
Previously, there were already one of each, and I think now there are two of each. I haven't checked explicitly where they are, and I don't know what they mean yet.

test/test-variable-investment.jl
```diff
@@ -0,0 +1,95 @@
...
+    variables = Dict{Symbol,TulipaEnergyModel.TulipaVariable}(
+        key => TulipaEnergyModel.TulipaVariable(connection, "var_$key") for
+        key in (:flows_investment, :assets_investment, :assets_investment_energy)
+    )
+
+    # Call the function under test
+    TulipaEnergyModel.add_investment_variables!(model, variables)
+
+    # Test that variables were created correctly
+    @test length(variables[:assets_investment].container) ==
```
gnawin, line 54: Do we want to do something similar to the tests for constraints, i.e., we have a expected var, and compare it against an observed var? It is fine to test this way, but to be more accurate and consistent, maybe we would want to do that.

datejada: Great! I like the proposal. Thanks, I will update the code with this function.

abelsiqueira:
I've update the MPS comparison in https://github.com/TulipaEnergy/TulipaEnergyModel.jl/pull/1293

datejada:
@abelsiqueira rebased with your fix to the MPS, so no need to update the MPS file 😄 Also, I have included the new function as suggested in https://github.com/TulipaEnergy/TulipaEnergyModel.jl/pull/1291#discussion_r2224777254.
Please let me know if you have further comments.
Thanks!

PR #1293, opened 2025-07-23
Add generic functions to create variables with specifications per row

abelsiqueira:
The MARKER INTORG/INTEND blocks denote blocks of integer variables.
If a model reorders the variables, there may be a different number of blocks
leading to a false negative.

## Related issues

Observed in PR #1291

## Checklist

- [ ] Tests are passing


datejada (approved):
LGTM. Thanks for the improvement 😃

PR #1323, opened 2025-08-28
[Multi-year Part II] Add low sum limit for transport vintage outgoing flows for semi-compact method

### LINKED ISSUE #1314: Add lower limit for the sum of the flows

gnawin:
- Add a new lower bound constraint for the semi-compact method similar to the 2nd constraint here [lower bound](https://tulipaenergy.github.io/TulipaEnergyModel.jl/stable/40-scientific-foundation/40-formulation/#Lower-Limit-for-Flows), which are for other methods
- Refactor the whole transport constraint into separate functions per constraint, so that it would be easier for testing.

## Related issues

Closes #1314 

## Checklist

- [x] Tests are passing

### DIFF
--- a/src/sql/create-constraints.sql
+++ b/src/sql/create-constraints.sql
@@ -277,19 +277,19 @@ create table cons_min_outgoing_flow_for_transport_flows_without_unit_commitment 
 -- This information is gathered from the flow table
 -- COALESCE is used to handle the case where there are no outgoing flows
 with
-    transport_flow_info as (
+    cte_transport_flow_info as (
         select
             asset.asset,
             coalesce(
                 (
                     select
-                        bool_or(flow.is_transport)
+                        bool_or(flow.is_transport) -- true if any outgoing flow is transport
                     from
                         flow
                     where
                         flow.from_asset = asset.asset
                 ),
-                false
+                false -- coalescing to false in case there are no outgoing flows
             ) as outgoing_flows_have_transport_flows,
         from
             asset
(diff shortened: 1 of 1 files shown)


src/constraints/transport.jl
```diff
@@ -131,8 +187,6 @@ function add_transport_constraints!(
             ],
         )
     end
-
-    return
```
abelsiqueira, line 135: The formatter was supposed to enforce adding a `return` to the end of functions (we have an option for that). I'm guessing this is another random issue with the formatter version. We should consider alternatives in October.
No action required.

src/sql/create-constraints.sql
```diff
@@ -304,6 +306,61 @@ where
     and transport_flow_info.outgoing_flows_have_transport_flows
     -- Assets with unit commitment already have a minimum outgoing flow constraints
     and not asset.unit_commitment
+    and asset.investment_method in ('compact','simple', 'none')
```
abelsiqueira, line 309:
```suggestion
    and asset.investment_method in ('compact', 'simple', 'none')
```

src/sql/create-constraints.sql
```diff
@@ -304,6 +306,61 @@ where
...
+                        flow
+                    where
+                        flow.from_asset = asset.asset
+                ),
+                false
+            ) as outgoing_flows_have_transport_flows,
+        from
+            asset
+    )
+select
```
abelsiqueira, line 342: It might be good to benchmark this, just to be sure there are no surprises. Maybe run locally if its easier than the branch change.

src/sql/create-constraints.sql
```diff
@@ -304,6 +306,61 @@ where
...
+-- but it applies to vintage flows instead of regular flows
+create table cons_min_outgoing_flow_for_transport_vintage_flows as
+with
+    transport_flow_info as (
+        select
+            asset.asset,
+            coalesce(
+                (
+                    select
+                        bool_or(flow.is_transport)
```
abelsiqueira, line 331:
```suggestion
                        bool_or(flow.is_transport) -- true if any outgoing flow is transport
```

src/sql/create-constraints.sql
```diff
@@ -304,6 +306,61 @@ where
...
+                    from
+                        flow
+                    where
+                        flow.from_asset = asset.asset
+                ),
+                false
```
abelsiqueira, line 337:
```suggestion
                false -- coalescing to false in case there are no outgoing flows
```

abelsiqueira (changes requested):
Thanks, @gnawin. I've left some commenting suggestions and a benchmark suggestion.
One of the big change improvements for the future would be the ability to test tables more easily. I.e., in this case that this table produces the expected rows (without having to test the whole constraints).

PR #1326, opened 2025-08-29
[Multi-year Part II] Add low sum limit for transport vintage outgoing flows for semi-compact method

### LINKED ISSUE #1314: Add lower limit for the sum of the flows

gnawin:
Copy of #1323, pushed to upstream to run benchmark. 

- Add a new lower bound constraint for the semi-compact method similar to the 2nd constraint here [lower bound](https://tulipaenergy.github.io/TulipaEnergyModel.jl/stable/40-scientific-foundation/40-formulation/#Lower-Limit-for-Flows), which are for other methods
- Refactor the whole transport constraint into separate functions per constraint, so that it would be easier for testing.

## Related issues

Closes #1314 

## Checklist

- [x] Tests are passing


abelsiqueira (approved):
The benchmark doesn't look bad. Thanks for moving branches.

Issue #1269, opened 2025-07-09
Add profiles to the cost coefficients in the objective function


datejada:
### Description

The current costs in the objective function are constant values, such as the variable cost of the flow. But there are cases where having a profile for the cost is also useful/relevant. For example, the gas cost/price is higher during winter than in summer. So, having a general implementation where the costs can be multiplied by a profile (as with the availability and demand) will allow the model to model these situations more straightforwardly.

### Validation and testing

Modify the Norse case to have a gas cost profile and check that it works (.LP file)

### Motivation

Cover more use cases

### Target audience

developers

### Can you help?

Always ;)

datejada:
@aporrasc I added the Use Case 2 to this issue. Please feel free to add or comment more, if you need to. Thanks!

datejada:
For the DEMOSES project, we will implement only the commodity price profiles. For operational costs, we will decide whether there is a use case for them. The current need for having a profile of electricity prices is covered in #1410

PR #1327, opened 2025-09-02
[WIP] Rolling horizon

abelsiqueira:
Playing around with rolling horizon.

## Related issues

Closes #

## Checklist

- [ ] Tests are passing


test/test-case-studies.jl
```diff
@@ -143,6 +143,22 @@ end
...
+    energy_problem = TulipaEnergyModel.run_scenario(connection; show_log = false)
+    @test energy_problem.objective_value ≈ 10000.0 atol = 1e-5
+    # populate_with_defaults shouldn't change the solution
+    TulipaEnergyModel.populate_with_defaults!(connection)
+    energy_problem = TulipaEnergyModel.run_scenario(connection; show_log = false)
+    @test energy_problem.objective_value ≈ 10000.0 atol = 1e-5
+    # Rolling horizon
+    energy_problem =
+        TulipaEnergyModel.run_rolling_horizon(connection, 24 * 7, 48 * 7; show_log = false)
+    # @test energy_problem.objective_value ≈ 10000.0 atol = 1e-5
```
datejada, line 159: We need to test the objective function of each one of the windows instead of the overall objective function

datejada:
> @datejada, I've created a first draft of the rolling horizon implementation. I don't expect you to read everything, but I've marked a few interesting places, and we can discuss and go over the details if you have time.
> 
> I'm trying something with the profiles which hides the complexity in `_profile_aggregate`. This avoids changing the code everywhere and manually handling rolling horizon inside the model. I am still not sure whether this is efficient or not, but it's what I thought of so far.
> 
> I also have the following questions:
> 
> * Are there other places that will require rolling horizon? Are we trying to make it more generic than this? If so, how generic?
> * I am not sure how to validate that the rolling horizon is working. I've created some random problem that possibly doesn't make a lot of sense. Do you have some Tulipa-compatible test? If not, maybe we can go over the TulipaBuilder implementation and see what comes out of it.

@abelsiqueira, thanks for the PR. It is starting to take shape 😃 

1. I like the approach with the profiles. Is it general enough to add the cases in issue #1269? I am trying to see how this works with the changes we foresee in the model.
2. **(TL;DR: we need a generic implementation for the scalar parameters)**.
The initial storage level looks okay, but I am afraid we need something more general here. The big picture is that more "Float" parameters can be updated in the model during the rolling horizon. Some of those parameters will be boundary conditions when k=1 (similar to the initial storage level). We don't have them yet in the model, but we need to add them soon, because they are relevant for the rolling_horizon. For instance, the initial flow for the ramping constraints #1328, for a whole year, it is not that relevant to this initial condition, but if we solve weekly windows, then it becomes relevant since then you are not including this condition 52 times 🤷 There a couple more of this "initial/boundary condition" parameters that will come from the developments in the unit commitment we have with the TU-Delft students.
3. This is how it think we should test the case study of the rolling horizon (but open to discussion, with the rest of the team): https://github.com/TulipaEnergy/TulipaEnergyModel.jl/pull/1327/files#r2321431664
4. The last point leads me to question how the results are handled. For debugging, it would be nice to have the results of each optimization per window, and then one on which we aggregate the final result, similar to here: https://jump.dev/JuMP.jl/stable/tutorials/algorithms/rolling_horizon/#Solution
5. It will be nice to have the same example that is in the JuMP documentation (https://jump.dev/JuMP.jl/stable/tutorials/algorithms/rolling_horizon/), which could be the example for rolling horizon (or an extra one, this feature might need a couple of examples). What do you think?

Thanks again 😄 I like the changes

abelsiqueira:
Thanks for the initial review, @datejada. Some comments:

1. I don't know what issue #1269 will require. It's not clear what would need to be implemented to handle that. That said, from working on the inflows changes is that we need to use `_profile_aggregate`. I.e., the aggregation cannot happen at SQL level, it must happen in Julia level. If that's the case, it should be fine.
2. There are currently two categories of "Rolled Parameters":
   - Profiles (starting value and updates clearly defined from the window)
   - Starting values (starting value is the same as non-rolling, and the rolling update uses the solution from a window as new starting value)
   Are you proposing a new category, or do these scalar parameters fall into these categories?
3. Makes sense. Probably some individual components can be tested as well. It might not be enough to capture everything, like a PR that only passes because it's not covered in the rolling horizon case study, as we've seen with other features. 
4. Indeed. I was considering whether it made sense to attach all solutions to existing variables, but it will probably be messy. Instead, maybe we can create new tables for the rolling horizon windows with ids and some tables for solutions.
5. It would be nice, but I have no idea how to translate that into TEM format. If you have the time to help translate that into TEM, it would be great. Maybe we can sit together and write a TulipaBuilder script for it, if you have some time.

Thanks again for the review

Issue #1325, opened 2025-08-29
Change the objective function to allow for vintage costs


gnawin:
### Description

In the objective function, the flow variable cost is for all flows, and each of the flow is attached with a variable cost (among other factors). The flow variable cost belongs to flow_milestone. As it is now, it is impossible to consider the costs of vintage units which depend on commission years, instead of milestone years.

The proposal is to split the total variable cost into two parts, one for incoming flows, the other for outgoing flows. The outgoing flows are further split into two parts, one for semi-compact investment method, the other for the rest.

See the formulation below and it will be further explained using an example.

```math
\begin{aligned}
& \sum_{f \in F^{\text{out}}_{a,y} | a \in \mathcal{A}^{\text{semi-compact investment}}} \sum_{y \in \mathcal{Y}} \sum_{v \in \mathcal{V}} (p^{\text{commodity price}}_{a, y} / p^{\text{efficiency}}_{a, v} + p^{\text{operational cost}}_{f, y}) * v^{\text{vintage flow}}_{f,y,v} \\
& \sum_{f \in F^{\text{out}}_{a,y} | a \in \mathcal{A}^{\text{compact investment}} \cup \mathcal{A}^{\text{simple investment}} \cup \mathcal{A}^{\text{operation}} } \sum_{y \in \mathcal{Y}}  (p^{\text{commodity price}}_{a, y} / p^{\text{efficiency}}_{a, v} + p^{\text{operational cost}}_{f, y}) * v^{\text{flow}}_{f,y}
\end{aligned}
```
To align with [INES-spec](https://github.com/ines-tools/ines-spec/blob/v1.x/ines-spec-parameter-definitions.csv), we need to make two changes in the data.
- variable cost needs to be renamed to operational cost in flow_milestone
- new parameter commodify price in asset_milestone, which only applies to producers.

gnawin:
### Example

We will use an numerical example to explain the formula. The example is to model CCGT in two different ways.

#### CCGT as a producer

```mermaid
flowchart LR
  A[Producer: CCGT] --> B[demand]
```

_Semi-compact method for CCGT_
This method indicates we would like to consider vintages of CCGT. Consider two commission years 2030 and 2050 at the milestone year 2050, i.e., we have two different units of CCGT with two vintage flows. For illustrative purposes, we assume both of them equal to 1, so satisfying a total demand of 2, i.e., 
```math
[...]

PR #1329, opened 2025-09-04
Two changes in data

gnawin:
To align with [INES-spec](https://github.com/ines-tools/ines-spec/blob/v1.x/ines-spec-parameter-definitions.csv), we need to make two changes in the data.

- variable_cost needs to be renamed to operational_cost in flow_milestone
- new parameter commodify_price in asset_milestone, which only applies to producers.

Also renamed the file names in MIMO to be consistent with other test inputs.

## Related issues

Related to #1325.

## Checklist

- [x] Tests are passing


gnawin:
Hi @datejada, this is the first PR for closing #1325. It's only data changes based on our discussions.

There will be more PRs for #1325, and the next step would be to change the objective function itself. Could you have a look there also and see if the proposal makes sense? Thanks a lot!

PR #1330, opened 2025-09-05
[Multi-year Part II] Change the name of conversion efficiency

gnawin:
## Related issues

Related to #1325 

## Checklist

- [x] Tests are passing


datejada (approved):
Thanks! @gnawin It is according to what we discussed and aligns better with INES, so I am approving it. I agree that we need to have an overall review in the documents at the end of this epic.

PR #1332, opened 2025-09-05
[Multi-year Part II] Modify objective function

### LINKED ISSUE #1325: Change the objective function to allow for vintage costs

gnawin:
The content of this PR:
- Update existing objective function to account for commodity prices
- Add a new term to the objective function to consider the vintage flow costs

## Related issues

Closes #1325 

## Checklist

- [x] Tests are passing

### DIFF
--- a/src/objective.jl
+++ b/src/objective.jl
@@ -247,10 +247,12 @@ function add_objective!(connection, model, variables, expressions, model_paramet
         LEFT JOIN asset
             ON asset.asset = var.from_asset
         WHERE asset.investment_method != 'semi-compact'
-        ORDER BY var.id
         ",
     )
 
+    # For the flows_operational_cost, we cannot use the zip method as done in all other terms,
+    # because there are more flow variables than the number of rows in indices,
+    # i.e., we only consider the costs of the flows that are not in semi-compact method
     var_flow = variables[:flow].container
 
     flows_operational_cost = @expression(model, sum(row.cost * var_flow[row.id] for row in indices))
@@ -307,10 +309,13 @@ function add_objective!(connection, model, variables, expressions, model_paramet
         ",
     )
 
-    var_vintage_flow = variables[:vintage_flow].container
-
-    vintage_flows_operational_cost =
-        @expression(model, sum(row.cost * var_vintage_flow[row.id] for row in indices))
+    vintage_flows_operational_cost = @expression(
+        model,
+        sum(
+            row.cost * vintage_flow for
+            (row, vintage_flow) in zip(indices, variables[:vintage_flow].container)
+        )
+    )
 
     indices = DuckDB.query(
         connection,
@@ -353,8 +358,12 @@ function add_objective!(connection, model, variables, expressions, model_paramet
         ",
     )
 
-    var_units_on = variables[:units_on].container
-    units_on_cost = @expression(model, sum(row.cost * var_units_on[row.id] for row in indices))
+    units_on_cost = @expression(
+        model,
+        sum(
+            row.cost * units_on for (row, units_on) in zip(indices, variables[:units_on].container)
+        )
+    )
 
     ## Objective function
     @objective(


gnawin:
Hi @datejada this PR is easier to follow if you look commit by commit. You could check, especially, the first commit (https://github.com/TulipaEnergy/TulipaEnergyModel.jl/pull/1332/commits/047d3fef14a5a2a34adb0c5c35ef99f783142ae7), which is to modify the existing objective function (by adding the commodify price). The following commits are just to include the vintage flow cost.

@abelsiqueira I would like especially your opinion on the second commit (https://github.com/TulipaEnergy/TulipaEnergyModel.jl/pull/1332/commits/c0e909eb6db3dd52734991a6ed16cc91834532f7) for the new term of vintage flow cost. The way I implemented does what I want it to do (at least from what I see, so please raise any questions where you find confusing). However, the new table `t_objective_vintage_flows` is constructed differently from `t_objective_flows`. Let me know what you think.

Thank you both! 

Ps. I also updated MPS temporarily, but I'll modify the case for a (better) integration test after this PR.
I do not yet understand why the codecov fails at the line for vintage flow cost, because I can see in the results lp file this term is already printed.

src/objective.jl
```diff
@@ -253,6 +253,46 @@ function add_objective!(connection, model, variables, expressions, model_paramet
...
+            FROM rep_periods_mapping AS rpmap
+            LEFT JOIN rep_periods_data AS rpdata
+                ON rpmap.year=rpdata.year AND rpmap.rep_period=rpdata.rep_period
+            GROUP BY rpmap.year, rpmap.rep_period
+        ) AS rpinfo
+            ON var.milestone_year = rpinfo.year
+            AND var.rep_period = rpinfo.rep_period
+        GROUP BY var.id, t_objective_vintage_flows.weight_for_operation_discounts, rpinfo.weight_sum, rpinfo.resolution,
+            var.time_block_end, var.time_block_start, t_objective_vintage_flows.marginal_cost
+        ORDER BY var.id
```
datejada, line 287: ORDER BY cost in the performance, I would only use it if the order really matters, like the expressions that need chronology (e.g., ramping, storage, etc). I think we can avoid it here, since we aggregate all the costs, and the order doesn't matter in the objective function. @abelsiqueira, what do you think?
```suggestion
```

src/objective.jl
```diff
@@ -253,6 +253,46 @@ function add_objective!(connection, model, variables, expressions, model_paramet
...
+                ANY_VALUE(rpdata.resolution) AS resolution
+            FROM rep_periods_mapping AS rpmap
+            LEFT JOIN rep_periods_data AS rpdata
+                ON rpmap.year=rpdata.year AND rpmap.rep_period=rpdata.rep_period
+            GROUP BY rpmap.year, rpmap.rep_period
+        ) AS rpinfo
+            ON var.milestone_year = rpinfo.year
+            AND var.rep_period = rpinfo.rep_period
+        GROUP BY var.id, t_objective_vintage_flows.weight_for_operation_discounts, rpinfo.weight_sum, rpinfo.resolution,
+            var.time_block_end, var.time_block_start, t_objective_vintage_flows.marginal_cost
```
datejada, line 286: I think this way reads better and only does the groupbys once, so I think it also has a benefit in the performance. I know it is not a pattern we have now, but I am trying to spot places in our SQL that we can have better performance with large volumes of data. @abelsiqueira, what do you think?

```suggestion
        "
         WITH rp_weight AS (
           SELECT year, rep_period, SUM(weight) AS weight_sum
           FROM rep_periods_mapping
           GROUP BY year, rep_period
         ),
         rp_res AS (
           SELECT year, rep_period, ANY_VALUE(resolution) AS resolution
           FROM rep_periods_data
           GROUP BY year, rep_period
         )
         SELECT
           var.id,
           tof.weight_for_operation_discounts
           * rpw.weight_sum
           * rpr.resolution
           * (var.time_block_end - var.time_block_start + 1)
           * tof.marginal_cost AS cost
         FROM var_vintage_flow AS var
         LEFT JOIN t_objective_vintage_flows AS tof
           ON var.from_asset = tof.from_asset
          AND var.to_asset = tof.to_asset
          AND var.milestone_year = tof.milestone_year
          AND var.commission_year = tof.commission_year
         LEFT JOIN rp_weight AS rpw
           ON var.milestone_year = rpw.year
          AND var.rep_period = rpw.rep_period
         LEFT JOIN rp_res AS rpr
           ON var.milestone_year = rpr.year
          AND var.rep_period = rpr.rep_period
```

src/objective.jl
```diff
@@ -493,5 +534,66 @@ function _create_objective_auxiliary_table(connection, constants)
         ",
     )
 
+    DuckDB.execute(
+        connection,
+        "CREATE OR REPLACE TEMP TABLE t_objective_vintage_flows AS
```
datejada, line 538: Related to my suggestion in the SQL query: This pattern reads better. I mean, I select stuff, calculate a bunch of things, and get the info from other tables (i.e., the JOINS) (no groupbys inside the main query).

src/objective.jl
```diff
@@ -222,7 +222,7 @@ function add_objective!(connection, model, variables, expressions, model_paramet
                 * rpinfo.weight_sum
                 * rpinfo.resolution
                 * (var.time_block_end - var.time_block_start + 1)
-                * t_objective_flows.operational_cost
+                * t_objective_flows.marginal_cost
```
datejada, line 225: Would you mind explaining why there is a change in the name?

To be honest, it is confusing for me. At least marginal cost is more commonly associated with the model's dual variables (results) than with the terms in the objective function.

datejada (commented):
@gnawin Thanks for the PR. In general terms, it is looking good. I left some comments in the code to discuss before merging.

src/objective.jl
gnawin: I understand your confusion, so indeed better to change the name.

The original name is `operational_cost`, which is now a computed cost `fuel_cost + flow_milestone.operational_cost`. It has the same name as the parameter `flow_milestone.operational_cost`, so better to rename to differentiate them.

A new name should indicate `fuel_cost + flow_milestone.operational_cost`. It's a computed cost so I don't see it in the INES either. I used the name `marginal_cost` because that's the COMPETES naming convention. 

But you have a valid point, what about `variable_cost` (now that we don't have this name in the schema so no confusion there)?

datejada: Thanks! Yeah, it makes sense. I would propose being even more explicit `total_variable_cost`, since it is the aggregation of fuel cost + operational cost. What do you think?

src/objective.jl
datejada: So, we should keep it for the time being, create an issue to define the impact of this order in general, and add a rule to the developer's guide in the documentation. @abelsiqueira, do you agree? If so, would you mind @gnawin creating the issue? Thanks, both!

src/objective.jl
datejada: Thanks! I agree with the comments, especially on keeping the same SQL formatting (sorry, my bad). I like the idea of the .sql file, as long as we can make the SQL formatter work 😆

gnawin:
Hi @datejada @abelsiqueira thanks for the nice discussions and comments. I addressed all of them.

The only thing left is the codecov hiccup. 
image
Julia Formatter forces such a line break and codecov thinks the first line is not covered.

To fix this, I could do
```
vintage_flows_operational_cost = begin
        @expression(model, sum(row.cost * var_vintage_flow[row.id] for row in indices))
    end
```
but this makes the code unnecessarily more complex. 

What do you think? Shall we do this, or ignore the coverage in this PR, or find other solutions?

PR #1346, opened 2025-09-16
[Multi-year Part II] Validate consistency between `asset_both` and `flow_commission`

### LINKED ISSUE #1345: [Multi-year Part II] Add validation to check consistency between `asset_both` and `flow_commission`
To create capacity constraints for semi-compact method, we need to ensure that when `asset_both` has a certain `commission_year`, that the outgoing flow of that asset has to have the same `commission_year` in `flow_commission`. And vice versa.

Otherwise the outgoing flow will not be generated correctly (due to missing data), and the error message is confusing, from my experience,

gnawin:
This validation ensures when you use `semi-compact` method, the data is available in the `flow_commission` that corresponds to `asset_both`. This is needed to generate capacity constraints.

This validation is similar to #1313 which checks consistency between `asset_both` and `asset_commission` because the model needs to have the fixed costs from the corresponding `commission_year`.

## Related issues

Closes #1345 

## Checklist

- [x] Tests are passing

### DIFF
--- a/src/data-validation.jl
+++ b/src/data-validation.jl
@@ -68,6 +68,11 @@ function validate_data!(connection)
             _validate_asset_commission_and_asset_both_consistency!,
             false,
         ),
+        (
+            "consistency between flow_commission and asset_both",
+            _validate_flow_commission_and_asset_both_consistency!,
+            false,
+        ),
     )
         @timeit to "$log_msg" append!(error_messages, validation_function(connection))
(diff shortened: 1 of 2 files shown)


gnawin:
Shoutout to @abelsiqueira, the @testitem implementation really smoothens the experience for developing tests 😃

datejada (approved):
Hi, Ni. The validation looks good to me. Thanks! It is good to have all these validations for new users.

PR #1361, opened 2025-10-07
Update tutorials

gnawin:
Fix a few things in the tutorials:
 - Correct input/output directories
 - Instruct to manually add 'results' folder
 - Update multi-year tutorials

## Related issues

There is no related issue.

## Checklist

- [ ] Tests are passing

### DIFF
--- a/docs/src/10-tutorials/12-basics.md
+++ b/docs/src/10-tutorials/12-basics.md
@@ -19,7 +19,10 @@ using Plots
 
 !!! tip
     Follow along in this section, copy-pasting into your my_workflow.jl file.\
-    In VS Code, you can highlight each section and press (SHIFT+ENTER) to run the code.\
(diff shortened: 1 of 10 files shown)


datejada (approved):
Thanks for the update, good catch with the `variable_cost` to `operational_cost`.
To avoid this problem in the future, we should not only test that files load correctly, but that they give the same objective function.
Would you mind creating an issue to improve the current test that are in the following link?
https://github.com/TulipaEnergy/TulipaEnergyModel.jl/blob/main/test/test-tutorial-files.jl

gnawin:
@datejada thanks for the comment, I created #1362. 
I pushed another commit to fix the flexible temporal solution tutorial, you can have a look again if you want, otherwise I'll merge it  by end of the day.

PR #1364, opened 2025-10-08
Fix bug discount rate

gnawin:
We use closed-form equations for annuity and salvage value, they are correct but when discount rate = 0, the denominator will become 0 and therefore, result in NaN.

The solution is to either change to a non-closed form where discount rate is no longer in the denominator (if possible), or to add a condition to handle the case when discount rate = 0. This PR adopts the latter approach. 

## Related issues

There is no related issue.

## Checklist

- [x] Tests are passing

### DIFF
--- a/src/objective.jl
+++ b/src/objective.jl
@@ -392,7 +392,7 @@ function _create_objective_auxiliary_table(connection, constants)
     # Same for flows
     DuckDB.execute(
         connection,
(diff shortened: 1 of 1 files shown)


gnawin:
@abelsiqueira thanks for approving! I forgot to do the same for the flows, will check also if anywhere else needs the change.
Will do it tomorrow.

PR #1409, opened 2025-10-28
More robust test of printed EnergyProblem

### LINKED ISSUE #1408: [Bug] Tests are failing because of some update
### Description

The tests are failing because something got updated either in which Julia version is used for testing, or in some package that we depend, and that is causing the solution found in one the tests to be different around the 12th significant digit, which would be fine, except that it causes the printing tests to fail.

### Package Version

latest

### Julia Version

all

### Reproduction steps

See tests on `main`: https://github.com/TulipaEnergy/TulipaEnergyModel.jl/actions/runs/18877928678
These were run manually, so no code changed between the previous test and this one.

### Relevant log output

```shell
Test printing EnergyProblem validation: Test Failed at /home/runner/work/TulipaEnergyModel.jl/TulipaEnergyModel.jl/test/test-io.jl:63
  Expression: split(String(take!(io))) == split(read("io-outputs/energy-problem-model-solved.txt", String))
   Evaluated: SubString{String}["EnergyProblem:", "-", "Model", "created!", "-", "Number", "of", "variables:", "364", "-"  …  "Model", "solved!", "-", "Termination", "status:", "OPTIMAL", "-", "Objective", "value:", "269238.4382415648"] == SubString{String}["EnergyProblem:", "-", "Model", "created!", "-", "Number", "of", "variables:", "364", "-"  …  "Model", "solved!", "-", "Termination", "status:", "OPTIMAL", "-", "Objective", "value:", "269238.4382415647"]
```

### Operating System

_No response_

abelsiqueira:
The objective value at the solution can change slightly because of our dependencies,
so we compare only the first 9 significant digits. Since this is done in the text
level, it looks odd, but it is similar to the case studies.

## Related issues

Closes #1408

## Checklist

- [x] Tests are passing

### DIFF
--- a/test/test-io.jl
+++ b/test/test-io.jl
@@ -60,6 +60,12 @@ end
     io = IOBuffer()
     TulipaEnergyModel.solve_model!(energy_problem)
     print(io, energy_problem)
-    @test split(String(take!(io))) ==
-          split(read("io-outputs/energy-problem-model-solved.txt", String))
+    expected_output = split(read("io-outputs/energy-problem-model-solved.txt", String))
+    actual_output = split(String(take!(io)))
+    # To make the comparison more robust, the only look at the the first
+    # 9 significant digits of the solution (+ '.')
+    expected_output[end] = expected_output[end][1:10]
+    actual_output[end] = actual_output[end][1:10]
+
+    @test expected_output == actual_output
 end


gnawin (approved):
Thanks looks good to me

PR #1410, opened 2025-10-28
Allow commodity_price to be a profile

abelsiqueira:
Allow commodity_price to be a profile.
Add validation between commodity_price in the profiles and in the flow_milestone table.
Test that the `:flows_operational_cost` expression in the objective is correct.

## Related issues

Part of #1269

## Checklist

- [x] Tests are passing

### DIFF
--- a/src/objective.jl
+++ b/src/objective.jl
@@ -256,6 +256,8 @@ function add_objective!(connection, model, variables, expressions, profiles, mod
                 * rp_weight_prob.total_weight_prob
                 * rp_res.resolution
                 AS cost_coefficient,
+            cost_coefficient * obj.total_variable_cost
+                * (var.time_block_end - var.time_block_start + 1) AS total_cost_if_no_profile,
             var.time_block_start,
             var.time_block_end,
             var.year,
@@ -300,10 +302,9 @@ function add_objective!(connection, model, variables, expressions, profiles, mod
                             row.time_block_start:row.time_block_end,
                             Statistics.mean,
                             1.0,
-                        ) / row.producer_efficiency::Float64 +
-                        row.operational_cost::Float64 *
-                        (row.time_block_end - row.time_block_start + 1)
+                        ) / row.producer_efficiency::Float64 + row.operational_cost::Float64
                     ) *
+                    (row.time_block_end - row.time_block_start + 1) *
                     var_flow[row.id::Int64] for row in indices
                 )
             )
@@ -311,13 +312,8 @@ function add_objective!(connection, model, variables, expressions, profiles, mod
             @expression(
                 model,
                 sum(
-                    row.cost_coefficient::Float64 *
-                    (
-                        row.commodity_price::Float64 / row.producer_efficiency::Float64 +
-                        row.operational_cost::Float64 *
-                        (row.time_block_end - row.time_block_start + 1)
-                    ) *
-                    var_flow[row.id::Int64] for row in indices
+                    row.total_cost_if_no_profile::Float64 * var_flow[row.id::Int64] for
+                    row in indices
                 )
             )
         end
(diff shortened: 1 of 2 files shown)


gnawin:
Hi @abelsiqueira, nice you're working on this.

One related issue is #1411, where I'd like to move commodity_price from asset_milestone to flow_milestone. Maybe I'll do it next week, but I'm not sure.

If we merge this PR first, do you think it'll be complicated for me to change it later? If yes, then I'd propose to finish #1411 first and I'll prioritize it.

abelsiqueira:
Hi @gnawin, this is my first time dealing with `commodity_price`, so I don't know what will be more complicated. This is not ready yet, anyway, so feel free to work on #1411.

src/objective.jl
```diff
@@ -257,7 +268,23 @@ function add_objective!(connection, model, variables, expressions, model_paramet
...
+    flows_operational_cost = @expression(
+        model,
+        sum(
+            row.cost_coefficient *
+            (
+                row.commodity_price * _profile_aggregate( # commodity_price aggregation
+                    profiles.rep_period,
+                    (row.profile_name, row.year, row.rep_period),
+                    row.time_block_start:row.time_block_end,
+                    sum,
```
abelsiqueira:
```suggestion
                    mean,
```
cf. @datejada

datejada (commented):
@abelsiqueira thank you for the PR. It looks really good, especially the testing. I have minor comments, see below.

I left a comment in the issue mentioning that we are not going to implement the operational cost profile for now. So, I would say that this PR closes the issue. We can create another one in the future if a use case appears.

I also created an extra issue for a Tutorial for this feature #1464

BTW: I think there is not much we can do with the loss of performance in the benchmark, at least for now. Maybe we need to have a performance hackathon for these things.

src/objective.jl
```diff
@@ -257,7 +268,51 @@ function add_objective!(connection, model, variables, expressions, model_paramet
...
+                sum(
+                    row.cost_coefficient::Float64 *
+                    (
+                        row.commodity_price::Float64 / row.producer_efficiency::Float64 +
+                        row.operational_cost::Float64 *
+                        (row.time_block_end - row.time_block_start + 1)
+                    ) *
+                    var_flow[row.id] for row in indices
+                )
+            )
```
datejada: Repeated computation in else-branch: When there's no profile, the code still fetches all these columns and computes at Julia runtime. The previous code computed this in SQL as total_variable_cost.

Recommended approach: Keep the SQL precomputation for the non-profile case:

Add conditional SQL: if no profile exists, compute total_variable_cost in SQL
This moves computation to DuckDB (optimized native code) instead of Julia loops

src/data-validation.jl
```diff
@@ -596,3 +601,43 @@ function _validate_flow_commission_and_asset_both_consistency!(connection)
...
+                flow_milestone.from_asset,
+                flow_milestone.to_asset,
+                flows_profiles.profile_type,
+                flow_milestone.commodity_price,
+            FROM flows_profiles
+            LEFT JOIN flow_milestone
+                ON flow_milestone.from_asset = flows_profiles.from_asset
+                AND flow_milestone.to_asset = flows_profiles.to_asset
+            WHERE
+                flows_profiles.profile_type = 'commodity_price'
```
datejada: Validation query could be more efficient: The query fetches ALL flows with commodity_price profiles, then filters in Julia. This is fine for small datasets but could be optimized.
```suggestion
            WHERE
                flows_profiles.profile_type = 'commodity_price'
            AND flow_milestone.commodity_price <= 0  -- Only fetch violations
```

src/objective.jl
```diff
@@ -248,6 +254,11 @@ function add_objective!(connection, model, variables, expressions, model_paramet
             AND var.rep_period = rp_res.rep_period
         LEFT JOIN asset
             ON asset.asset = var.from_asset
+        LEFT JOIN flows_profiles AS commodity_price_profiles
+            ON commodity_price_profiles.from_asset = var.from_asset
+            AND commodity_price_profiles.to_asset = var.to_asset
+            AND commodity_price_profiles.year = var.year
+            AND commodity_price_profiles.profile_type = 'commodity_price' -- TODO: inconsistent year naming to assets_profiles
```
datejada: Redundant LEFT JOIN when no profiles exist: This JOIN is always executed, even when no commodity_price profiles exist. For large models, this adds unnecessary overhead.
Recommended approach: Use conditional query building:
- First check if profiles exist (could cache this)
- Then build appropriate query without the JOIN if not needed

PR #1418, opened 2025-11-10
[2S-SO]: Add examples for Two-stage stochastic optimization

### LINKED ISSUE #1397: [2S-SO]: Add example for stochastic optimization case study
### Description

Add an example for stochastic optimization feature

datejada:
Add two new case studies for two-stage stochastic optimization:

- RPs per scenario
- RPs cross scenarios

## Related issues

Closes #1397

## Checklist

- [x] Tests are passing


datejada:
@gnawin, these are the new examples for the two-stage stochastic optimization. For this example, we need to add the stochastic column as the primary key for the representative period mapping table (that is, to guarantee unique values for the combination of columns (:year, :scenario, :period, :rep_period) for this development. The remaining changes should be straightforward, but let me know if you have any comments.

Issue #1298, opened 2025-07-28
[2S-SO]: Update input schema for the new feature


datejada:
### Description

For the two-stage stochastic programming, we must create a new set, Stochastic Scenario (s), and a probability for each stochastic scenario. In addition, the parameters rp_weight and map need to account for the new set.

### Validation and testing

N.A.

### Motivation

They are needed for the new feature :)

### Target audience

developers

### Can you help?

Always!

datejada:
With the latest changes in TulipaClustering, we need to update the schema again, so instead of creating a new issue, I am reusing this one

Issue #1086, opened 2025-03-11
Rename input data to make more sense


abelsiqueira:

gnawin:
Good point. I already see a few places where we have `year` which is not explicit enough.

abelsiqueira:
Do you think it would be useful to try to do this before your PR? I am not touching this at the moment, so no rush.

gnawin:
This is not a big deal to me because since it is just naming changes. Practically it is probably easier to change after my PR :)

PR #1419, opened 2025-11-11
[2S-SO]: Update schema for timeframe profiles information

datejada:
The timeframe profiles also need to include the scenario information. This PR updates the schema and the CSV files.

## Related issues

Part of #1298 

## Checklist

- [x] Tests are passing

### DIFF
--- a/src/input-schemas.json
+++ b/src/input-schemas.json
@@ -479,13 +479,13 @@
         ]
       },
       "description": "Type of profile, used to determine DuckDB table with source profile",
-      "scenario": {
-        "default": 1,
-        "description": "Stochastic scenario number.",
-        "type": "INTEGER",
-        "unit_of_measure": "number"
-      },
       "type": "VARCHAR"
+    },
+    "scenario": {
+      "default": 1,
+      "description": "Stochastic scenario number.",
+      "type": "INTEGER",
+      "unit_of_measure": "number"
     },
     "year": {
       "description": "Milestone year.",


gnawin:
Thanks for the PR, this one gives me a more complete picture so I'm also looking back the past few PRs to get a better understanding. 

My questions are:
1. Accordingly to the docs, the scenario has index (asset, period per milestone year). 
image

So that corresponds to the change in profiles-timeframes.csv.

Do we need the change in the other file? I would imagine that the profile also depends on the commission year, but we tentatively did not do that (at least in the docs, and in the way that compact method was implemented).

2. For the rep periods (related to earlier changes), since the scenario determines the mapping between periods and rep periods, does it mean that the rep periods only depend on the milestone year, regardless of the commission year? I understand that this feature change mainly affects TulipaClustering, i.e., how the profiles are generated, so in general, regardless of the feature development, maybe a more relevant question is: how do we intend to generate profiles for different commission years? Do we use TulipaClustering, and how?

3. In this file https://github.com/TulipaEnergy/TulipaEnergyModel.jl/blob/main/test/inputs/TwoStage-StochOpt%20RPs%20per%20Scenario/rep-periods-data.csv, there is an extra column "scenario" compared to the same file in other cases. Is it supposed to be removed?

datejada:
> Thanks for the PR, this one gives me a more complete picture so I'm also looking back the past few PRs to get a better understanding.
> 
> My questions are:
> 
> 1. Accordingly to the docs, the scenario has index (asset, period per milestone year).
> 
> image
> So that corresponds to the change in profiles-timeframes.csv.
> 
> Do we need the change in the other file? I would imagine that the profile also depends on the commission year, but we tentatively did not do that (at least in the docs, and in the way that compact method was implemented).
> 
> 2. For the rep periods (related to earlier changes), since the scenario determines the mapping between periods and rep periods, does it mean that the rep periods only depend on the milestone year, regardless of the commission year? I understand that this feature change mainly affects TulipaClustering, i.e., how the profiles are generated, so in general, regardless of the feature development, maybe a more relevant question is: how do we intend to generate profiles for different commission years? Do we use TulipaClustering, and how?
> 3. In this file https://github.com/TulipaEnergy/TulipaEnergyModel.jl/blob/main/test/inputs/TwoStage-StochOpt%20RPs%20per%20Scenario/rep-periods-data.csv, there is an extra column "scenario" compared to the same file in other cases. Is it supposed to be removed?

@gnawin thanks for the review. So:

1. Thank you for your observation. I have revised the formulation for clarity. Done ✅
The limits we see are influenced by the milestone year, reflecting operational constraints rather than capacity constraints. In the context of hydro reservoirs, these limits are determined by seasonal storage considerations. For example, during summer, minimum water levels are maintained due to ecological constraints, as draining the reservoir could interfere with its multiple purposes. Conversely, maximum levels need to be carefully managed during winter to prevent flooding.
While the actual capacity of the reservoirs might be higher (which depends on the commission year), these operational parameters should be primarily tied to the milestone year. In addition, they vary based on the scenario, as operational conditions change depending on whether it’s a dry or wet year, a phenomenon that is also visible in hydro-dominated systems. I hope this explanation provides a more straightforward understanding of the parameters. In the future, I plan to include a case study on hydro reservoirs to further elaborate on this subject (but not a priority at the moment).
3. This is a question that goes beyond the changes in this PR. I think it is more for a workflow with multi-year investment (which we don't have yet) 🤓 I think I see why you are asking this, I don't have it 100% clear yet, but I think I have an idea that we all will need to discuss when defining a workflow with multi-year investment.
4. You're right, that is not necessary. I forgot to delete it. Done ✅

gnawin:
@datejada thanks for the detailed explanation and the changes. One last question 😅:

I understand, from your explanation, that these profiles are more relevant for milestone years, not for commission years. So why do you have the scenario index in asset-timeframe-profiles.csv, is it needed?

gnawin (approved):
LGTM! Thanks a lot.

Now that year actually means milestone year, it is at least consistent. I'll try to change the names in the coming weeks. The issue is #1086.

PR #1427, opened 2025-11-17
Move OBZ data to repo instead of downloading from Zenodo

### LINKED ISSUE #1404: Find a better solution for hosting data for test, docs, and benchmark
### Description

Although we have #1225 with the idea of using the Zenodo links for benchmark, sometimes we have download failures that cause the documentation build to fail with the following error:

```plaintext
│   exception =
│    LoadError: RequestError: Operation too slow. Less than 1 bytes/sec transferred the last 20 seconds while requesting https://zenodo.org/api/records/15498101/files/assets-storage-min-max-reservoir-level-profiles.csv/content
```

It is not clear to me whether this is caused by the interaction of GitHub machines with Zenodo, by rate limits being applied, or just Zenodo randomly failing. Either way, scaling the use of Zenodo will make the issue more frequent.

Some alternatives:

- Just keep the data in TEM.jl
- Create a TulipaData repo only to store the data
- Find a different place to host the data

I don't have a strong opinion on the alternatives

### Validation and testing

_No response_

### Motivation

_No response_

### Target audience

_No response_

### Can you help?

_No response_

abelsiqueira:
To avoid relying on Zenodo, we move the data to the repo itself.

## Related issues

Closes #1404

## Checklist

- [x] Tests are passing


abelsiqueira:
The link checker fails because the changes reference a new folder (`docs/src/data/obz`)

PR #1455, opened 2026-01-06
Ignore TulipaEnergyModel version in dependabot

abelsiqueira:
Related to #1453


datejada (approved):
It seems like a good approach. Thanks!

PR #1516, opened 2026-03-10
Add copilot instructions for issues and reviews

datejada:
@abelsiqueira @gnawin I have created this file to start assigning the good-first-issues to Copilot. We now have Copilot licenses at TNO, so we can start using them more to help with minor issues and possibly initial reviews.

I followed the instructions here: https://docs.github.com/en/copilot/how-tos/configure-custom-instructions/add-repository-instructions

I also took inspiration from here: https://github.com/NREL-Sienna/InfrastructureOptimizationModels.jl/blob/main/.claude/Sienna.md

Please feel free to add/modify/delete the instructions as you consider best.

Thanks! 

## Related issues

There is no related issue.

## Checklist

- [N.A.] Tests are passing

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -153,9 +153,10 @@ Docstring requirements:
 Branch naming: `feature/description` or `fix/description`
 
 1. Create feature branch
-2. Follow style guide and run pre-commit
-3. Ensure tests pass
-4. Submit pull request
+2. Follow style guide and run formatter
+3. Follow the style guide and run pre-commit when committing
+4. Ensure tests pass
+5. Submit pull request
 
 **CRITICAL**: When making commits, always add a co-authored line with the tool name, the agent model, and the relevant e-mail. For instance "Co-Authored-By: Claude Code (claude-sonnet-4-6) <noreply@anthropic.com>"
 


abelsiqueira:
Some people split it to avoid adding unnecessary context to the agent, but I don't have any real intuition on how much it matters at the moment. We can start with a single file and split later, if we think it's useful.

AGENTS.md
```diff
@@ -0,0 +1,152 @@
...
+
+#### Unnecessary allocations
+
+- Use views instead of copies (`@view`, `@views`)
+- Pre-allocate arrays instead of `push!` in loops
+- Use in-place operations (functions ending with `!`)
+
+#### Captured variables
+
+Avoid closures that capture variables causing boxing. Pass variables as function arguments instead.
```
abelsiqueira: I don't know if this is clear enough for the AI, because it's not clear to me. The description must be very correct, and my understanding is "When creating anonymous functions inside a local scope, don't use variables from that local scope". I think that's easier to understand, at least.

abelsiqueira (changes requested):
Thanks @datejada. I checked this and used claude to compare with CLAUDE.md from BestieTemplate.jl and with the developer documentation. I selected what I think were the most relevant modifications and made the suggestions below.

I don't know how strict you want to be here. Should we focus on getting something good to start and update it later, or do you want to refine as much as possible for a few days? I focused on the first.

AGENTS.md
```diff
@@ -0,0 +1,200 @@
...
+- Fail fast with actionable error messages rather than hiding problems
+- Validate invariants explicitly in subtle cases
+- Avoid over-adherence to backwards compatibility for internal helpers
+
+## Contribution Workflow
+
+Branch naming: `feature/description` or `fix/description`
+
+1. Create feature branch
+2. Follow style guide and run formatter
```
gnawin: We should ask the agent to do pre-commit instead of only the formatter, if possible.

datejada:
@gnawin thanks for the review. So, if I understand correctly, you want to add guidelines to the developers' documentation and maybe also to the README for people who want to contribute, to explain how and when we would like to use agents. Is that correct? 
If so, I think it is a good idea. I will add it in the next commit of this PR 🤓. Please add any extra comments if I am missing anything.

PR #1533, opened 2026-03-13
Change in energy to power ratio method for energy storage assets

### LINKED ISSUE #1446: [Bug] The parameter `energy_to_power_ratio` is no longer used in the code
### Description

The parameter `energy_to_power_ratio` was used in the code for energy storage investments, where a fixed ratio determined the storage component. For example `energy_to_power_ratio=4`, investing in 4h batteries means that for each 1MW invested, you get 4MWh of storage capacity. However, the current version (and probably since before), this is no longer the case. The current version only supports the case where we optimize the storage capacity (which is also a use case/method we want to have in Tulipa). The docs still have the old explanation of both methods, but only the first one works at the moment.

https://tulipaenergy.github.io/TulipaEnergyModel.jl/stable/20-user-guide/20-how-to-use/#storage-investment-setup

I recognize that the explanation in the docs might not have been straightforward to show how to make the method work (since there was an implicit assumption that `storage_method_energy = false` is the use of this method). So, we might need to redefine this method to have three options: `none`, `optimize_storage_capacity`, `use_fixed_energy_to_power_ratio`; then it is explicit which method one is setting up. Such that:

- `none`: is equivalent to the current behaviour with false (no `var_assets_investment_energy` variables)
- `optimize_storage_capacity`: current functionality where `var_assets_investment_energy` are created.
- `use_fixed_energy_to_power_ratio`: in this case `var_assets_investment_energy = energy_to_power_ratio * var_assets_investment`, so either an extra constraint is added, but it might be more efficient to use an expression to avoid creating extra variables. This might me tricky with the multiyear investment. So, we need to double-check the best way to do it.

@gnawin please give your thoughts on this, since I think this functionality got lost during the multi-year refactor, and also this should be compatible with the `simple method`, so you might have more suggestions or ideas to make fix this bug.

### Package Version

v0.19.0

### Julia Version

1.12

### Reproduction steps

search for the parameter `energy_to_power_ratio` in the code, in it is no longer use in the constraints.

### Relevant log output

```shell

```

### Operating System

_No response_

datejada:
The changes read more easily by commit:

1. [Update input schema](https://github.com/TulipaEnergy/TulipaEnergyModel.jl/commit/e2ae12b40bca585540f7da95952395285fa4922b)

1. [Update files with new schema](https://github.com/TulipaEnergy/TulipaEnergyModel.jl/commit/c21f35ea934ec7b6af337b9aadf4b08503527c7e)

1. [Update storage expressions](https://github.com/TulipaEnergy/TulipaEnergyModel.jl/commit/a8cc5e3e6ec1230e7388e71cd47b4aa3b4827928)

1. [Update storage investment variable creation](https://github.com/TulipaEnergy/TulipaEnergyModel.jl/commit/6be0cf18e086a3601773d2425f566f1591ae9e4d)

1. [Update docs information](https://github.com/TulipaEnergy/TulipaEnergyModel.jl/commit/8f99d443f7a25a3e8cf51368e9ccb3540e19ca3d)

1. [Update MPS file for Norse (intended)](https://github.com/TulipaEnergy/TulipaEnergyModel.jl/commit/ca604049cdb30bebe715ba7d7d6173ea747a39b1)

1. [Add tests for min/max constraints](https://github.com/TulipaEnergy/TulipaEnergyModel.jl/pull/1533/commits/040eca8b556f670cc4ff21fe63cef7a521de3f98) 

## Related issues

Closes #1446 

## Checklist

- [x] Tests are passing

### DIFF
--- a/test/test-constraint-storage-min-max-level.jl
+++ b/test/test-constraint-storage-min-max-level.jl
@@ -1,10 +1,10 @@
-@testsnippet StorageMinMaxLevelSetup begin
+@testsnippet ConsStorageMinMaxLevelSetup begin
     using DuckDB: DuckDB
     using TulipaBuilder: TulipaBuilder as TB
     using TulipaClustering: TulipaClustering as TC
 
-    # Storage configuration struct for testing
-    @kwdef struct StorageMinMaxLevelConfig
+    # Configuration struct for testing
+    @kwdef struct ConsStorageMinMaxLevelConfig
         is_seasonal::Bool
         storage_method_energy::String
         name::String = "dummy_storage"
@@ -20,22 +20,17 @@
         min_storage_level_profile::Dict{Tuple{Int,Int},Vector{Float64}} =
             Dict((2030, 1) => [0.1, 0.4, 0.0])
         inflows_profile::Dict{Tuple{Int,Int},Vector{Float64}} = Dict((2030, 1) => [0.3, 0.7, 0.2])
+        num_timesteps::Int = 1
+        num_rps::Int = 2
     end
 
     """
-        create_storage_min_max_level_test_problem(
-            storage_asset;
-            num_timesteps, num_rps
-        )
+        create_storage_min_max_level_test_problem(storage_asset)
 
     Create a storage min-max level test problem with storage asset configuration.
     Returns the database connection with configured storage asset and clustering.
     """
-    function create_storage_min_max_level_test_problem(
-        storage_asset::StorageMinMaxLevelConfig;
-        num_timesteps::Int = 1,
-        num_rps::Int = 2,
-    )
+    function create_storage_min_max_level_test_problem(storage_asset::ConsStorageMinMaxLevelConfig)
         tulipa = TB.TulipaData()
 
         # Add basic producer and consumer to connect the storage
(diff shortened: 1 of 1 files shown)


datejada:
Removing benchmark run since there is a change in the schema, it can compare the branches 😞

gnawin (approved):
Great, thanks. This is good to me now :)

test/test-constraint-storage-min-max-level.jl
```diff
@@ -0,0 +1,870 @@
...
+        initial_units::Float64
+        initial_storage_units::Float64
+        capacity::Float64
+        capacity_storage_energy::Float64
+        storage_method_energy::String
+        energy_to_power_ratio::Float64
+        investable::Bool
+        investment_method::String
+        investment_integer_storage_energy::Bool
+    end
```
abelsiqueira: I don't like this here. In the other test, it was debatable, but here it is not being used properly, so it's probably not as necessary. Depending on what you need, a `NamedTuple` is a good substitute.

test/test-constraint-storage-min-max-level.jl
```diff
@@ -0,0 +1,585 @@
+@testsnippet StorageMinMaxLevelSetup begin
+    using DuckDB: DuckDB
+    using TulipaBuilder: TulipaBuilder as TB
+    using TulipaClustering: TulipaClustering as TC
+
+    # Storage configuration struct for testing
+    @kwdef struct StorageMinMaxLevelConfig
```
abelsiqueira:
```suggestion
    @kwdef struct ConsStorageMinMaxLevelConfig
```

PR #1535, opened 2026-03-16
Add variable Value at Risk threshold mu and tests

### LINKED ISSUE #1525: [CVaR]: Add VaR threshold variable
### Description

- Create the variable VaR threshold variable ($\mu$) at confidence level $\alpha$. Use a descriptive name `value_at_risk_threshold_mu`
  - Add a new file for this variable here https://github.com/TulipaEnergy/TulipaEnergyModel.jl/tree/main/src/variables
  - Double-check if we actually need to add the corresponding SQL query: https://github.com/TulipaEnergy/TulipaEnergyModel.jl/blob/main/src/sql/create-variables.sql
  - Only create the new variable if the value of the $\lambda$ parameter (see #1524) is greater than zero
- The variable is continuous and positive (double check for upper bound)
- It doesn't depend on any index, and the units are the currency of the objective function
- Add a test that checks that the variable is created when the $\lambda$ parameter is greater than zero

florinelaseur:
Add variable Value at Risk threshold mu and tests

Closes #1525 

## Checklist

- [ ] Tests are passing

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -273,7 +273,7 @@ function _create_filter(args)
             filters,
             test_item ->
                 contains(test_item.name, args.pattern) ||
-                contains(test_item.filename, args.pattern),
+                    contains(test_item.filename, args.pattern),
         )
     end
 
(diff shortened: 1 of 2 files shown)


src/sql/create-variables.sql
```diff
@@ -499,3 +499,18 @@ where
...
+;
+
+create table var_value_at_risk_threshold_mu as
+select
+    nextval('id') as id,
+    cast(null as float8) as solution,
+;
+
+drop sequence id
+;
```
datejada:
```suggestion
```

PR #1537, opened 2026-03-16
Add objective cost breakdown to existing components in the objective

### LINKED ISSUE #1389: Output breakdown of the objective value
### Description

Currently TEM only outputs the objective value of the problem, it is not straightforward to have a breakdown of the objective values. 

For example, what are the investment costs, operational costs, and flow costs? As a quick fix, it is doable by going back to the definition of the objective function, the get the breakdown per component, as defined already, such as investment costs, operational costs, and flow costs. But it still requires developer level knowledge.

More detailed breakdowns are trickier, when you need a breakdown beyond what's an existing component of the objective function. For example, what is the production of a certain asset?

I propose to store the breakdown somewhere, so users can get it easier. Maybe a new structure?
Open to suggestions.

### Validation and testing

N/A

### Motivation

This issue came out from the multi-year case study where we wanted to analyze the problem further by looking at different cost components, but this is applicable to all investment cases.

### Target audience

Users

### Can you help?

Sure

gnawin:
## Related issues

Closes #1389 partly

- [X] create a new table obj_breakdown with the name of the expressions that are attached to the objective, where one of the columns is filled by save_solution! with the final value of the expression.
- [X] in src/objective.jl, attach the expressions to the objective using a new function. This function creates a new entry in obj_breakdown table in addition to what's existing.
- [X] a test of the breakdown.

## Checklist

- [x] Tests are passing

### DIFF
--- a/src/solve-model.jl
+++ b/src/solve-model.jl
@@ -111,10 +111,8 @@ function save_solution!(connection, model, variables, constraints; compute_duals
             error("Table obj_breakdown has no corresponding objective in model")
         end
     end
-    breakdown_table = [
-        (name = name, value = JuMP.value(model[Symbol(name)])) for
-        name in breakdown_names
-    ]
+    breakdown_table =
+        [(name = name, value = JuMP.value(model[Symbol(name)])) for name in breakdown_names]
     DuckDB.register_table(connection, breakdown_table, "t_obj_breakdown_solution")
     DuckDB.execute(
         connection,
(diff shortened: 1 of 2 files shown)


test/test-objective-breakdown.jl
```diff
@@ -0,0 +1,19 @@
...
+    TulipaEnergyModel.create_model!(energy_problem)
+    TulipaEnergyModel.solve_model!(energy_problem)
+    TulipaEnergyModel.save_solution!(energy_problem)
+
+    @test length(collect(DuckDB.query(connection, "SELECT name FROM obj_breakdown"))) == 10
+    for row in DuckDB.query(connection, "SELECT name, value FROM obj_breakdown")
+        @test !ismissing(row.value)
+    end
+    # Sum of components ≈ total objective value
+    total = get_single_element_from_query_and_ensure_its_only_one(
```
abelsiqueira:
```suggestion
    total = TulipaEnergyModel.get_single_element_from_query_and_ensure_its_only_one(
```

PR #1552, opened 2026-03-25
Remove year_data table

### LINKED ISSUE #1356: Rename or remove year-data
### Description

This table is only used to define the default discount year, by finding the minimum of the milestone years. All the rest columns are redundant.

So it does not need to be a table, maybe we should find another way to supply this information, e.g., putting the milestone years in model-parameters? Or leave only 1 column in this table? 

It could also be an idea that we have milestone years implicitly, e.g., in other tables. On one hand, it reduces the information we need to supply. On the other hand, it may be confusing because too much info is in a certain table.

gnawin:
## Related issues

Closes #1356

## Checklist

- [x] Tests are passing

### DIFF
--- a/src/data-validation.jl
+++ b/src/data-validation.jl
@@ -769,7 +769,9 @@ function _validate_bid_related_data!(connection)
             DuckDB.query(connection, "SELECT COUNT(DISTINCT milestone_year) FROM rep_periods_data"),
         ),
     )
-    num_rep_periods = count_rows_from(connection, "rep_periods_data")
+    num_rep_periods = only(
+        only(DuckDB.query(connection, "SELECT COUNT(DISTINCT rep_period) FROM rep_periods_data")),
+    )
 
     for (justification, condition_dict) in (
         ("consumer with unit commitment", consumers_with_unit_commitment),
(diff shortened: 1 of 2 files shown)


docs/src/20-user-guide/20-how-to-use.md
```diff
@@ -398,7 +388,7 @@ We have 3 profiles for `wind` commissioned in 2020, 2030, and 2050, respectively
...
+- [optional] `discount_year` and `discount_rate` in the `model_parameters` table (for CSV input, in `model-parameters.csv`): model-wide discount year and rate. By default, the model will use a discount rate of 0. The `discount_year` is calculated in the code as the minimum value between the input value in the table `model_parameters` and the minimum of all milestone years. In other words, the costs will be discounted to the cost of the first milestone year is the user don't provide a specific value or if the input value is greater than the minimum milestone year.
```
abelsiqueira:
```suggestion
- [optional] `discount_year` and `discount_rate` in the `model_parameters` table (for CSV input, in `model-parameters.csv`): model-wide discount year and rate. By default, the model will use a discount rate of 0. The `discount_year` is calculated in the code as the minimum value between the input value in the table `model_parameters` and the minimum of all milestone years (i.e. years in the `rep_periods_data` table). In other words, the costs will be discounted to the cost of the first milestone year is the user don't provide a specific value or if the input value is greater than the minimum milestone year.
```

abelsiqueira (commented):
Thanks @gnawin, I suggested a fix for the validation test

PR #1560, opened 2026-03-30
Casestudycvar

florinelaseur:
WIP casestudycvar
-does not find value_at_risk_threshold_mu (in lines 139-140 or 144-145) (errors are that key does not exist and  "var_value_at_risk_threshold_mu" not found). DuckDB queries do not have to be inserted for variables right? What is the correct way to call the optimal VaR both for the benchmark in this block (and I will repeat this later in main for each case study -and their 6 different numbers of rps - such that I can add the column to results_df?) 

Example-tulipa-single-year with 3 scenarios is chosen as input data to reduce the computational complexity for first (not realistic for CVaR) test results. 
Thank you!! 

UPDATE 
Hi Diego Thank you very much! I performed the steps. I cannot seem to create a new PR on the same branch (as this one is still open?) but in the current version of Tulipa, The line time_to_read = @elapsed energy_problem_benchmark = TEM.EnergyProblem(connection_benchmark) gives "Error: No implicit schema for table named model_paramets". moving it to different folders then than the input_data_path did not work and it seems the only way to correctly read the 2 parameters for the benchmark, as we dont want them to be read from casestudy files. Could you please look at my final commit "Implicit schema for table named model_parameters"? Thanks! Cheers,

Florine

## Checklist

- [ ] Tests are passing


datejada:
replaced by: https://github.com/[florinelaseur/Experiments_CVaR](https://github.com/florinelaseur/Experiments_CVaR)

PR #1567, opened 2026-04-03
Remove ModelParameters struct

### LINKED ISSUE #1554: Remove ModelParameter structure
### Description

The ModelParameter structure has two purpose:
- have default values
- compute the `discount_year` with a logic

After #1542 , it becomes redundant in a few places, e.g., `add_conditional_value_at_risk_variables!`, where it is directly available from the DuckDB connection, without needing to have a separate argument for the function.

The logic for computing the `discount_year` should be explicit in the places where it is the needed.

I propose to remove this structure. Ideas? @abelsiqueira @datejada

gnawin:
## Related issues

Closes #1554

## Checklist

- [x] Tests are passing


gnawin:
@abelsiqueira This PR is more complicated than I thought, or maybe I missed an important piece from our discussion in-person. 😅 

The main problem is that I thought model_parameters table can be missing, and it is safe to call populate_with_defaults!, then we just need to manually populate it with _create_model_parameters_unless_exists!. However, I found out that missing this table will actually break populate_with_defaults!. 

Other mandatory tables have primary keys. Model_parameters is mandatory but has no primary keys, and hence the special treatment in this commit https://github.com/TulipaEnergy/TulipaEnergyModel.jl/pull/1567/changes/836d16dc58ac975314e1780db7a8c165d4551e28.

Another solution is to actually ALLOW this table to be missing. I'll need to modify _create_model_parameters_unless_exists!, but not populate_with_defaults!. This makes more sense since we're already fixing this table using _create_model_parameters_unless_exists!. And it is less ugly.

Let me know what you think. Thanks!

test/test-model-parameters.jl
```diff
@@ -57,75 +57,82 @@
...
     connection = connection_with_norse()
-    create_model_parameters_table!(
-        connection;
-        discount_rate = 0.03,
-        discount_year = 2020,
-        power_system_base = 90.0,
-        risk_aversion_confidence_level_alpha = 0.90,
-        risk_aversion_weight_lambda = 0.1,
-    )
+    TulipaEnergyModel.populate_with_defaults!(connection)
```
abelsiqueira, line 59: To make sure table is missing
```suggestion
    @test !_check_table_exists(connection, "model_parameters")
    TulipaEnergyModel.populate_with_defaults!(connection)
```

test/test-model-parameters.jl
```diff
@@ -57,75 +57,82 @@
...
+    @test mp.discount_year == 2030
+    # all the other parameters should be filled with default values
+    @test mp.discount_rate == 0.0
+    @test mp.power_system_base == 100.0
+    @test mp.risk_aversion_confidence_level_alpha == 0.95
+    @test mp.risk_aversion_weight_lambda == 0.0
 end
 
-@testitem "Test model parameters - uses default values when table is missing" setup =
+@testitem "Test model parameters - table exists but only has headers" setup =
```
abelsiqueira:
```suggestion
@testitem "Test model parameters - empty table, i.e., exists but only has headers" setup =
```

test/test-model-parameters.jl
```diff
@@ -57,75 +57,82 @@
...
-
-@testitem "Test model parameters - clamps discount_year to earliest milestone year" setup =
+@testitem "Test model parameters - if discount_year is missing" setup =
     [CommonSetup, ModelParametersSetup] tags = [:unit, :validation, :fast] begin
     connection = connection_with_norse()
     create_model_parameters_table!(
         connection;
         discount_rate = 0.03,
-        discount_year = 2040,
+        discount_year = missing,
```
abelsiqueira: Does this work?
```suggestion
```

test/test-model-parameters.jl
```diff
@@ -134,65 +141,26 @@ end
...
     connection = connection_with_norse()
     create_model_parameters_table!(
         connection;
         discount_rate = missing,
-        discount_year = 2040,
-        power_system_base = missing,
-        risk_aversion_confidence_level_alpha = missing,
+        discount_year = 2025,
+        power_system_base = 80.0,
+        risk_aversion_confidence_level_alpha = 0.8,
```
abelsiqueira: Can we also leave this one out, so it's a different use case with a non-zero default. Shouldn't make any difference in the end, but I'm superstitious 👻 

```suggestion
```

test/test-model-parameters.jl
```diff
@@ -134,65 +141,26 @@ end
...
+    mp = query_model_parameters(connection)
 
     @test mp.discount_rate == 0.0
-    @test mp.discount_year == 2030
-    @test mp.power_system_base == 100.0
-    @test mp.risk_aversion_confidence_level_alpha == 0.95
+    # discount_year should be the value given in the table, since it's not missing
+    @test mp.discount_year == 2025
+    @test mp.power_system_base == 80.0
+    @test mp.risk_aversion_confidence_level_alpha == 0.8
```
abelsiqueira: If we leave it out
```suggestion
    @test mp.risk_aversion_confidence_level_alpha == 0.95
```

test/test-model-parameters.jl
```diff
@@ -134,65 +141,26 @@ end
...
     @test mp.risk_aversion_weight_lambda == 0.2
 end
```
abelsiqueira, line 190: It would be nice to have a test with more than one year, just to check that the `min` is computed.

src/data-preparation.jl
```diff
@@ -171,6 +175,20 @@ function populate_with_defaults!(connection)
...
+            end
+            # we only update discount_year when it's NULL, otherwise we might overwrite a user-provided value
+            DuckDB.execute(
+                connection,
+                "UPDATE $table_name SET discount_year = (
+                    SELECT MIN(milestone_year) FROM rep_periods_data
+                ) WHERE discount_year IS NULL",
+            )
+        end
     end
```
abelsiqueira, line 194: I was going to try it out before suggesting, but I won't have time right now.
The suggestion would be to try to have the fix outside the loop, but it's not straightforward, so I don't know if it is possible without making the code worse.
We can check it out later.

abelsiqueira (commented):
Thanks, Ni. It looks much easier to follow now. I've left a few comments for now. The performance looks good to me

test/test-model-parameters.jl
gnawin: Done

test/test-model-parameters.jl
gnawin: Done

test/test-model-parameters.jl
gnawin: Yes, it is the default

test/test-model-parameters.jl
gnawin: Done :)

test/test-model-parameters.jl
gnawin: Exactly

test/test-model-parameters.jl
gnawin: Done. I used the _multi_year_fixture()

src/data-preparation.jl
gnawin: I tried here https://github.com/TulipaEnergy/TulipaEnergyModel.jl/pull/1567/changes/d462a80af2fb2715765a51caa5c9e2bc25cc4f8f :)

gnawin:
Thanks for the review @abelsiqueira. I updated the PR following your suggestions and a refactor of `populate_with_defaults!`. Let me know what you think now. Thanks!

src/input-schemas.json
```diff
@@ -849,8 +849,8 @@
       "type": "DOUBLE"
     },
     "discount_year": {
-      "default": 9999,
-      "description": "The model discount year. The `discount_year` is calculated in the code as the minimum value between the input value in the table `model_parameters` and the minimum of all milestone years (in the table `rep_periods_data`).",
+      "default": null,
```
abelsiqueira: I thought this would have no default. Does it work without setting a default? Otherwise it sets a wrong expectation for non-populate_with_default users, because for normal Tulipa usage, it should be required

abelsiqueira (approved):
Hi @gnawin, I did a small review of the things I already saw. It's not really easy to follow all the changes, so I only looked at the comments.
I left a comment about the default of `discount_year`, because it shouldn't have a default. I hope it's an easy fix, because otherwise it's approved.

src/data-preparation.jl
gnawin, line 98: @abelsiqueira if discount_year does not have a default, it will error here once `_sql_arguments_for_defaults(connection, table_name, table_schema)` is called in line 186 below.
(EDIT: it will not error in every case, but in some cases).

Hence removing the default is possible, but some extra treatment for this column needs to be done in `_populate_model_parameters_with_defaults!`. I added a default of `NULL` only to avoid this special treatment.

That said, if having this default is misleading, I can add this special treatment. It is fine, since we already do different things to model-parameters anyway.

Let me know what you think.

gnawin:
Thanks @abelsiqueira for checking. I understand it is taking long and also the changes are massive in this case. I left a comment to address your question, let me know what you think then I can take actions accordingly.

src/input-schemas.json
gnawin: See comment https://github.com/TulipaEnergy/TulipaEnergyModel.jl/pull/1567#discussion_r3079694291

src/data-preparation.jl
abelsiqueira: I think we have to do this to ensure `discount_year` is a required field. Technically, this only becomes relevant once we have other ways of creating Tulipa problems, which is related to our assumption that `populate_with_defaults!` is not mandatory.

gnawin:
@abelsiqueira I removed the default for `discount_year` with the following extra changes:

- `src/data-preparation.jl`: `_populate_model_parameters_with_defaults!` now filters `discount_year` out of the schema-driven table build, then ensures `discount_year` exists. Its value is computed later.
- `src/data-validation.jl`: `_validate_model_parameters_discount_year!` now checks if `discount_year` is missing when `populate_with_defaults!` wasn't called.
- `test/test-model-parameters.jl` and `test/test-data-validation.jl` covering: missing value, absent column, empty table without the column, and "don't overwrite fully-valid input".

I think this covers what we discussed. You can check the latest two commits, or only the comments inside. Let me know 😉 !

And I'd need to revive the commented out snippets in `benchmarks.jl` before merging.

abelsiqueira (approved):
Thanks for the update, @gnawin

PR #1574, opened 2026-04-09
[Validation] Add validator for investable and investment_method consistency

### LINKED ISSUE #1249: [Validation] consistency between `investable` and  investment method `none`
### Description

Given #1247 (fixed by #1248), it can be confusing what it means to the user or even what the user really wants in the model, when investment method is set to `none`, but `investable` or `decommission` is set to `true`.

We should add a validation to make sure that does not happen, and therefore, no confusion in the data.

abelsiqueira:
An asset should be investable in asset_milestone if and only if its investment_method in asset is not 'none'. This validator enforces both directions of that constraint.

Co-Authored-By: Claude Code (claude-sonnet-4-6) <noreply@anthropic.com>

## Related issues

Closes #1249 

## Checklist

- [x] Tests are passing

### DIFF
--- a/src/data-validation.jl
+++ b/src/data-validation.jl
@@ -85,6 +85,11 @@ function validate_data!(connection)
             _validate_commodity_price_consistency!,
             false,
         ),
+        (
+            "consistency between investable and investment_method",
+            _validate_investable_and_investment_method_consistency!,
+            false,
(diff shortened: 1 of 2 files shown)


datejada:
No longer needed after #1596

Issue #1523, opened 2026-03-12
[CVaR]: Conditional Value-at-Risk Feature


datejada:
### Description

Implementation of the Conditional Value-at-Risk (CVaR) in Blended Representative Periods-Based Energy Planning in Tulipa.

Hold on tight! 🚀

Issue #1335, opened 2025-09-09
Define a rule when tables need to be ordered or not


gnawin:
### Summary
Currently there is no specific rule for a table to be ordered or not, it is implemented on an ad-hoc basis (usually with good reasons). We need to review the tables and define an explicit rule for that.

Below are some of the examples/rules
- due to chronological orders (e.g., ramping, storage),
- due to tests (i.e., multi-year expressions), 
- due to `zip`ping the rows with the other containers (i.e., in the obj)

### Original discussion
This is a good performance check. On one hand, there is the cost of ordering, and in the other there is the cost of indexing by `row.id`. Benchmarking this change would be useful to know for sure (maybe something for the blogpost?). The problem here is that we expect some tables to be ordered, but not others. It's a possible pitfall if we allow the mixup. The ordering reasoning here is the same in other places (`zip`ping the rows with the other containers), so my initial intuition would be ordering as well. 
I think we should try to define a rule for when the tables NEED to be ordered or not.

_Originally posted by @abelsiqueira in https://github.com/TulipaEnergy/TulipaEnergyModel.jl/pull/1332#discussion_r2333384986_

PR #1575, opened 2026-04-09
Add CVaR scenario tail excess constraints

### LINKED ISSUE #1528: [CVaR]: Add tail excess constraint
### Description

- Add constraint for tail excess for s ∈ S. Use descriptive name `tail_excess`
  - Add a new file for this constraint here: https://github.com/TulipaEnergy/TulipaEnergyModel.jl/tree/main/src/constraints
  - Add the corresponding SQL query: https://github.com/TulipaEnergy/TulipaEnergyModel.jl/blob/main/src/sql/create-constraints.sql
  - Only create the new constraints if the value of the $\lambda$ parameter (see #1524) is greater than zero
- It depends on the scenario index, and the units are the currency of the objective function
- Add a test that checks that the constraints are created when the $\lambda$ parameter is greater than zero (test for one and for more than one scenarios)

datejada:
The main changes include adding new expressions and constraints for scenario tail excess, integrating these into the model creation workflow, and refactoring some objective function components for clarity and consistency.

The changes read better by commit: 

- [Update obj. func expression registration](https://github.com/TulipaEnergy/TulipaEnergyModel.jl/commit/a3ba5f498beb003ac71a1250171a007efcb90dfb)

- [Enforce scenario ordering in CVaR variable and constraint tables](https://github.com/TulipaEnergy/TulipaEnergyModel.jl/commit/fb3d09f81ddbbb70766e81438b4ec07face81043) 

- [Extract per-scenario operational cost expressions from objective](https://github.com/TulipaEnergy/TulipaEnergyModel.jl/commit/eab9a64e08f2fc159b3c25f4dca46a93f9239ce4) 

- [Add CVaR tail excess expressions and constraints](https://github.com/TulipaEnergy/TulipaEnergyModel.jl/commit/dd6485847aefddec31f1ef07db9e3a6a35f75c3b) 

Co-authored-by: GitHub Copilot <github-copilot[bot]@users.noreply.github.com>

## Related issues

Closes #1528 

## Checklist

- [x] Tests are passing


datejada:
@abelsiqueira, this PR is a bit big, so I would appreciate your review. I co-developed it using Copilot and following the AGENTS.md. After a couple of iterations, I am happy with this version for your review.

EDIT: ~I still want to add a constraint test using TulipaBuilder, but in the meantime, the current tests cover the basic ones for that constraint. So, you can start the review while I work on it.~ Done ✅ 

Thanks!

P.S. the Julia MCP is amazing 🤓

datejada:
@abelsiqueira I was expecting the PR to be slower with these changes in the benchmarks, but I think the slowdown is not significant. What do you think?

src/objectives/create.jl
```diff
@@ -1,15 +1,13 @@
...
+    _add_to_objective!(connection, objective_expr, name, expr)
 
-Add `expr` to the running objective sum, register it on the model under `name`, and
-insert a placeholder row in the `obj_breakdown` table (value filled in by
-`save_solution!` after the solve).
+Add `expr` to the running objective sum  and insert a placeholder row
+in the `obj_breakdown` table (value filled in by `save_solution!` after the solve).
 """
-function _add_to_objective!(connection, model, objective_expr, name::String, expr)
+function _add_to_objective!(connection, objective_expr, name::String, expr)
```
abelsiqueira, line 7: I'm reading the commits one at a time. I don't know why this commit was made, i.e., why remove `model` from `_add_to_objective!` and why change how the expressions were defined

src/sql/create-variables.sql
```diff
@@ -539,13 +539,22 @@ drop table if exists var_tail_excess_slack_xi
...
 create table var_tail_excess_slack_xi as
+with ordered_scenarios as (
+    select
+        scenario,
+        probability
+    from
+        stochastic_scenario
+    order by
+        scenario
+)
```
abelsiqueira, line 588: I'm not really opposed, just wondering if you have a reason to order by scenario

test/test-objective-operational-costs-per-scenario.jl
```diff
@@ -0,0 +1,50 @@
...
+        expr_rows = DuckDB.query(
+            connection,
+            """
+            SELECT
+                scenario,
+                probability
+            FROM $(expr.table_name)
+            ORDER BY id
+            """,
+        ) |> collect
```
abelsiqueira: This could be a DataFrame (it's already loaded in the tests). It would make the tests easier to follow

test/test-objective-operational-costs-per-scenario.jl
```diff
@@ -0,0 +1,50 @@
...
+    scenario_rows = DuckDB.query(
+        connection,
+        """
+        SELECT
+            scenario,
+            probability
+        FROM stochastic_scenario
+        ORDER BY scenario
+        """,
+    ) |> collect
```
abelsiqueira: As well

src/expressions/operational-costs.jl
```diff
@@ -0,0 +1,375 @@
...
+        return Dict{Int32,UnitRange{Int}}()
+    end
+
+    scenario_ids = [row.scenario for row in indices]
+    group_starts = [1; [i for i in 2:n if scenario_ids[i] != scenario_ids[i-1]]]
+    group_ends = [group_starts[2:end] .- 1; n]
+    scenarios = @view scenario_ids[group_starts]
+
+    return Dict(scenarios[k] => group_starts[k]:group_ends[k] for k in eachindex(scenarios))
+end
```
abelsiqueira: This needs a docstring explaining what it does and why we need it

src/expressions/operational-costs.jl
```diff
@@ -0,0 +1,375 @@
...
+            )
+            costs[i] =
+                row.cost_coefficient::Float64 *
+                (
+                    row.commodity_price::Float64 * commodity_price_agg /
+                    row.producer_efficiency::Float64 + row.operational_cost::Float64
+                ) *
+                (row.time_block_end - row.time_block_start + 1)
+        end
+    end
```
abelsiqueira: what's happening here?

src/expressions/operational-costs.jl
abelsiqueira: I'll have to recheck this section to understand it after the docstring and the other comment above

src/constraints/conditional_value_at_risk.jl
```diff
@@ -0,0 +1,61 @@
...
+)
+    let table_name = :scenario_tail_excess, cons = constraints[table_name]
+        indices = _append_scenario_tail_excess_data_to_indices(connection, table_name)
+        if isempty(variables[:tail_excess_slack_xi].container)
+            return nothing
+        end
+        var_tail_excess_slack_xi = variables[:tail_excess_slack_xi].container
+        total_cost_per_scenario =
+            expressions[:scenario_tail_excess].expressions[:total_cost_per_scenario]
+        var_value_at_risk_threshold_mu = variables[:value_at_risk_threshold_mu].container[1]
```
abelsiqueira, line 36: Use `only` instead to be clear that it should be only 1

src/expressions/conditional_value_at_risk.jl
```diff
@@ -0,0 +1,73 @@
...
+        attach_expression!(
+            expressions[:scenario_tail_excess],
+            :total_cost_per_scenario,
+            JuMP.AffExpr[],
+        )
+        return nothing
+    end
+
+    let table_name = :scenario_tail_excess, expr = expressions[table_name]
+        n = expr.num_rows
```
abelsiqueira, line 32: I'd rather be explicit below

test/test-constraint-scenario-tail-excess.jl
```diff
@@ -0,0 +1,112 @@
...
+    DuckDB.query(
+        connection,
+        """
+        CREATE OR REPLACE TABLE model_parameters AS
+        SELECT *
+        FROM (VALUES
+        	(0.1, 0.98)
+        ) AS t(risk_aversion_weight_lambda, risk_aversion_confidence_level_alpha);
+        """,
+    )
```
abelsiqueira: There is a function to create tables in the tests, I think. If so, also ask AI to update AGENTS with that info

test/test-constraint-scenario-tail-excess.jl
```diff
@@ -0,0 +1,112 @@
...
+    energy_problem = TulipaEnergyModel.EnergyProblem(connection)
+    TulipaEnergyModel.create_model!(energy_problem)
+
+    total_cost_per_scenario =
+        energy_problem.expressions[:scenario_tail_excess].expressions[:total_cost_per_scenario]
+
+    model = energy_problem.model
+    tail_excess_indices = energy_problem.variables[:tail_excess_slack_xi].indices
+    tail_excess_vars = energy_problem.variables[:tail_excess_slack_xi].container
+    value_at_risk_threshold_mu = energy_problem.variables[:value_at_risk_threshold_mu].container[1]
```
abelsiqueira: only

test/test-constraint-scenario-tail-excess.jl
```diff
@@ -0,0 +1,112 @@
+@testitem "Create scenario tail excess for case study" setup = [CommonSetup] tags =
```
abelsiqueira: The name feels off, but if you like it, that's fine

test/test-constraint-scenario-tail-excess.jl
```diff
@@ -0,0 +1,112 @@
...
+            expected_total_cost,
+            units_on_operational_cost_per_scenario[row.scenario],
+        )
+
+        @test expected_total_cost == total_cost_per_scenario[row.id]
+
+        expected_cons = JuMP.@build_constraint(
+            tail_excess_vars[row.id] >=
+            total_cost_per_scenario[row.id] - value_at_risk_threshold_mu
+        )
```
abelsiqueira, line 255: It's hard to follow which constraint is expected to be implemented.
Maybe we can be explicit here with the values?

test/test-constraint-scenario-tail-excess.jl
```diff
@@ -1,3 +1,186 @@
...
+        ((investment_cost, assets_investment), (decommission_cost, assets_decommission))
+        JuMP.add_to_expression!(base_cost, coef, var)
+    end
+
+    # Create expected flows operational cost per scenario
+    operational_cost_per_scenario = Dict{Int64,JuMP.AffExpr}()
+    for scenario in 1:num_scenarios
+        operational_cost_per_scenario[scenario] = JuMP.AffExpr(0.0)
+        for rp in 1:num_rep_periods
+            weight = get_weight(connection, rp, scenario)
```
abelsiqueira: It should be faster to create `weights_lookup` by

```julia
weight_lookup = Dict((row.rp, row.scenario) => row.weight) for row in DuckDB.query(...))
```

Can still be split into a separate function call if necessary.

abelsiqueira (changes requested):
Hi @datejada, thanks for the PR.

This was a bit hard to follow, and it's possible some things were left behind. In particular, there was code moving between files that I assume was correct.
I left a few comments per commit. Sometimes the suggestion function didn't work.

test/test-constraint-scenario-tail-excess.jl
datejada: Great! I like it better. Changed in the latest update.

test/test-constraint-scenario-tail-excess.jl
datejada: There is an update to make more clear (I hope). Lines https://github.com/TulipaEnergy/TulipaEnergyModel.jl/pull/1575/changes#diff-87fdb66609efbeb19cb26ed96255336168cfc850de9876f31c30e0532edd0632R137-R160

Construct the terms from the input data; I think it reads better from there what you expect. The tricky part is to create it per scenario.

I can write term by term, instead of creating the expressions, to make it more explicit. Let me know if the change improved or if you prefer the explicit version. Both are okay.

Thanks

test/test-constraint-scenario-tail-excess.jl
datejada: True, I changed it so it read clearly

test/test-constraint-scenario-tail-excess.jl
datejada: updated here and in all the places that I have used it for testing.

test/test-constraint-scenario-tail-excess.jl
datejada: True, I forgot about that one. It is updated and also added to the AGENTS.md

src/expressions/conditional_value_at_risk.jl
datejada: I am not sure if I follow the comment. Do you mean, not using `Let`? I was trying to keep the same structure as in other constraints. But elaborate a bit more on your comment; maybe it's something else you are referring to.

src/constraints/conditional_value_at_risk.jl
datejada: done!

src/expressions/operational-costs.jl
datejada: The main structure is the same as in main:
https://github.com/TulipaEnergy/TulipaEnergyModel.jl/blob/f95cb9193fa11ea545f4d6862786bf858a17dee2/src/objectives/flows_operational_cost.jl#L96-L117

But here we need the cost per scenario rather than the total costs across scenarios. I have added docstrings to the functions to help follow what is happening in this function. I hope it helps.

Let me know if it is enough or if you want more details.

src/expressions/operational-costs.jl
datejada: I have added docstrings to all the functions, including this one to help understand what is happening here.

I created this function with the help of Copilot. Here is the explanation of why this is a good implementation for the purpose of this PR (I think it is a good approach, but please let me know if you think differently):

```text
Why _scenario_ranges is needed
The three SQL queries (_query_flows_operational_cost_per_scenario_indices, etc.) return flat rows sorted by (scenario, var.id). To build a Vector{JuMP.AffExpr} with one expression per scenario, the code needs to group those rows by scenario. _scenario_ranges does exactly that in O(n) — detecting contiguous run boundaries — and returns a Dict{Int32, UnitRange{Int}} used to slice the flat array per scenario.

Without it, each of the 3 expression builders would need either:

Additional SQL queries (one per scenario), or
An O(n × S) scan (filter the full array for each scenario)
It's called 3 times (once per expression type) and its output directly enables the CVaR expression builder in conditional_value_at_risk.jl to access expressions[:flows_operational_cost_per_scenario].expressions[:cost][i] by position.

The current _scenario_ranges approach is more efficient. Here's why:

SQL-per-scenario approach

S separate DuckDB queries, each with: query parsing, planning, execution, and result materialization overhead
DuckDB round-trip cost is non-trivial (milliseconds per call)
Total cost scales as O(S × query_overhead + n)
Current _scenario_ranges approach

1 SQL query fetching all n rows, already sorted by (scenario, var.id)
1 O(n) Julia pass to detect contiguous run boundaries
Total cost: O(query_overhead + n)
The SQL results are already sorted by scenario (the ORDER BY rp_weight.scenario, var.id at the end of each query). _scenario_ranges exploits that sort to do a single linear scan — it just detects where consecutive scenario_ids differ. This is essentially free compared to S additional query round-trips.

In stochastic optimization problems the number of scenarios S can be large (tens to hundreds), making S repeated queries significantly more expensive than the current O(n) scan. The current implementation is the right choice.
```

test/test-objective-operational-costs-per-scenario.jl
datejada: Updated, thanks!

test/test-objective-operational-costs-per-scenario.jl
datejada: Good point. It was updated. Thanks

src/sql/create-variables.sql
datejada: This is to ensure the alignment when building the constraints. I got an error when running one example because I gave the scenario order 1998, 2009, 2008. So, since the assumption of order is other constraints, I followed that idea and ensured the order here as well for this feature.

src/objectives/create.jl
datejada: The main reason is that I want to store in an expression the total cost per component and scenario for this new constraint. But, in the objective function, the cost is multiplied by a parameter (1-lambda), so if registered here, then it includes the (1-lambda) factor. So, that is why I register the expression first, before calling the _add_to_objective! without the (1-lamdba) (to reuse in the constraint) but I pass to the objective the expression times the (1-lamdba). This follows the formulation in here:

https://github.com/TulipaEnergy/TulipaEnergyModel.jl/issues/1523#issuecomment-4044693505

I think with that context, the change makes more sense, but please let me know if you need more details. Thanks!

datejada:
> Hi @datejada, thanks for the PR.
> 
> This was a bit hard to follow, and it's possible some things were left behind. In particular, there was code moving between files that I assume was correct. I left a few comments per commit. Sometimes the suggestion function didn't work.

Thanks, @abelsiqueira. Sorry, this is a huge and not easy-to-follow PR. Also, I didn't provide enough context for the PR's aim. It boils down to trying to implement constraint (6j) here:

https://github.com/TulipaEnergy/TulipaEnergyModel.jl/issues/1523#issuecomment-4044693505

For that, you would need the cost per scenario. So, I tried to create expressions that I was able to reuse in the constraint. Otherwise, I would be repeating the cost calculation in two places. In addition, the cost in the objective function is multiplied by (1-lambda). So, the expressions don't include that, so I can reuse them in the constraint.

Anyway, I have implemented the changes and replied to your comments. Please let me know if you have more comments or suggestions.

Thanks!

src/sql/create-variables.sql
abelsiqueira: Makes sense. I'll tag #1335 here to create the x-ref in case we work on it in the future
Marking as resolved

src/expressions/operational-costs.jl
abelsiqueira: Got it. I think the `collect` might become a problem, especially for larger results.
So something like

```julia
costs = [
    ...
    for row in indices
]
```

might be necessary, like we do in the expressions and constraints

src/expressions/operational-costs.jl
```diff
@@ -0,0 +1,375 @@
...
+    connection,
+    model,
+    variables,
+    expressions,
+)
+    expr_name = :vintage_flows_operational_cost_per_scenario
+    expr = _create_scenario_cost_expression!(connection, expressions, expr_name)
+    expr_indices = DuckDB.query(connection, "FROM expr_$expr_name ORDER BY id") |> collect
+
+    indices = _query_vintage_flows_operational_cost_per_scenario_indices(connection) |> collect
```
abelsiqueira: Here again we're collecting. It's more complicated to fix this one, since you want to relate two tables.
Our normal strategy would be to use `array_agg` to link one index to another, but here we have `_scenario_ranges` building ranges, so I don't know what is possible.
Is the benchmark picking up this change? I would expect some slowdown here if it is. If there is no slowdown even with the benchmark picking it up, then I'm overthinking it.

src/expressions/conditional_value_at_risk.jl
abelsiqueira: Sorry, I meant to use `expr.num_rows` explicitly below instead of `n`

src/expressions/conditional_value_at_risk.jl
```diff
@@ -0,0 +1,73 @@
...
+        flows_operational_cost_per_scenario =
+            expressions[:flows_operational_cost_per_scenario].expressions[:cost]
+        vintage_flows_operational_cost_per_scenario =
+            expressions[:vintage_flows_operational_cost_per_scenario].expressions[:cost]
+        units_on_operational_cost_per_scenario =
+            expressions[:units_on_operational_cost_per_scenario].expressions[:cost]
+
+        # Construct the total cost per scenario expression for each scenario
+        total_cost_per_scenario = Vector{JuMP.AffExpr}(undef, n)
+        for i in 1:n
```
abelsiqueira:
```suggestion
        total_cost_per_scenario = Vector{JuMP.AffExpr}(undef, expr.num_rows)
        for i in 1:expr.num_rows
```

src/expressions/operational-costs.jl
abelsiqueira: Looking around, it looks like there `_query` functions are only used once each. Their main goal being to create these indices, which are then aggregated below per scenario to create an expression.
If that's the case, then it might be possible to create `group by scenario` and use `array_agg(id) as var_vintage_flow_ids, array_agg(cost) as arr_costs` and loop over those instead in the `attach_expression!`.
This is the same idea as the group investment in `src/constraints/investment_group.jl`.
Sorry if this is obviously not the case, but I can't see all the connected pieces clearly.

src/expressions/operational-costs.jl
```diff
@@ -0,0 +1,473 @@
...
+                    sum(
+                        indices[i].cost * vintage_flow[indices[i].id::Int64] for
+                        i in range_per_scenario[row.scenario]
+                    ),
+                )
+            else
+                @expression(model, 0.0)
+            end for row in expr_indices
+        ],
+    )
```
abelsiqueira, line 171: Code suggestion related to the discussion above.
Assuming it's possible to group the stuff from `indices` and join into `expr_indices`

```suggestion
    attach_expression!(
        expr,
        :cost,
        JuMP.AffExpr[
            if length(row.var_vintage_flow_ids) > 0
                @expression(
                    model,
                    sum(
                        cost * vintage_flow[var_id::Int64]
                        for (cost, var_id) in zip(row.arr_cost, row.var_vintage_flow_ids)
                    ),
                )
            else
                @expression(model, 0.0)
            end for row in expr_indices
        ],
    )
```

src/expressions/operational-costs.jl
abelsiqueira: I made comments in other places. The AI answer is too localized. It justifies this function assuming the other decisions around the function are also necessary. I made a few comments in a few places that maybe clarify what I'm essence, this raised a flag because in my mind it looks like an aggregation of the indices, which we do directly in SQL in other places

test/test-constraint-scenario-tail-excess.jl
```diff
@@ -0,0 +1,279 @@
...
+        cost_per_scenario = JuMP.AffExpr(0.0)
+        JuMP.add_to_expression!(cost_per_scenario, base_cost)
+        JuMP.add_to_expression!(cost_per_scenario, flows_operational_cost_per_scenario[scenario])
+        JuMP.add_to_expression!(
+            cost_per_scenario,
+            vintage_flows_operational_cost_per_scenario[scenario],
+        )
+        JuMP.add_to_expression!(cost_per_scenario, units_on_operational_cost_per_scenario[scenario])
+
+        @test cost_per_scenario == total_cost_per_scenario[id]
```
abelsiqueira:
```suggestion
        @test cost_per_scenario ≈ total_cost_per_scenario[id]
```

test/test-constraint-scenario-tail-excess.jl
abelsiqueira: That looks good enough, I think, thanks

abelsiqueira (changes requested):
Thanks, @datejada. I've made a few more comments now that I get the picture better

test/test-constraint-scenario-tail-excess.jl
datejada: Actually, it is not possible to do it like this. It throws an error: "approx doesn't exist" when comparing JuMP expressions. This test compares the expression, not its value. So, `==` is correct here. If we were solving the problem, then comparing the values with approx would be correct.

src/expressions/operational-costs.jl
datejada: I see your point and I agree, I didn't think about it like that, but it is true, it is an aggregation at the end. So, having consistency with other parts of the code is a good idea. I will change it to use SQL aggregations. Thanks!

datejada:
@abelsiqueira, thank you for your comments. I have updated the code, and I think now it is more aligned with the other standards we have elsewhere, as you suggested. Here is a summary of the main changes (I also had to update the tests when rebasing):

**1. Replace `collect` + Julia-side grouping with SQL `ARRAY_AGG`**

All aggregation was moved into SQL using `ARRAY_AGG(... ORDER BY id)` grouped by scenario, so each query returns **one row per scenario** carrying all the per-flow/per-vintage-flow/per-unit arrays.

**2. Delete `_scenario_ranges`**

The helper function `_scenario_ranges` (which computed per-scenario index ranges from a flat sorted result set) became unused after the SQL `ARRAY_AGG` refactor and was removed.

Please let me know if you have any further comment or suggestion

src/expressions/operational-costs.jl
```diff
@@ -70,53 +71,53 @@ function _add_flows_operational_cost_per_scenario_expressions!(
...
-        end
-    end
 
-    range_per_scenario = _scenario_ranges(indices)
+    indices = _query_flows_operational_cost_per_scenario_indices(
+        connection,
+        expr_name,
+        has_commodity_price_profile,
+    )
+
```
abelsiqueira, line 82: A `costs` array is being created inside the loop, so pre-allocating should be better, since we don't use the full vector, only arrays.
Something like

```julia
max_n_flows = get_single...("SELECT MAX(ARRAY_LENGTH(var_flow_ids)) FROM ...")
costs = Vector{Float64}(undef, max_n_flows)
```

src/expressions/operational-costs.jl
```diff
@@ -70,53 +71,53 @@ function _add_flows_operational_cost_per_scenario_expressions!(
...
     attach_expression!(
         expr,
         :cost,
         JuMP.AffExpr[
-            if haskey(range_per_scenario, row.scenario)
+            if length(row.var_flow_ids) > 0
+                n_flows = length(row.var_flow_ids)
+                costs = Vector{Float64}(undef, n_flows)
```
abelsiqueira: Then this line can be removed

src/expressions/operational-costs.jl
```diff
@@ -70,53 +71,53 @@ function _add_flows_operational_cost_per_scenario_expressions!(
...
+                                1.0,
+                            )
+                            row.arr_cost_coefficient[i]::Float64 *
+                            (
+                                row.arr_commodity_price[i]::Float64 * commodity_price_agg /
+                                row.arr_producer_efficiency[i]::Float64 +
+                                row.arr_operational_cost[i]::Float64
+                            ) *
+                            (row.arr_time_block_end[i] - row.arr_time_block_start[i] + 1)
+                        end
```
abelsiqueira: To increase readability, this can be

```julia
if ...
    costs[i] = ...
else
    commodity_...
    costs[i] = ...
end
```

abelsiqueira (approved):
Hi @datejada, thanks for the update. It's very helpful. I've made three extra comments below, which are not blocking for approval, so it's pre-approved.

Two are about making a small modification to pre-allocate the costs. It shouldn't be hard to implement, but if it is, that's fine, we can skip it.
The last is about readability, but it's opinion.

I've made the comments in commits, so they're marked as outdated, I hope you can see them.

src/expressions/operational-costs.jl
datejada: Thanks! `costs` is now pre-allocated once before `attach_expression!` using `max_n_flows` (a single `SELECT MAX(cnt)` query over flows grouped by scenario).

src/expressions/operational-costs.jl
datejada: Great! With the change in the other comment, the `costs = Vector{Float64}(undef, n_flows)` allocation that was inside the per-scenario loop body is removed.

src/expressions/operational-costs.jl
datejada: I agree, it helps with the readability. So, the `costs[i] = if ... else ... end` form is replaced with the more readable `if ... costs[i] = ... else ... commodity_price_agg = ...; costs[i] = ... end` form.

datejada:
> Hi @datejada, thanks for the update. It's very helpful. I've made three extra comments below, which are not blocking for approval, so it's pre-approved.
> 
> Two are about making a small modification to pre-allocate the costs. It shouldn't be hard to implement, but if it is, that's fine, we can skip it. The last is about readability, but it's opinion.
> 
> I've made the comments in commits, so they're marked as outdated, I hope you can see them.

Thanks @abelsiqueira, thanks for the suggestions. I have implemented all three of them 🤓

PR #1576, opened 2026-04-10
Add SQL guidelines for coding

datejada:
More instructions for developers, both humans and agents 😉 

Co-Authored-By: GitHub Copilot (gpt-5-4) <noreply@github.com>

## Related issues

There is no related issue.

## Checklist

- [ ] Tests are passing

### DIFF
--- a/docs/src/90-contributing/91-developer.md
+++ b/docs/src/90-contributing/91-developer.md
@@ -249,10 +249,10 @@ We will try to follow these during development and reviews.
 - Exports: define exports in the source file that owns the public functions.
 - Comments: write complete sentences and prefer explaining why over how.
 - DuckDB / SQL
-  - Put large reusable DDL or staging pipelines in `src/sql/*.sql`; keep short single-use queries inline in Julia near the code that consumes them.
+  - Put large reusable SQL queries or staging pipelines (preparing indices tables) in `src/sql/*.sql`; keep short single-use queries inline in Julia near the code that consumes them.
   - Prefer `CREATE OR REPLACE TEMP TABLE` for intermediates reused across multiple later queries, loops, or model-building steps. This is the main staging pattern used across the codebase.
   - Prefer `WITH` clauses (CTEs) over deeply nested subqueries when a single query has multiple logical steps. Use descriptive names such as `milestones`, `discounts`, or `cte_group_expression`. If the same intermediate is reused several times or is expensive to recompute, promote it to a temp table instead of repeating the CTE.
-  - Prefer set-based SQL over Julia-side row-by-row database work: `CREATE TABLE AS SELECT`, `INSERT INTO ... SELECT`, `UPDATE ... FROM`, window functions, `ARRAY_AGG`, and `ANTI JOIN`.
+  - Often, aggregating with SQL is better than using Julia. For instance, using `GROUP BY` and `ARRAY_AGG` to create an array of `id`s can be faster that not grouping and looping row-by-row in Julia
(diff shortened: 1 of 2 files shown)


datejada:
@abelsiqueira @gnawin I think we should add these patterns to the instructions and developers' documentation. I extracted them from our current patterns in the code, but please add/modify/suggest if you consider that it needs modifications.
Thanks!

docs/src/90-contributing/91-developer.md
```diff
@@ -248,6 +248,16 @@ We will try to follow these during development and reviews.
 - Globals: use `UPPER_CASE` for constants.
 - Exports: define exports in the source file that owns the public functions.
 - Comments: write complete sentences and prefer explaining why over how.
+- DuckDB / SQL
+  - Put large reusable DDL or staging pipelines in `src/sql/*.sql`; keep short single-use queries inline in Julia near the code that consumes them.
```
abelsiqueira: It's not obvious to me what DDL is

docs/src/90-contributing/91-developer.md
```diff
@@ -248,6 +248,16 @@ We will try to follow these during development and reviews.
...
+  - Prefer `CREATE OR REPLACE TEMP TABLE` for intermediates reused across multiple later queries, loops, or model-building steps. This is the main staging pattern used across the codebase.
+  - Prefer `WITH` clauses (CTEs) over deeply nested subqueries when a single query has multiple logical steps. Use descriptive names such as `milestones`, `discounts`, or `cte_group_expression`. If the same intermediate is reused several times or is expensive to recompute, promote it to a temp table instead of repeating the CTE.
+  - Prefer set-based SQL over Julia-side row-by-row database work: `CREATE TABLE AS SELECT`, `INSERT INTO ... SELECT`, `UPDATE ... FROM`, window functions, `ARRAY_AGG`, and `ANTI JOIN`.
```
abelsiqueira: It's not clear what this means

abelsiqueira (commented):
Thanks, I've left a few comments

docs/src/90-contributing/91-developer.md
datejada: DDL stands for Data Definition Language. We can add it explicitly. It is because our *.sql queries there define tables rather than manipulating them, which I found is called DML (Data Manipulation Language).

source: https://www.geeksforgeeks.org/sql/sql-ddl-dml-tcl-dcl/

Please add any suggestions you think can help the sentence read better 🤓

docs/src/90-contributing/91-developer.md
datejada: Those are the main patterns found in the current code. I read it as that we prefer using SQL over iterating in Julia objects. But I might be misunderstanding it.

Please add any suggestions you think can help the sentences read better 🤓

docs/src/90-contributing/91-developer.md
abelsiqueira:
```suggestion
  - Put large reusable SQL queries or staging pipelines (preparing indices tables) in `src/sql/*.sql`; keep short single-use queries inline in Julia near the code that consumes them.
```

docs/src/90-contributing/91-developer.md
abelsiqueira: This is the developer docs, so it should be clear to us, at least.
Maybe this is useful:

```suggestion
  - Often, aggregating with SQL is better than using Julia. For instance, using `GROUP BY` and `ARRAY_AGG` to create an array of `id`s can be faster that not grouping and looping row-by-row in Julia
```

abelsiqueira (approved):
Thanks, I've made it clearer what I meant. It's pre-approved

Issue #1355, opened 2025-10-02
Rename investment methods and multi-year expressions


gnawin:
### Description

**Problem**
From the discussions, it appears to be difficult to understand the actual effects of different investment methods. 

Current names: none, simple, compact, semi-compact

In addition, it is also difficult to understand or perceived as inconsistent for the multi-year expression available_units_compact_method, as this one is used for both compact and semi-compact method.

**Proposal**

New names: none or operation, aggregate_initial_capacity (simple), aggregate_vintage_profiles (compact), aggregate_vintage_efficiencies (semi-compact)

New expression name: available_units_aggregate_initial_capacity, available_units_vintage

(Optional) commission_year -> vintage_year

I'm open to other suggestions! But I want to align with both of you @datejada @g-moralesespana

gnawin:
Closed by #1596

Issue #1424, opened 2025-11-14
Add documentation on vintage modeling


gnawin:
Explain different investment methods (will be renamed to vintage method after #1355), what are they, when and how to use them, in the concepts section.

PR #1596, opened 2026-05-04
Refactor and rename investment methods

### LINKED ISSUE #1250: Change the name of `investment_method`
### Description

Currently `investment_method` controls both the creation of investment variables and decommission variables, with the reason that decommission variables are created differently for `simple` or `compact` methods. 

It makes sense for the code, but the name is confusing. I have a few suggestions:
- `investment_and_decommission_method`
- `capacity_expansion_method`
- `capacity_accumulation_method`

Open to suggestions!

gnawin:
**Goal** 
Rename the investment method concept to clearer terminology, and tighten the schema/validation.

**Rename**
The below table is from https://github.com/TulipaEnergy/TulipaEnergyModel.jl/issues/1355#issuecomment-3467483695 , https://github.com/TulipaEnergy/TulipaEnergyModel.jl/issues/1355#issuecomment-3468804478, and https://github.com/TulipaEnergy/TulipaEnergyModel.jl/issues/1250#issuecomment-3095955117,

|  | old | new  |
| - | - | - |
| column | investment_method | vintage_method |
| ~option~ | ~none~ | ~none~ |
| option | simple | aggregated |
| option | compact | compact_profiles |
| option | semi-compact | compact_efficiencies |

The renaming is everywhere, including file names, constraint names etc.

The key takeaway is that `vintage_method` only controls how vintage units are treated, i.e., `aggregated`, `compact_profiles`, and `compact_efficiencies`. It has nothing to do with investment mode or operation mode. Investment and decommission decisions are controlled only by `investable` and `decommissionable`.  

~'none' vintage method is still kept, but only for consumer assets (see below validation). This is a special case so schema defaults cannot be applied anymore (unless we make the default depending on the asset type - a more disruptive change that we have not done before), therefore `vintage_method` becomes mandatory. This makes sense since users need to know how vintage units are treated anyway to properly prepare the input data, and we have some validations already to help.~

**Schema** 
~Removed the schema default for `vintage_method`. Users now must set it explicitly.~
Removed the `none` method. The `none` method was just a nameholder for consumers but was identical to `aggregated` operationally. The default is `aggregated`, also for consumers. 

**Validation ~tightening~**
~Added biconditional validation: consumer assets must have vintage_method = "none"; non-consumer assets must have any other value.~ 
Update validation: consumer assets must have vintage_method = "aggregated". Other assets can have any.

## Related issues

Closes #1250 #1355

## Checklist

- [x] Tests are passing


gnawin:
Hi @datejada, this PR is about refactoring and renaming investment methods, see the main post. The key decisions have been taken in the past (only not explicit about the schema default which is a natural decision to me in this case).

The behavior changes are not big, but the code changes are massive, similar to the previous PR on year -> milestone_year. I recommend you to read the main post and see if you have any questions or comments, and for the code changes you can quickly go through the PR to check the completeness and see if I made some changes you find inappropriate.

Two related issues are about documentation #1424 and validation #1249, which I'll do after this one.

datejada (commented):
@gnawin, thanks for the PR. It is a lot of renaming, and I like the new names since they are more explicit about what they do in the code.

What I don't like so much is the change to the default, and it's being made mandatory. I understand that the default per-asset isn't something we support (and maybe we shouldn't). So maybe we shouldn't have "none" as an option, and the default should be "aggregated", but that parameter does nothing for consumers. That already happens with some parameters, like the unit commitment method (and others); you can add a "basic" method to a consumer, but it does nothing. Also, not ideal, since the user might expect something to happen, but it doesn't.

This is a consequence of having all the assets in a single table, mixed with parameters that affect only a subset of those assets.

Anyway, I have no ideal answer, but I would like first to explore the option of not making that column mandatory.

Please let me know what you think.

datejada:
@gnawin, I have created issue #1600 to add the mandatory columns to the documentation. This is relevant in any case, but more so if we proceed with your initial proposal. Please add/modify/comment as you consider 🤓

gnawin:
@datejada thanks for the brainstorm.

I think it is the most clear to give "none" to consumers, which indicates that it is irrelevant (or maybe rename it to "irrelevant"). The only drawback of this is that it creates more overhead for users, although necessary in my opinion, but it makes it less user-friendly. 

BUT, I agree with your comment
> This is a consequence of having all the assets in a single table, mixed with parameters that affect only a subset of those assets.

This states that this is an issue over the codebase and currently we do not have an ideal way to deal with it. Therefore, your proposal - use aggregated as default - would be consistent with the rest. I'd rather prefer consistency over correctness/clearness in this case, given that it is unlikely that we'll change other parameters with similar issues. If you also agree (i.e., we won't want to change the behavior of existing parameters in the near future), then I'll remove 'none' and use 'aggregated' as the default.

datejada:
> @datejada thanks for the brainstorm.
> 
> I think it is the most clear to give "none" to consumers, which indicates that it is irrelevant (or maybe rename it to "irrelevant"). The only drawback of this is that it creates more overhead for users, although necessary in my opinion, but it makes it less user-friendly.
> 
> BUT, I agree with your comment
> 
> > This is a consequence of having all the assets in a single table, mixed with parameters that affect only a subset of those assets.
> 
> This states that this is an issue over the codebase and currently we do not have an ideal way to deal with it. Therefore, your proposal - use aggregated as default - would be consistent with the rest. I'd rather prefer consistency over correctness/clearness in this case, given that it is unlikely that we'll change other parameters with similar issues. If you also agree (i.e., we won't want to change the behavior of existing parameters in the near future), then I'll remove 'none' and use 'aggregated' as the default.

Thanks, I think it is better because it is what we have in the other parameters that only apply to a certain type of asset. So, I would add in the description of the .json file that it only applies to producers, storage, and conversion. To make it clear in the schema.

src/input-schemas.json
```diff
@@ -188,6 +175,18 @@
...
+    "vintage_method": {
+      "constraints": {
+        "oneOf": [
+          "none",
+          "aggregated",
+          "compact_efficiencies",
+          "compact_profiles"
+        ]
+      },
+      "description": "How vintage units are treated.",
```
datejada:
```suggestion
      "description": "How vintage units are treated. It only applies to producer, storage, or conversion assets.",
```

gnawin: Addressed manually in the latest commits.

gnawin:
@datejada I addressed your comments, see edits in the main post. 

`none` was removed, and `aggregated` becomes the default, also for consumers. 

Let me know if this is fine to you.

datejada (approved):
Thanks for addressing the comments. LGTM now.

PR #1602, opened 2026-05-08
Refactor unit_commitment parameter

### LINKED ISSUE #1598: Change default for the `unit_commitment_method` parameter or remove it
Currently, to enable unit commitment, we need to set two parameters: unit_commitment and unit_commitment_method. Some users find this confusing (feedback from the Hychain case study). 

https://github.com/TulipaEnergy/TulipaEnergyModel.jl/blob/f00de48592ad5a5d9d28f6c9839b51d11c7ba637/src/input-schemas.json#L158

https://github.com/TulipaEnergy/TulipaEnergyModel.jl/blob/f00de48592ad5a5d9d28f6c9839b51d11c7ba637/src/input-schemas.json#L176

I see two options:

**Option A**

We should have "basic" in the method by default, and let unit_commitment be the one that activates the constraints on/off.

**Option B**
Get rid of the unit commitment method parameter and merge it with the unit commitment one, so it has the options like: "none", "basic", "3var" in it, so it is all in one parameter

datejada:
- Replace `unit_commitment_method` with `unit_commitment` and then change the existing parameter from a boolean to string values to account for the methods for unit commitment.

The changes in the MPS file are expected since the changes fix the bug we found in #1598  

The rest are updates in inputs, docs, creation of variables and constraints, etc.

## Related issues

Closes #1598 

## Checklist

- [ ] Tests are passing


datejada:
@gnawin, this is the PR for the changes we discussed earlier. The changes read better going commit by commit. Please let me know if you have any comments or suggestions. Thanks!

EDIT: since there is a change in the schema, we can't run the benchmarks

gnawin (approved):
Thanks for the PR. We have discussed this change, and it looks as expected, so I'm approving.

PR #1606, opened 2026-05-20
Add instructions for finer objective breakdown-down

### LINKED ISSUE #1538: Objective breakdown in post-processing
Follow up of #1389: 
some of the breakdown should be done in post-processing outside the obj_breakdown table. The post-processing consists of documentation of the logic for the KPIs and a SQL implementation of the logic. The KPIs may differ per use case, but we should give some examples. For instance, we want the sum of the outgoing flows per asset for every country (notice that country is not part of the schema).

 _Originally posted by @gnawin in [#1389](https://github.com/TulipaEnergy/TulipaEnergyModel.jl/issues/1389#issuecomment-4067440305)_

gnawin:
## Related issues

Closes #1538 

## Checklist

- [x] Tests are passing

### DIFF
--- a/docs/src/20-user-guide/20-how-to-use.md
+++ b/docs/src/20-user-guide/20-how-to-use.md
@@ -58,6 +58,194 @@ Outputs are sent from Tulipa to DuckDB and can be exported to various file forma
 
 To save the solution to CSV files, you can use [`export_solution_to_csv_files`](@ref). See the [Workflow Tutorial](@ref step-export) for an example showcasing this function.
 
+## [Cost breakdown in post-processing](@id cost-breakdown)
+
+After solving the model, you can compute a detailed cost breakdown per asset or per flow directly from the solved variable and expression values. This does not modify the optimization model — it is purely a post-processing step.
+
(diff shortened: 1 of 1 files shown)


gnawin:
@abelsiqueira the detailed cost breakdown (per asset/flow) is a feature that is requested quite often. @datejada is on holiday, that's why I'm tagging you, hopefully you can have a read 😅 .
 
In view of #1538 I put some exemplary instructions in **post-processing** in the documentation. Since these examples create DuckDB tables, we can consider, in the future, directly output these tables. At the moment, it is in the documentation because 
1) they cover some important breakdowns but are not complete, others use these directly and can learn from these if they want something different, 
2) if we want to integrate them into the main code, some design choices need to be taken, which is unnecessary at the moment, and
3) we do have a few cases running from older versions, and putting these examples in the docs will be the most generic for backward compatibility considerations.

Thank you!

datejada (approved):
LGTM

abelsiqueira:
Sorry for missing this, @gnawin. I have to update my Tulipa e-mail filters, but please ping me in Tulipa's slack next time

PR #1614, opened 2026-06-03
Add min flow limit without unit commitment constraint

### LINKED ISSUE #1597: Add `minimum_flow_limit` constraint to represent must-run features
### Description

From the COMPETES comparison and the HyChain case study, we have identified the need for a parameter to enforce a minimum flow limit without the unit-commitment feature (e.g., must-run assets such as run-of-river, biomass, nuclear, etc..). So, it would be part of the model's flexibility, allowing different levels of detail to represent constraints in the assets. 

We can use this parameter with the profiles to have something like: `flow >= min_value [p.u.] x profile [p.u.] x capacity [MW] x available_units [# of units]`

It feels more flexible if it belongs to the flow's parameters, because then we can define different minimum outputs/inputs. However, the capacity is on the asset, so maybe it's better to have it on the asset. @gnawin @g-moralesespana What do you think?

It should be an easy addition, once we have a clear idea of where to put it. Please let me know your thoughts, and we can add this one.

### Validation and testing

_No response_

### Motivation

_No response_

### Target audience

_No response_

### Can you help?

_No response_

datejada:
The implementation includes:

- **SQL**: Two new `cons_min_output_flow_without_unit_commitment_*` tables in create-constraints.sql
- **Registration**: Keys added to create.jl
- **Constraint logic**: New file min-output-flow.jl with aggregated and compact vintage method variants
- **Expression attachment**: Loop added in model-preparation.jl
- **Add to the Model**: `@timeit` call added in create-model.jl
- **Tests**: test-constraint-min-output-flow.jl
- **Update docs**: update the formulation, how-to, and outputs sections in the docs.

For the test, different edge cases are included:
- producer, aggregated, non-investable
- producer, aggregated, investable
- conversion asset, aggregated
- CO2 output with zero capacity coefficient
- compact profiles vintage method
- unit commitment asset is excluded
- zero min operating point is excluded
- full solve integration test

## Related issues

Closes #1597

## Checklist

- [x] Tests are passing

### DIFF
--- a/src/constraints/min-output-flow.jl
+++ b/src/constraints/min-output-flow.jl
@@ -1,0 +1,131 @@
+"""
+    add_min_output_flow_without_unit_commitment_constraints!(
+        connection, model, expressions, constraints
+    )
+
+Adds the minimum output flow constraints for producer and conversion assets without
+unit commitment and a positive `min_operating_point`.
+
+The constraint enforces:
+```
+sum(capacity_coefficient * outgoing flows) >= min_operating_point [p.u.] * capacity [MW] * available_units [# of units]
+```
+
+This activates only when `unit_commitment = 'none'` and `min_operating_point > 0`.
+Assets with unit commitment already enforce a minimum output flow through the
+`min_output_flow_with_unit_commitment` constraint.
+
+Note that the `capacity_coefficient` ensures that when output flows are scaled down in the capacity constraints,
+they are also reflected in the minimum output flow constraint.
+This is particularly useful for outputs like CO2 emissions,
+which have a coefficient of zero because they do not contribute to the energy output flow and are regarded as byproducts.
+"""
+function add_min_output_flow_without_unit_commitment_constraints!(
+    connection,
+    model,
+    expressions,
+    constraints,
+)
+    expr_avail_aggregated_vintage_method =
+        expressions[:available_asset_units_aggregated_vintage_method].expressions[:assets]
+    expr_avail_compact_method =
+        expressions[:available_asset_units_compact_vintage_method].expressions[:assets]
+
+    # - Aggregated vintage method
+    let table_name = :min_output_flow_without_unit_commitment_aggregated_vintage_method,
+        cons = constraints[table_name]
+
+        indices = _append_min_output_flow_data_to_indices_aggregated_vintage_method(
+            connection,
+            table_name,
+        )
+        attach_constraint!(
+            model,
+            cons,
+            table_name,
+            [
+                @constraint(
+                    model,
+                    outgoing_flow ≥
+                    row.min_operating_point *
+                    row.capacity *
+                    expr_avail_aggregated_vintage_method[row.avail_id],
+                    base_name = "min_output_flow_without_unit_commitment_aggregated_vintage_method[$(row.asset),$(row.milestone_year),$(row.rep_period),$(row.time_block_start):$(row.time_block_end)]"
+                ) for (row, outgoing_flow) in zip(indices, cons.expressions[:outgoing])
+            ],
+        )
+    end
+
+    # - Compact profiles vintage method
+    let table_name = :min_output_flow_without_unit_commitment_compact_vintage_method,
+        cons = constraints[table_name]
+
+        indices = _append_min_output_flow_data_to_indices_compact_method(connection, table_name)
+        attach_constraint!(
+            model,
+            cons,
+            table_name,
+            [
+                @constraint(
+                    model,
+                    outgoing_flow ≥
+                    row.min_operating_point *
+                    row.capacity *
+                    sum(expr_avail_compact_method[avail_id] for avail_id in row.avail_indices),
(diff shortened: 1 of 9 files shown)


datejada:
@abelsiqueira Sorry to bother you. Since Ni is on paternity leave, would you mind having a look at this PR? It is a simple one, since it adds constraints; it doesn't change any of the current logic or structure. The benchmark shows basically no impact in the EU case study, which is good and expected (since there are no constraints like this there). Of course, there is always the question of how these new constraints scale, but since this one doesn't change any current logic and follows the same pattern as other constraints, it should be fine. Until we take benchmarking all the different constraints more seriously (which I think should be the way).

Anyway, please let me know your comments, and if you don't have time for this, it is okay. I will see what we can do internally.

As always, the changes read better by commit.

Thanks!

Diego

src/constraints/min-output-flow.jl
```diff
@@ -0,0 +1,131 @@
...
+            ARRAY_AGG(expr_avail.id ORDER BY expr_avail.id) AS avail_indices,
+        FROM cons_$table_name AS cons
+        LEFT JOIN asset
+            ON cons.asset = asset.asset
+        LEFT JOIN expr_available_asset_units_compact_vintage_method AS expr_avail
+            ON cons.asset = expr_avail.asset
+            AND cons.milestone_year = expr_avail.milestone_year
+        GROUP BY cons.id
+        ORDER BY cons.id
+        ",
```
abelsiqueira: This can be done cleaner with a CTE, so you avoid the `ANY_VALUE`s.
I've asked claude sonnet to write it below. The essence is doing the `GROUP BY` in the CTE so you only join later.

```sql
WITH avail_agg AS (
      SELECT
          asset,
          milestone_year,
          ARRAY_AGG(id ORDER BY id) AS avail_indices
      FROM expr_available_asset_units_compact_vintage_method
      GROUP BY asset, milestone_year
  )
  SELECT
      cons.id,
      cons.asset,
      cons.milestone_year,
      cons.rep_period,
      cons.time_block_start,
      cons.time_block_end,
      asset.capacity,
      asset.min_operating_point,
      avail_agg.avail_indices,
  FROM cons_$table_name AS cons
  LEFT JOIN asset
      ON cons.asset = asset.asset
  LEFT JOIN avail_agg
      ON cons.asset = avail_agg.asset
      AND cons.milestone_year = avail_agg.milestone_year
  ORDER BY cons.id
```

test/test-constraint-min-output-flow.jl
```diff
@@ -0,0 +1,421 @@
...
+        asset_type = :producer,
+        vintage_method = "aggregated",
+        initial_units = 2.0,
+        capacity = 100.0,
+        min_operating_point = 0.4,
+        unit_commitment = "none",
+        investable = false,
+        num_timesteps = 1,
+        num_rps = 2,
+    )
```
abelsiqueira: No need to re-define default parameters since they're already default.

test/test-constraint-min-output-flow.jl
```diff
@@ -0,0 +1,421 @@
...
+        id = row.id
+        flow_id =
+            get_flow_var_id(connection, config.name, "demand", row.rep_period, row.time_block_start)
+
+        # Expected: flow >= min_op * capacity * (initial_units + assets_investment)
+        expected_cons = JuMP.@build_constraint(
+            flow[flow_id] >=
+            config.min_operating_point *
+            config.capacity *
+            (config.initial_units + assets_investment)
```
abelsiqueira, line 170: Is this the equivalent to the `sum(expr_...)` in the constraint? I think so, just making sure

abelsiqueira (approved):
Hi @datejada, sorry for the delay on answering, and thanks for pinging on slack, otherwise I'd miss it.

I took a quick look.
The constraints tables and objects seems to be correct. I haven't checked the mathematical formulation in the docs or the issue to compare to what is implemented, since it's beyond my expertise.
I took a quick look at the tests, and they seem to capture the relevant main cases. I haven't evaluated corner cases.
The benchmark speed changed are well within our expected variation. Since you mentioned that EU doesn't have these constraints, I think the memory changes are due to either just creating small (order of Kb) variables, or type instabilities that we still can't avoid.

Let me know if this review is enough.

I made small suggestions that don't affect correctness, so it's approved (with the caveats above)

src/constraints/min-output-flow.jl
datejada: Great! Thanks. I will implement it like this.

test/test-constraint-min-output-flow.jl
datejada: Correct! I will simplify them. Thanks

test/test-constraint-min-output-flow.jl
datejada: Exactly! That is the expected result from the `sum(expr_...)`

datejada:
@abelsiqueira, thanks for the comments and suggestions. I have included them in the latest commit in this PR.

PR #1635, opened 2026-06-29
Improve Tulipa performance when running in a loop

### LINKED ISSUE #1634: Slow performance when running Tulipa in a loop
@alicezurru found, while running experiments, that running in a loop causes the times to increase with each iteration across many re-runs.

After some investigation, these are potential problems that we have in the current code:

- no reset of the timing state
- avoidable profile-slice allocations in _profile_aggregate
- DuckDB work inside Julia loops for the storage constraints
- Reduce inner-loop allocation in expression builders in model preparations and expressions

datejada:
This PR introduces several changes to improve performance in Tulipa, especially when running in a loop. Here is a summary:
 
- Timer reset for `run_scenario` and `run_rolling_horizon`.
- Profile aggregation now uses views instead of slices.
- Storage constraints now use SQL window-computed `previous_id` / `cycle_id` instead of per-row DuckDB lookups and `row.id - 1`.
- Seasonal storage inter-period inflow profile prep is now one grouped SQL query.
- Reused inner-loop aggregation dictionaries in model/expression builders.
- Removed `only(collect(...))` in `get_model_parameters`.

Changes done with the help of CODEX:

`Co-authored-by: Codex <noreply@openai.com>`

## Related issues

Closes #1634 

## Checklist

- [x] Tests are passing

### DIFF
--- a/src/constraints/storage.jl
+++ b/src/constraints/storage.jl
@@ -326,7 +326,7 @@ function _append_storage_data_to_indices(connection, table_name)
     LAST_VALUE(cons.id) OVER (
         PARTITION BY $id_partition_columns
         ORDER BY $id_order_column
-        ROWS BETWEEN UNBOUNDED PRECEDING AND UNBOUNDED FOLLOWING
+        ROWS BETWEEN UNBOUNDED PRECEDING AND UNBOUNDED FOLLOWING -- frame expanding window to all rows, to avoid assuming ordering of ids
     ) AS cycle_id,
     """
 
--- a/src/constraints/storage.jl
+++ b/src/constraints/storage.jl
@@ -66,20 +66,9 @@ function add_storage_constraints!(
                     else
                         # Initial storage is the previous level (a JuMP variable)
                         previous_level::JuMP.VariableRef = if row.time_block_start > 1
-                            var_storage_level[row.id-1]
+                            var_storage_level[row.previous_id::Int]
                         else
-                            # Find last id of this group (there are probably cheaper ways, in case this becomes expensive)
-                            last_id = get_single_element_from_query_and_ensure_its_only_one(
-                                DuckDB.query(
-                                    connection,
-                                    "SELECT
-                                        MAX(id)
-                                    FROM cons_$table_name
-                                    WHERE asset = '$(row.asset)' AND milestone_year = $(row.milestone_year) AND rep_period = $(row.rep_period)
-                                    ",
-                                ),
-                            )::Int
-                            var_storage_level[last_id]
+                            var_storage_level[row.cycle_id::Int]
                         end
                         computed_storage_loss_coef = 1.0
                         if row.storage_loss_from_stored_energy > 0.0
@@ -193,19 +182,9 @@ function add_storage_constraints!(
                     else
                         # Initial storage is the previous level (a JuMP variable)
                         previous_level::JuMP.VariableRef = if row.period_block_start > 1
-                            var_storage_level[row.id-1]
+                            var_storage_level[row.previous_id::Int]
                         else
-                            last_id = get_single_element_from_query_and_ensure_its_only_one(
-                                DuckDB.query(
-                                    connection,
-                                    "SELECT
-                                        MAX(id)
-                                    FROM cons_$table_name
-                                    WHERE asset = '$(row.asset)' AND milestone_year = $(row.milestone_year) AND scenario = $(row.scenario)
-                                    ",
-                                ),
-                            )::Int
-                            var_storage_level[last_id]
+                            var_storage_level[row.cycle_id::Int]
                         end
                         @constraint(
                             model,
@@ -307,7 +286,7 @@ function add_storage_constraints!(
                     else
                         # Initial accumulated storage intra period is the previous level (a JuMP variable)
                         previous_accumulated_level::JuMP.VariableRef =
-                            var_accumulated_storage_level[row.id-1]
+                            var_accumulated_storage_level[row.previous_id::Int]
                         computed_storage_loss_coef = 1.0
                         if row.storage_loss_from_stored_energy > 0.0
                             duration = row.time_block_end - row.time_block_start + 1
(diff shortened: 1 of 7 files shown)


datejada:
Extra BenchmarkTools check, minimum of 3 samples, fresh connection per sample, baseline from detached `HEAD` worktree:

| Case | Phase | Before ms | After ms | Before MiB | After MiB | Before allocs | After allocs |
|---|---:|---:|---:|---:|---:|---:|---:|
| Tiny | `EnergyProblem` | 7417.308 | 405.962 | 1.179 | 1.179 | 28401 | 28425 |
| Tiny | `create_model!` | 4727.813 | 489.156 | 4.942 | 4.762 | 93909 | 92109 |
| Storage | `EnergyProblem` | 2170.441 | 273.232 | 1.419 | 1.420 | 34803 | 34827 |
| Storage | `create_model!` | 1725.075 | 490.894 | 9.441 | 8.914 | 178740 | 173394 |
| Norse | `EnergyProblem` | 943.347 | 285.182 | 9.392 | 9.341 | 234540 | 232933 |
| Norse | `create_model!` | 2209.009 | 663.451 | 88.049 | 79.720 | 1475587 | 1395424 |

datejada:
@abelsiqueira, this PR implements improvements to speed up Tulipa's performance, especially when it's used repeatedly in a loop. I did the diagnosis and implementation with the help of CODEX. I decided to implement everything in one PR (but introduced by commit), since they all aim to achieve the same goal. I recognise that some of them are simple and might not have a major benefit. I think the most significant commits are the last two. For me, the refactor of id-1 is relevant, since using the LAG function in SQL makes sense to be faster and less time-consuming (I like that approach), and in the last commit you will see that the refactor also pre-allocates the workspace for the profiles and a refactor due to an old TO-DO (Creating inter_period profiles of inflows using the inflows profiles). All that helps to improve, as the benchmarks show. I ran the benchmark locally with the other cases to assess the benefits, as noted in the previous comment.

Anyway, before merging, it would be nice if you could have a look to see if it makes sense or if you have another suggestion.

Thanks!

Diego

src/constraints/storage.jl
```diff
@@ -334,6 +313,22 @@ end
...
+    end
+    select_neighbor_ids = """
+    LAG(cons.id) OVER (
+        PARTITION BY $id_partition_columns
+        ORDER BY $id_order_column
+    ) AS previous_id,
+    LAST_VALUE(cons.id) OVER (
+        PARTITION BY $id_partition_columns
+        ORDER BY $id_order_column
+        ROWS BETWEEN UNBOUNDED PRECEDING AND UNBOUNDED FOLLOWING
```
abelsiqueira: I think it's useful to explain what this does, but the full explanation is not really short.

From Claude, what I got is that this expands the window frame to all the rows, otherwise it would consider up to the current row. There is an alternative using descending ordering, but I don't think it's clearer. Also, if we assume that id is monotonically increase, we could use `MAX`, which would be easier.
(I feel like we do assume that, but I don't remember if we enforce it, so I'll continue as if we didn't).

Maybe something like
```suggestion
        ROWS BETWEEN UNBOUNDED PRECEDING AND UNBOUNDED FOLLOWING -- frame expanding window to all rows, to avoid assuming ordering of ids
```

src/model-preparation.jl
```diff
@@ -812,58 +814,48 @@ function prepare_profiles_structure(connection)
         connection,
         """
         WITH
-            cte_scenarios AS (
-                SELECT DISTINCT scenario
-                FROM rep_periods_mapping
+            cte_storage_profiles AS (
+                SELECT DISTINCT
+                    assets_profiles.profile_name,
+                    assets_profiles.commission_year AS milestone_year
```
abelsiqueira: I don't remember our assumptions anymore, but this is one of the places (the only?) where commission_year = milestone_year. I don't remember why, but it's useful to mark this as a explicit choice for the future.
```suggestion
                    assets_profiles.commission_year AS milestone_year -- here commission_year = milestone_year
```

src/model-preparation.jl
```diff
@@ -812,58 +814,48 @@ function prepare_profiles_structure(connection)
...
-                GROUP BY rp_map.period
-                ORDER BY rp_map.period
-                """,
-            )
-        ]
-        if length(values) > 0
-            inter_period[(profile_name, milestone_year, scenario)] = values
-        end
+        inter_period[(row.profile_name, row.milestone_year, row.scenario)] =
+            convert(Vector{Float64}, row.values)
```
abelsiqueira, line 858: It's not clear to me why this doesn't need the `if length(values) > 0` as it did before, but I also don't remember why we needed it, so I guess it's fine

abelsiqueira (approved):
Thanks @datejada, these look great, especially the profile changes. I've made non-blocking comments, so it's approved.
One thing, though, these extra benchmarks look really suspicious. Maybe they include precompilation time? Otherwise, it's over 10 times faster?!
If they are indeed correct, them an "easy" win would be to add Tiny, Storage, and Norse to benchmarks, and let the our normal benchmark script capture that. Should be more or less straightforward for the agent to create a new "level", like `higher_level/Tiny/create_model`. Either way, not a requirement for this PR because the improvements are clear

datejada:
> Thanks @datejada, these look great, especially the profile changes. I've made non-blocking comments, so it's approved. One thing, though, these extra benchmarks look really suspicious. Maybe they include precompilation time? Otherwise, it's over 10 times faster?! If they are indeed correct, them an "easy" win would be to add Tiny, Storage, and Norse to benchmarks, and let the our normal benchmark script capture that. Should be more or less straightforward for the agent to create a new "level", like `higher_level/Tiny/create_model`. Either way, not a requirement for this PR because the improvements are clear

Thanks Abel. I didn't pay attention to the time haha, I was focusing on the memory and allocations 😝 but, I agree the compilation time was included in the first run (sorry about that). Anyway, I think running the benchmarks on the small cases can also give some extra info, so I created issue #1638 for that (it should be easy, as you said).

src/constraints/storage.jl
datejada: I tried using MAX, but it doesn't return the correct value, so maybe it's how the partition works (since MAX worked in the previous code). Anyway, I will add this comment and keep the partition as it is, and we can check later if there is further improvement here with MAX (in a different way)

src/model-preparation.jl
datejada: Yes, this comes from Ni's changes. In this case, it's an assumption, so I think it is better to add a comment to make it explicit. Thanks

src/model-preparation.jl
datejada: Since `row.values` is guaranteed to be an empty vector from the query, `convert` returns an empty `Vector{Float64}`, and this line is safe 🤓

PR #1643, opened 2026-07-16
Add tolerance to clustering to fix tests

### LINKED ISSUE #1641: [Bug] Test fail due to tolerances
The tests were passing when using v1.24.0, see here

https://github.com/TulipaEnergy/TulipaEnergyModel.jl/actions/runs/28568447371/job/84700613791

After the release of v1.24.1, the test failed due to tolerances in the objective function:

https://github.com/TulipaEnergy/TulipaEnergyModel.jl/actions/runs/29332424054/job/87083320664

```log
Tutorial 9 objective value: Test Failed at /home/runner/work/TulipaEnergyModel.jl/TulipaEnergyModel.jl/test/test-tutorial-files.jl:163
  Expression: ≈(energy_problem.objective_value, 497.19915030358084, atol = 1.0e-5)
   Evaluated: 497.1989540103249 ≈ 497.19915030358084 (atol=1.0e-5)
```

datejada:
This pull request added new parameters to clustering function calls so we always get the same solution within the test's tolerance. As consequence, it also updated expected objective values in several integration tests

Test parameterization improvements:
* Added `tol` and `weight_fitting_kwargs` arguments to several calls, providing finer control over convex hull fitting and optimization within the test cases in `test/test-tutorial-files.jl`. [[1]](diffhunk://#diff-7eefcb1eb9c2d18b7e6189b3050fde8b6fb157f249f1979592902934f8752a65R29-R30) [[2]](diffhunk://#diff-7eefcb1eb9c2d18b7e6189b3050fde8b6fb157f249f1979592902934f8752a65R50-R51) [[3]](diffhunk://#diff-7eefcb1eb9c2d18b7e6189b3050fde8b6fb157f249f1979592902934f8752a65R76-R77)

Test expectation updates:
* Updated expected objective values in integration tests for Tutorials 4, 5, and 9 to match new model outputs, ensuring the tests remain valid after recent changes. [[1]](diffhunk://#diff-7eefcb1eb9c2d18b7e6189b3050fde8b6fb157f249f1979592902934f8752a65L115-R136) [[2]](diffhunk://#diff-7eefcb1eb9c2d18b7e6189b3050fde8b6fb157f249f1979592902934f8752a65L159-R169)

## Related issues

Closes #1641

## Checklist

- [x] Tests are passing

### DIFF
--- a/test/test-tutorial-files.jl
+++ b/test/test-tutorial-files.jl
@@ -26,6 +26,8 @@
             method = :convex_hull,
             distance = TC.Distances.CosineDist(),
             weight_type = :convex,
+            tol = 1e-6,
+            weight_fitting_kwargs = Dict(:learning_rate => 0.001, :niters => 1000),
             layout = TC.ProfilesTableLayout(; year = :milestone_year),
         )
         return nothing
@@ -45,6 +47,8 @@
             method = :convex_hull,
             distance = TC.Distances.CosineDist(),
             weight_type = :convex,
+            tol = 1e-6,
+            weight_fitting_kwargs = Dict(:learning_rate => 0.001, :niters => 1000),
             layout = TC.ProfilesTableLayout(; year = :milestone_year),
         )
         return nothing
@@ -69,6 +73,8 @@
             method = :convex_hull,
             distance = TC.Distances.CosineDist(),
             weight_type = :convex,
+            tol = 1e-6,
+            weight_fitting_kwargs = Dict(:learning_rate => 0.001, :niters => 1000),
             layout,
         )
         return nothing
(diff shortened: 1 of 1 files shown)


datejada:
@gnawin this PR is to fix the tests in main. thanks!

gnawin:
It looks good @datejada thanks for the fix. All tests are passing so I approve.

Still a few questions to understand better:
- why did not we see this error before the release and that it only happened after this release?
- why are the objectives changed quite a bit?
- TC influences the profiles, right? then why are the MPS files not changing? Or for these specific tutorials, TC is only something internal?

gnawin (approved):

datejada:
> It looks good @datejada thanks for the fix. All tests are passing so I approve.
> 
> Still a few questions to understand better:
> 
> * why did not we see this error before the release and that it only happened after this release?
> * why are the objectives changed quite a bit?
> * TC influences the profiles, right? then why are the MPS files not changing? Or for these specific tutorials, TC is only something internal?

Thanks @gnawin, here my replies:

"why did not we see this error before the release and that it only happened after this release?"

It is not 100% clear to me, but it seems like it was random, depending on the machine. The default tolerance in clustering is 1e-5 and I changed to 1e-6 since we test to 1e-5. But, with 1e-5 the clustering can still give a slightly different outcome (but not always)...anyway...I incresed the tolerance in the clustering and then done.


"why are the objectives changed quite a bit?"
Because I increased the number of iterations and decreased the step size, it guarantees that it finds the same values (for this length and number of time series). The convex hull clustering is sensitive to these parameters. So, it found more extreme clusters, hence more expensive objective function.


"TC influences the profiles, right? then why are the MPS files not changing? Or for these specific tutorials, TC is only something internal?"

These are the objective functions of the tutorials; we don't create/compare the MPS of the tutorials. Maybe we should...let me know what you think

Issue #988, opened 2025-01-08
Add demand response feature


datejada:
### Description

Adapt the formulation proposed in this paper to use it in Tulipa:

https://www.sciencedirect.com/science/article/pii/S0360544221027936

### Sub issues

- [ ] Write formulation down in Tulipa style
- [ ] Define new input data and files
- [ ] Add variables
- [ ] Add constraints
- [ ] Add objective function terms
- [ ] Add test case study
- [ ] Add test for the code
- [ ] Update documentation

datejada:
@IsaiMaganTNO please have a look at the paper in the link. Thanks!

PR #1652, opened 2026-07-23
Add simple demand response shifting with recovery

IsaiMaganTNO:
Implement first version demand response (simple) based on demand shift and recovery with time windows. Based on formulation in paper https://www.researchgate.net/publication/336086018_Modelling_Demand_Response_in_Power_Systems  and related to issue  #988 

EDIT: FFTR details

Small comment on compatibility with FFTR that I missed in my email:

As I scanned earlier features, I noticed that the easiest way for the initial (simple) DR implementation is probably to work with the time blocks, where I assume time block = partition size * timestep length. So for a consumer asset with a partition size 5 and timestep hourly, each time block would be 5 hours. 

The big trouble with this is how to handle if the size of the representative period does not divide nicely by the time blocks. So for example, if we work with representative days (24h) and DR recovery window "1" time block, then the time blocks would be: hours 1-5 (1), hours 6-10 (2), hours 11-15 (3) and hours 16-20 (4) as the 4 "normal" time blocks, but since there are only 4 remaining hours in the representative period, currently the implementation takes the remaining 4 hours as its own time block for recovery. If you can think of any way to better approach this please let me know, this was our initial idea in our brainstorm session but maybe there is another approach possible. 

Of course this is also more difficult when we consider non-uniform partitioning, so for example if a user sets explicitly the partitions for an asset to be varying. Because then, if the recovery window is more than 1 time block, for example 2: hours 1-5 (1) and hours 6-7 (2). The net energy shift needs to be weighted now to consider the fact that the first time block in the recovery window is larger than the second one. That's why we do the balancing over the weighted energy shift across the time blocks in the recovery window in every case, when (I think?) it is only necessary if the partioning of the consumer asset is nonuniform. Let me know what you think! 

## Related issues

 #988 

## Checklist

- [ ] Tests are passing


src/constraints/demand-response.jl
```diff
@@ -0,0 +1,105 @@
...
+            profiles.rep_period,
+            (row.profile_name, row.milestone_year, row.rep_period),
+            row.time_block_start:row.time_block_end,
+            Statistics.mean,
+            1.0,
+        )
+        max_deviation = row.dr_max_shift_fraction * demand_agg * row.peak_demand
+        JuMP.set_upper_bound(d_increase[row.id], max_deviation)
+        JuMP.set_upper_bound(d_decrease[row.id], max_deviation)
+    end
```
datejada, line 53: The implementation relies on two parallel tables and positional id alignment in several places, which can break easily in the future. var_dr_demand_increase and var_dr_demand_decrease are created separately in create-variables.sql and then assumed to stay in aligned when bounds and balance terms are applied here. This works today, but a future SQL ordering or filter change could silently misapply bounds or objective terms to the wrong JuMP variable.

Here is a suggestion in the code such that the demand-response upper-bound logic no longer assumes the increase and decrease tables stay aligned by positional id. It now joins the two tables on the full DR key and uses the explicit ids from that join when setting bounds.

```suggestion
    for row in DuckDB.query(
        connection,
        "SELECT
            dr_inc.id AS dr_increase_id,
            dr_dec.id AS dr_decrease_id,
            dr_inc.milestone_year,
            dr_inc.rep_period,
            dr_inc.time_block_start,
            dr_inc.time_block_end,
            asset_milestone.peak_demand,
            asset_milestone.dr_max_shift_fraction,
            assets_profiles.profile_name,
        FROM var_dr_demand_increase AS dr_inc
        INNER JOIN var_dr_demand_decrease AS dr_dec
            ON dr_inc.asset = dr_dec.asset
            AND dr_inc.milestone_year = dr_dec.milestone_year
            AND dr_inc.rep_period = dr_dec.rep_period
            AND dr_inc.time_block_start = dr_dec.time_block_start
            AND dr_inc.time_block_end = dr_dec.time_block_end
        LEFT JOIN asset_milestone
            ON dr_inc.asset = asset_milestone.asset
            AND dr_inc.milestone_year = asset_milestone.milestone_year
        LEFT OUTER JOIN assets_profiles
            ON dr_inc.asset = assets_profiles.asset
            AND dr_inc.milestone_year = assets_profiles.commission_year
            AND assets_profiles.profile_type = 'demand'
        ORDER BY dr_inc.id
        ",
    )
        demand_agg = _profile_aggregate(
            profiles.rep_period,
            (row.profile_name, row.milestone_year, row.rep_period),
            row.time_block_start:row.time_block_end,
            Statistics.mean,
            1.0,
        )
        max_deviation = row.dr_max_shift_fraction * demand_agg * row.peak_demand
        JuMP.set_upper_bound(d_increase[row.dr_increase_id], max_deviation)
        JuMP.set_upper_bound(d_decrease[row.dr_decrease_id], max_deviation)
    end
```

docs/src/20-user-guide/30-configuring-model-features.md
```diff
@@ -345,6 +345,20 @@ In summary:
...
+
+Consumer assets can shift their demand over time while preserving the total energy consumed within fixed recovery windows. This models flexible loads that can be advanced or postponed but not permanently curtailed. See the [scientific formulation](@ref demand-response-formulation) for the mathematical details.
+
+To enable demand response on a consumer asset:
+
+- In the `asset` table, set `demand_response_method = "shifting_with_recovery"` (the default is `"none"`, which disables demand response).
+- In the `asset_milestone` table, set the following parameters for the same asset and milestone year:
+  - `dr_max_shift_fraction`: the maximum instantaneous deviation as a fraction of the baseline demand, in the range $[0, 1]$. For example, `0.2` allows the demand to increase or decrease by up to 20% in any time block. The default is `0`, which disables shifting.
+  - `dr_window_blocks`: the size of each recovery window, expressed as a number of consecutive demand-response time blocks. The shifted energy must net to zero within each window. If the number of blocks in a representative period is not an exact multiple of this value, the final window is shorter. The default is `1`.
```
datejada, line 357: I am assuming that it must be also > 0, is it in the validation?

docs/src/40-scientific-foundation/40-formulation.md
datejada, line 1: please push the changes into `upstream` then we can see a preview of the documentation, which makes it easier to review. You should have rights to push to upstream, but please let me know if you can't. For big changes like this, it is better to push to upstream 😉

src/constraints/consumer.jl
```diff
@@ -22,11 +22,24 @@ function add_consumer_constraints!(connection, model, variables, constraints, pr
...
+                    JuMP.add_to_expression!(
+                        dr_term,
+                        variables[:dr_demand_increase].container[row.dr_increase_var_id],
+                    )
+                    JuMP.add_to_expression!(
+                        dr_term,
+                        -1.0,
+                        variables[:dr_demand_decrease].container[row.dr_decrease_var_id],
+                    )
+                end
```
datejada, line 37: we are normally storing all the expressions in the expressions structure and then unpacking them to use them in the constraint. See for example the incomming_flow or outgoing_flow. Here the calculation in in the constraints creation, which is not what we usually want to (unless there is no other way). Please check this part.

src/objectives/create.jl
```diff
@@ -48,6 +48,7 @@ function add_objective!(connection, model, variables, expressions, model_paramet
     _add_flows_operational_cost!(connection, model, expressions, objective_expr, lambda)
     _add_vintage_flows_operational_cost!(connection, model, expressions, objective_expr, lambda)
     _add_units_on_operational_cost!(connection, model, expressions, objective_expr, lambda)
+    _add_demand_response_penalty!(connection, model, variables, objective_expr, lambda)
```
datejada, line 51: penalty? or cost? the parameter uses cost in the name, I would use same terms for consistency

src/sql/create-constraints.sql
```diff
@@ -577,6 +577,66 @@ from
...
+            (
+                row_number() over (
+                    partition by
+                        var.asset,
+                        var.milestone_year,
+                        var.rep_period
+                    order by
+                        var.time_block_start
+                ) - 1
+            ) // asset_milestone.dr_window_blocks as window_id,
```
datejada, line 608: I think I can follow this code, but it is not 100% clear to me. Please add a comment to explain this part further.

test/test-demand-response.jl
datejada, line 1: The SQL comment in create-constraints.sql:583 says the last window may be shorter, but the new tests in this file only checks exact-multiple cases. Please add an edge case in the tests given the expected behavior in this case.

datejada (commented):
Thanks @IsaiMaganTNO, the implementation is looking really good; congrats. I have comments, but I am happy with this first version. Let's wait for more comments from @gnawin, but I hope mine are helpful.

Regarding your major concern. I do agree that with representative periods, there is a decision to make there. I see three options:

1. Keep it as it is now, and make explicit the assumption in the documentation, plus testing what we expect given that assumption. Also, in the code, make explicit that assumption for the future.
2. We assume a `cyclic constraint`, meaning that the end of the representative is not the end; it continues with the beginning of the same representative period. We are doing this with the storage constraints if there is no initial value. We can assume the same here.
3. We can create "inter-temporal" constraints for the demand response. But honestly, I think this is something to investigate further and maybe for the future, not for now.

A couple of more things that I consider important for this case (in general) for the code (not the formulation):
 - PRs like this are better if they are pushed to upstream because then we can check the documentation creation and benchmarks more easily. Those can't be done in a PR coming from a fork.
 - Divide the PR into two parts. One with the changes only in the schema and the files, and another on top of that with the rest of the changes. Why? Because as it is, we can't do a benchmark on this PR since there are changes in the schema (JSON file). So, if we merge first the changes in the schema into main, then we can run the benchmark in the second PR. I am really curious how much this change affects the current base code, since we are touching the consumer balance. So, for me it is important we benchmark this PR and the changes. And the only way to do it is by first merging the changes in the JSON file and the CSV files (without any other change in the code).


A final, maybe too much to ask, but if possible, add a tutorial in the docs. That would help people to understand better what is in there for this functionality.

EDIT: I didn't put too much emphasis on the formulation itself since you have discussed it with @g-moralesespana and this simple version seems like a good start for more complex versions of this feature. In addition, I think a review from @abelsiqueira will be nice to have (but we all know the limitations we have at the moment).

IsaiMaganTNO:
Thanks a lot for your extensive review and comments @datejada ! I have now placed a new PR in #1662 that should work as live on upstream repo, should I close this one? As I am on leave after today I cannot implement the rest of your suggestions but I will make sure to get to them once I have returned!

datejada:
> Thanks a lot for your extensive review and comments @datejada ! I have now placed a new PR in #1662 that should work as live on upstream repo, should I close this one? As I am on leave after today I cannot implement the rest of your suggestions but I will make sure to get to them once I have returned!

Thanks! I will close this PR and continue in the #1662

abelsiqueira:
Hi guys, thanks for tagging me, but indeed, as @datejada mentioned, I won't be able to review it

PR #1653, opened 2026-07-24
Add available-unit investment group constraints

### LINKED ISSUE #1302: Group capacity limit for available units
### Description

Current the [group capacity limit](https://tulipaenergy.github.io/TulipaEnergyModel.jl/stable/40-scientific-foundation/40-formulation/#investment-group-constraints) applies only to the investments per milestone year , it would be useful to have another constraint (or change the existing?) to limit the available units per milestone year.

### Validation and testing

_No response_

### Motivation

Previous discussions
- https://github.com/TulipaEnergy/TulipaEnergyModel.jl/issues/1299#issuecomment-3135218970
- https://github.com/TulipaEnergy/TulipaEnergyModel.jl/issues/1299#issuecomment-3139071018
- https://github.com/TulipaEnergy/TulipaEnergyModel.jl/issues/1299#issuecomment-3139084985

### Target audience

users

### Can you help?

always

datejada:
Replace the boolean `invest_method` with explicit methods:

- `none`
- `use_only_investment_units`
- `use_available_units`

Available-unit groups reuse the model’s existing aggregated and compact available-unit expressions, preserving initial capacity, surviving investments, and decommissioning behavior. A direct method branch keeps the constraint builder small while allowing groups to mix vintage methods.

The migration script maps legacy `true` to `use_only_investment_units` and `false` to `none`; Norse was migrated with it.

Validation now requires investable members only for investment-only groups and matches group membership by both name and milestone year. Tests cover all methods, both expression containers, compact commission years, and validation behavior. Documentation explains the methods, coefficient semantics, and formulation.

## Related issues

Closes #1302

## Checklist

- [ ] Tests are passing

### DIFF
--- a/src/constraints/investment_group.jl
+++ b/src/constraints/investment_group.jl
@@ -66,12 +66,13 @@ function add_investment_group_constraints!(connection, model, variables, express
     return
 end
 
-_get_expression_assets(expressions, k) =
-    if haskey(expressions, k)
-        get(expressions[k].expressions, :assets, JuMP.AffExpr[])
-    else
-        JuMP.AffExpr[]
+function _get_expression_assets(expressions, expression_name)
+    if !haskey(expressions, expression_name)
+        return JuMP.AffExpr[]
     end
+
+    return get(expressions[expression_name].expressions, :assets, JuMP.AffExpr[])
+end
 
 function _create_empty_available_units_table_if_missing!(connection, table_name)
     if isempty(
--- a/src/constraints/investment_group.jl
+++ b/src/constraints/investment_group.jl
@@ -5,10 +5,16 @@ Adds group constraints for assets that share a common limits or bounds.
 """
 function add_investment_group_constraints!(connection, model, variables, expressions, constraints)
     assets_investment = variables[:assets_investment].container
-    expr_available_asset_units_aggregated =
-        _get_expression_assets(expressions, :available_asset_units_aggregated_vintage_method)
-    expr_available_asset_units_compact =
-        _get_expression_assets(expressions, :available_asset_units_compact_vintage_method)
+    expr_available_asset_units_aggregated = get(
+        expressions[:available_asset_units_aggregated_vintage_method].expressions,
+        :assets,
+        JuMP.AffExpr[],
+    )
+    expr_available_asset_units_compact = get(
+        expressions[:available_asset_units_compact_vintage_method].expressions,
+        :assets,
+        JuMP.AffExpr[],
+    )
 
     let table_name = :group_investment
         cons = constraints[table_name]
@@ -66,44 +72,7 @@ function add_investment_group_constraints!(connection, model, variables, express
     return
 end
 
-function _get_expression_assets(expressions, expression_name)
-    if !haskey(expressions, expression_name)
-        return JuMP.AffExpr[]
-    end
-
-    return get(expressions[expression_name].expressions, :assets, JuMP.AffExpr[])
-end
-
-function _create_empty_available_units_table_if_missing!(connection, table_name)
-    if isempty(
-        DuckDB.query(connection, "FROM duckdb_tables() WHERE table_name = '$table_name' LIMIT 1"),
-    )
-        DuckDB.query(
-            connection,
-            """
-            CREATE TEMP TABLE $table_name (
-                id BIGINT,
-                asset VARCHAR,
-                milestone_year INT,
-                commission_year INT
-            );
-            """,
-        )
-    end
-
-    return
-end
-
 function _append_group_data_to_indices!(connection, cons_table_name)
-    _create_empty_available_units_table_if_missing!(
-        connection,
-        "expr_available_asset_units_aggregated_vintage_method",
-    )
-    _create_empty_available_units_table_if_missing!(
-        connection,
-        "expr_available_asset_units_compact_vintage_method",
-    )
-
     return DuckDB.query(
         connection,
         """
--- a/src/constraints/investment_group.jl
+++ b/src/constraints/investment_group.jl
@@ -66,36 +66,17 @@ function add_investment_group_constraints!(connection, model, variables, express
     return
 end
 
-function _get_expression_assets(expressions, expression_name)
-    if !haskey(expressions, expression_name)
-        return JuMP.AffExpr[]
+_get_expression_assets(expressions, k) =
+    if haskey(expressions, k)
+        get(expressions[k].expressions, :assets, JuMP.AffExpr[])
+    else
+        JuMP.AffExpr[]
     end
-
-    return get(expressions[expression_name].expressions, :assets, JuMP.AffExpr[])
-end
-
-function _ensure_group_available_units_tables!(connection)
-    _create_empty_available_units_table_if_missing!(
-        connection,
-        "expr_available_asset_units_aggregated_vintage_method",
-    )
-    _create_empty_available_units_table_if_missing!(
-        connection,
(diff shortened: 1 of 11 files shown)


datejada:
@gnawin, this is the PR; you were wise to say that the changes might not be straightforward. As always, the compact and aggregated method adds complexity to the changes. But with help from CODEX, I think it is pretty close to what we want/need. I added emphasis to the tests so they cover the aggregated + compact method in the available units method. But please double-check if it makes sense or if more tests should be added here (maybe with decommission variables?). I will let you decide, since you are more aware of this part of the code. If you think we need more tests for the available units method, please don't hesitate to list or suggest them.

Thanks and looking forward to your review!

EDIT: I stopped the benchmark since the PR includes schema changes, and those always failed in the Benchmark. I realised it afterwards. Sorry!
EDIT2: I added the review from copilot as an extra check. The summary of the changes in the files here https://github.com/TulipaEnergy/TulipaEnergyModel.jl/pull/1653#pullrequestreview-4773652186 is also useful for you if you want to have an general overview of the changes per file.

src/constraints/investment_group.jl
```diff
@@ -23,53 +27,149 @@ function add_investment_group_constraints!(connection, model, variables, constra
                         MathOptInterface.LessThan(0.0)
                     end
 
+                    group_expression = if row.invest_method == "use_only_investment_units"
```
gnawin, line 36: Did you miss the case when `row.invest_method == "none"`?

src/sql/create-constraints.sql
```diff
@@ -665,12 +665,13 @@ select
...
     ga.name,
     ga.milestone_year,
+    ga.invest_method,
     ga.constraint_sense,
     ga.rhs,
 from
     investment_group_asset as ga
 where
-    ga.invest_method
+    ga.invest_method != 'none'
```
gnawin, line 674: okay, it's here :)

src/constraints/investment_group.jl
gnawin: Nvm, I figured it out.

src/constraints/investment_group.jl
```diff
@@ -23,53 +27,149 @@ function add_investment_group_constraints!(connection, model, variables, constra
...
+                                row.available_asset_units_aggregated_ids,
+                                row.available_asset_units_aggregated_coefficients,
+                            )
+                        ) + sum(
+                            coefficient * expr_available_asset_units_compact[id] for
+                            (id, coefficient) in zip(
+                                row.available_asset_units_compact_ids,
+                                row.available_asset_units_compact_coefficients,
+                            )
+                        )
```
gnawin, line 59: A note: 
this means that when applying this constraint, assets do not need to have the same investment method, can either be aggregated or compact.

src/constraints/investment_group.jl
```diff
@@ -23,53 +27,149 @@ function add_investment_group_constraints!(connection, model, variables, constra
...
-            FROM investment_group_asset
-            LEFT JOIN investment_group_asset_membership
-                ON investment_group_asset.name = investment_group_asset_membership.group_name
-                AND investment_group_asset.milestone_year = investment_group_asset_membership.milestone_year
-            LEFT JOIN var_assets_investment AS var
-                ON investment_group_asset_membership.asset = var.asset
-                AND investment_group_asset.milestone_year = var.milestone_year
-            GROUP BY investment_group_asset.name, investment_group_asset.milestone_year
-        )
+        WITH
```
gnawin, line 79: This function reads too long, and lots of repetitions. 

Other ways I have in my mind would make it shorter but not necessarily increase the readability, so it's fine.

test/test-constraint-investment-groups.jl
```diff
@@ -66,14 +92,103 @@ end
...
+    constraints =
+        Dict(:group_investment => TEM.TulipaConstraint(connection, "cons_group_investment"))
+
+    TEM.add_investment_group_constraints!(connection, model, variables, expressions, constraints)
+
+    observed_constraint = only(_get_cons_object(model, :investment_group))
+    expected_constraint = JuMP.@build_constraint(
+        2.0 * available_units_aggregated +
+        3.0 * available_units_compact[1] +
+        3.0 * available_units_compact[2] <= 100.0
```
gnawin: Would you add a test case where only `available_units_aggregated` is used? This is the most common use case, and testing that would be desirable.

gnawin (changes requested):
Thanks a lot @datejada. It is quite a read for me. 
My only real comment is to add another test case, so we capture all the use cases to double-check the implementation.
Thanks!

src/constraints/investment_group.jl
datejada: Correct! I will add a note in the documentation to mention this. In principle, there is no special reason that avoids this, or at least not that I think of. Please let me know if you think differently.

test/test-constraint-investment-groups.jl
datejada: sure! Here I was covering the most complex one, which is when mixed methods are used. But I agree that the more common case should also be tested. Thanks!

src/constraints/investment_group.jl
datejada: I agree, but since it is a complex query, readability is a bit more relevant here (in my opinion), since the pattern is similar for all CTE-available unit tables. It is also a consequence of allowing this to work with both aggregated and compact methods mixed at the same time in these constraints. But, as I said, it should be allowed, because I can't see a reason of why not (please tell me if you have one). Thanks!

src/sql/create-constraints.sql
```diff
@@ -665,12 +665,13 @@ select
     nextval('id') as id,
     ga.name,
     ga.milestone_year,
+    ga.invest_method,
     ga.constraint_sense,
     ga.rhs,
 from
     investment_group_asset as ga
```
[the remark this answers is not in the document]
datejada: Not relevant; the suggestion doesn't apply.

datejada:
> Thanks a lot @datejada. It is quite a read for me. My only real comment is to add another test case, so we capture all the use cases to double-check the implementation. Thanks!

Thank you @gnawin. I have implemented the suggestions and added the tests for more cases (only aggregated, only compact, and mixed aggregated and compact)

Please let me know if you have more comments or suggestions.

src/constraints/investment_group.jl
```diff
@@ -63,7 +66,55 @@ function add_investment_group_constraints!(connection, model, variables, express
...
+        connection,
+        "expr_available_asset_units_aggregated_vintage_method",
+    )
+    _create_empty_available_units_table_if_missing!(
+        connection,
+        "expr_available_asset_units_compact_vintage_method",
+    )
+
+    return
+end
```
gnawin, line 88: Great to see the handling of missing cases.

src/constraints/investment_group.jl
```diff
@@ -32,21 +32,24 @@ function add_investment_group_constraints!(connection, model, variables, express
                             coefficient * assets_investment[id] for (id, coefficient) in zip(
                                 row.var_assets_investment_ids,
                                 row.var_assets_investment_coefficients,
-                            )
+                            );
+                            init = 0.0,
```
gnawin, line 36: And init values.

gnawin (approved):
Thanks for the improvements. I think now we cover more cases and it is more robust now.

src/constraints/investment_group.jl
datejada: Here is better like this `init = JuMP.AffExpr(),` to be type stable 😉

PR #1655, opened 2026-07-24
Update investment limits name and add available-unit bounds for assests

### LINKED ISSUE #1303: Add available unit upper limit
### Description

Currently we only have the investment_limit for asset investments, but it would be more useful to have a limit for the available units, i.e., considering decommissions.

Note this function can be realized after #1302, using group investment limits, as a workaround. 

But if we choose to implement this feature directly, to me, it makes more sense to remove the asset investment upper bound, and use this investment_limit as the upper limit for available units. But then the name needs to be changed, because it is no longer "investment limit" but "available unit limit". That we need a discussion since it may touch other functions. 

### Validation and testing

_No response_

### Motivation

https://github.com/TulipaEnergy/TulipaEnergyModel.jl/issues/1302#issuecomment-3139173000

### Target audience

users

### Can you help?

always

datejada:
## Summary

Adds minimum and renamed maximum investment limits for asset capacity, storage-energy capacity, and transport flows. Adds optional asset available-unit bounds, with separate minimum and maximum constraint tables sharing the same indexing logic.

## Changes

- Renamed `investment_limit` fields to `investment_max_limit`.
- Added default-zero `investment_min_limit` fields, including storage-energy equivalents.
- Added `min_available_units` and `max_available_units` to `asset_milestone`; flows and storage units are intentionally excluded (future dev).
- Added validation for negative/inconsistent bounds, infeasible integer ranges, and legacy column names.
- Added migration utility: `utils/scripts/migrate-investment-limits.py`.
- Migrated maintained inputs, updated data-generation utilities, documentation, tests, and the intentional Norse MPS snapshot.

## Migration script

This PR adds the migration script migrate-investment-limits.py with a small CLI for user-provided datasets, while preserving existing behavior.

Usage:
- python migrate-investment-limits.py --input-folder PATH
- python migrate-investment-limits.py --input-file PATH
- python migrate-investment-limits.py --input-folder PATH --include-default-datasets

Behavior and compatibility:
- Backward compatible: running the script with no arguments still migrates the default in-repo datasets (tests, benchmark, tutorials), exactly as before.
- With user inputs provided, the script targets those paths only, unless --include-default-datasets is set.
- Help works without duckdb preinstalled in the active Python env (duckdb import is lazy).

## Related issues

Closes #1303

## Checklist

- [ ] Tests are passing


datejada:
@gnawin this is another PR that was larger than expected, sorry about that. On reflection, It can be split into 3 smaller PRs (one for max limit, other for min limit, and one for the available min/max limit). Please let me know if you prefer the split or review it as it is here. Thanks and looking forward to your comments.

src/data-validation.jl
```diff
@@ -488,10 +498,238 @@ function _validate_aggregated_vintage_method_all_milestone_years_are_covered!(
...
+        for _ in DuckDB.query(
+            connection,
+            "SELECT 1
+            FROM duckdb_columns()
+            WHERE table_name = '$table_name'
+                AND column_name = '$column_name'",
+        )
+            push!(
+                error_messages,
+                "Legacy column '$column_name' found in '$table_name'. Run 'uv run --with duckdb python utils/scripts/migrate-investment-limits.py'.",
```
gnawin, line 516: I like the consideration of migration for users. It'd be useful for users that started with the current release and migrate to the upcoming one. Would be good it we have something more general, e.g., have a folder for migration with files for each one, and _somehow_ make it easy for users to upgrade, e.g., with a keyword argument _somewhere_. Any ideas?

It'd be even better for users with older versions can benefit either, but maybe that's not realistic - that I'd recommend using copilot.

src/data-validation.jl
```diff
@@ -488,10 +498,238 @@ function _validate_aggregated_vintage_method_all_milestone_years_are_covered!(
...
+            FROM $table_name
+            $join_clause
+            WHERE $where_clause
+                AND $min_column > 0
+                AND $capacity_column <= 0",
+        )
+            key_values = join([row[index] for index in 1:(length(row)-2)], ", ")
+            push!(
+                error_messages,
+                "Positive '$min_column' requires a positive capacity for '$table_name' at ($key_values).",
```
gnawin: I did not follow this validation. Do you just want to avoid negative capacity? What's special about this combination: positive min_limit and negative capacity?

src/data-validation.jl
```diff
@@ -488,10 +498,238 @@ function _validate_aggregated_vintage_method_all_milestone_years_are_covered!(
...
+    )
+        for row in DuckDB.query(
+            connection,
+            "SELECT $key_selection, $min_column, $max_column, $capacity_column AS capacity
+            FROM $table_name
+            $join_clause
+            WHERE $where_clause
+                AND $max_column IS NOT NULL
+                AND $capacity_column > 0
+                AND CEIL($min_column / $capacity_column) > FLOOR($max_column / $capacity_column)",
```
gnawin, line 722: Isn't this the same as the https://github.com/TulipaEnergy/TulipaEnergyModel.jl/pull/1655/changes#diff-d7d8c58ab1e38469775c325990968cd2a7a2370c6677987fccff29602820fc05R572-R574?

Could you give an example on how CEIL/FLOOR makes a difference here?

src/data-validation.jl
```diff
@@ -501,12 +739,15 @@ function _validate_use_binary_storage_method_has_investment_limit!(error_message
         WHERE asset.type = 'storage'
             AND asset_milestone.investable
             AND asset.use_binary_storage_method IS NOT NULL
-            AND (asset_commission.investment_limit IS NULL OR asset_commission.investment_limit <= 0)
+            AND (
+                asset_commission.investment_max_limit IS NULL
+                OR asset_commission.investment_max_limit <= 0
+            )
```
gnawin, line 751: And this compared to https://github.com/TulipaEnergy/TulipaEnergyModel.jl/pull/1655/changes#diff-d7d8c58ab1e38469775c325990968cd2a7a2370c6677987fccff29602820fc05R569-R571?

src/variables/investments.jl
```diff
@@ -61,3 +57,14 @@ function _find_var_upper_bound(limit, capacity, integer)
     end
     return bound_value
 end
+
+function _find_var_lower_bound(limit, capacity, integer)
+    if capacity <= 0 || ismissing(limit)
```
gnawin: Maybe this is what confuses me in some parts of the validation, and I know we have it in https://github.com/TulipaEnergy/TulipaEnergyModel.jl/pull/1655/changes#diff-76d84767f0c93a04a94a19661d65f50b21433425b668483c533be7b063b1a787L55. But how can capacity <= 0?

src/variables/investments.jl
```diff
@@ -1,53 +1,49 @@
...
-    )
-end
-
 """
     add_investment_variables!(model, variables)
 
 Adds investment variables to the optimization `model`,
 and sets bounds on selected variables based on the input data.
 """
 function add_investment_variables!(model, variables)
```
gnawin, line 7: this way reads better :)

src/constraints/available-units.jl
```diff
@@ -0,0 +1,110 @@
...
+                            (
+                                expr_available_asset_units_compact[id] for
+                                id in row.available_asset_units_compact_ids
+                            );
+                            init = JuMP.AffExpr(0.0),
+                        )
+
+                    @constraint(
+                        model,
+                        available_units - row.rhs in constraint_sense,
```
gnawin, line 45: Here I think it's better to just use something like 

available_units <= row.rhs

without using the sense. Both work but here I think it complicates unnecessarily.

src/data-validation.jl
```diff
@@ -488,10 +498,238 @@ function _validate_aggregated_vintage_method_all_milestone_years_are_covered!(
...
+            max_value = row[Symbol(max_column)]
+            failure_reasons = String[]
+
+            if min_value < 0
+                push!(failure_reasons, "'$min_column' is negative")
+            end
+            if !ismissing(max_value) && max_value < 0
+                push!(failure_reasons, "'$max_column' is negative")
+            end
+            if !ismissing(max_value) && min_value > max_value
```
gnawin:
```suggestion
            if !ismissing(min_value) && !ismissing(max_value) && min_value > max_value
```

src/data-validation.jl
```diff
@@ -488,10 +498,238 @@ function _validate_aggregated_vintage_method_all_milestone_years_are_covered!(
...
+                OR ($max_column IS NOT NULL AND $max_column < 0)
+                OR ($max_column IS NOT NULL AND $min_column > $max_column)
+            ORDER BY $key_selection",
+        )
+            key_values = join(["$key=$(row[Symbol(key)])" for key in key_columns], ", ")
+            min_value = row[Symbol(min_column)]
+            max_value = row[Symbol(max_column)]
+            failure_reasons = String[]
+
+            if min_value < 0
```
gnawin:
```suggestion
            if !ismissing(min_value) && min_value < 0
```

I think it will otherwise crash because min_available_units can be null

src/data-validation.jl
gnawin: Looking at these two, they look cumbersome, maybe a better way is to avoid a missing value from appearing in a boolean check. e.g., by using COALESCE to first convert missing to boolean. But I leave this consideration to you.

gnawin (commented):
Thanks for the PR @datejada indeed it's long, but I managed to read it through :). I have a few questions regarding the validation, and suggested a refactor of the new constraint.

utils/scripts/migrate-investment-limits.py
```diff
@@ -0,0 +1,118 @@
+"""Migrate investment-limit columns and asset available-unit bounds."""
+
+from pathlib import Path
+
+import duckdb
+
+
+ROOT = Path(__file__).resolve().parents[2]
```
gnawin, line 19: Does it only migrate files inside Tulipa? e.g., tutorials, test cases. I might be wrong in this, so please double check. I'd expect it to work also with user files.

src/data-validation.jl
datejada: I agree on the need for a general migration script, not just for this change. So, I created issue #1665 since it needs to be tackled in a separate issue rather than in this one.

src/data-validation.jl
datejada: Good point! It is mainly to avoid a minimum value > 0 with a capacity value of 0. It won't fail per sé (the lower bound will go to zero in this case), but it doesn't make sense to have a minimum investment if the capacity is zero. I will update it to explicitly set it to zero, not with the lower-than-or-equal. The capacity can't be negative from the schema, anyway.

src/data-validation.jl
datejada: I like the idea of COALESCE; I will implement it like that. Thanks

src/data-validation.jl
datejada: It is not the same, because here we check the integer feasibility after converting capacity limits into unit limits. In that case, the lower and upper bounds are calculated with CEIL/FLOOR when the variables are created, so we check not the input values themselves but the code implemented, because it can become infeasible if, after CEIL/FLOOR, the upper/lower bounds are contradictory. For example:

- capacity = 400
- min_limit = 401
- max_limit = 799

Then:

- min_limit <= max_limit is true (so the other check passes)
- but integer units require: CEIL(401/400)=2 and FLOOR(799/400) = 1
- so, 2 > 1 and would make the problem infeasible.

So CEIL/FLOOR catches a real infeasibility that initial min<=max cannot catch when variables are integer.

src/data-validation.jl
datejada: These two validations are complementary, not duplicates.  

The generic check in validates limit consistency for all relevant tables: non-negative bounds and min <= max (when max is present). That still allows max to be missing, and it does not enforce strict positivity for other features.
  
This part of the code is for a storage-specific check that enforces an additional rule only for investable storage assets with use_binary_storage_method enabled: investment_max_limit must be present and strictly > 0.
  
So, the first check ensures structurally valid ranges, while the second check enforces a mode-specific requirement needed by the binary storage formulation.

src/variables/investments.jl
datejada: I agree, the schema validation of the capacity having a minimum of zero makes this not necessary like this, but as `== 0`, I will change it.

src/constraints/available-units.jl
datejada: We need it as it is because the minimum constraint is with `>=` and the maximum is with `<=`
The alternative is to split the function in the `for loop` to have them separately, but it would mean that some part of the code would be duplicated, e.g., lhs (available_units), just for make it explicit the symbol in the constraint...not sure, maybe it is a bit in the middle of readability and maintainability...I prefer the current version, but if you still think otherwise, I don't mind changing it. Please let me know what you think. Thanks!

src/data-validation.jl
datejada: With the change in to COALASE this is no longer relevant

utils/scripts/migrate-investment-limits.py
datejada: The current script is intentionally scoped to Tulipa’s in-repo datasets (test inputs, benchmark EU, and tutorial folders), as required for repository migrations. It does not currently discover arbitrary external user folders because it follows the same idea that the other files in that folder. But, to start with some ideas for issue #1665, I have updated it to have a general script for these PR changes only that will help as an example later for that issue.

I updated the PR description to include how to use the migration script, so it helps with the issue #1665

datejada:
@gnawin thanks for the comments. I have replied to all of them. I rebased and fixed the CSV file conflicts with the latest changes in main.

Please let me know if you have more comments or suggestions.

Thanks!

src/constraints/available-units.jl
gnawin: I was missing an argument like, so I agree ;)

gnawin (approved):
Great! Thanks for addressing the feedback.

Issue #1660, opened 2026-07-28
[Tests] Revive test-constraint-capacity-outgoing-compact-efficiencies


gnawin:
This old test starts with @testset on the top level, and therefore was not run in the test suite.

Reviving it needs to comply with the latest changes.

datejada:
Thanks, I would add that we should try to use TulipaBuilder as much as possible to create the tests. Sometimes is not possible 100% since TB doesn't have all the functionalities. But the basic info can be done with TB, and the more complex part can be done with tailored SQL queries for the test case.

gnawin:
Implemented in #1655

PR #1661, opened 2026-07-28
Add storage coefficient

### LINKED ISSUE #1490: Model a storage with CO2 emissions
### Description

See the use case in #1488. Copied below (https://github.com/TulipaEnergy/TulipaEnergyModel.jl/issues/1488#issuecomment-3895647846):

It's a CAES diabatic technology that has a round-trip efficiency of 1.92. It's discharging more electricity than what's charged because it consumes gas and hence emits CO2.

In COMPETES, it's modeled similar to a battery but the marginal cost includes the gas cost.

In Tulipa, the current way of modeling is adding an outgoing flow from the storage to atmosphere for CO2. I checked the code and the .lp file, there is one flow for charging, and two flows for discharging, that lead to some weird results.

Possible directions:
https://github.com/TulipaEnergy/TulipaEnergyModel.jl/issues/1488#issuecomment-3896286043
 

### Validation and testing

_No response_

### Motivation

This use case comes from the COMPETES-Tulipa comparision.

### Target audience

_No response_

### Can you help?

_No response_

gnawin:
This PR adds `storage_coefficient` in the storage balance.

**Value effect**
- New `storage_coefficient` column on `flow_commission`, carried through the flow/variable tables

**Resolution effect**
- New merged tables `merged_storage_balance` -> `t_lowest_storage_balance` 

**Tests**
- Test merged tables
- Update storage case with CAES
- Update all input CSVs 

**Docs**
- New how-to-use section 
- Update formulation

## Related issues

Closes #1490

## Checklist

- [x] Tests are passing


gnawin:
Hi @datejada I started with the idea of trying to reuse `conversion_coefficient` but later realized that was not possible. Hence, I implemented for storage mimicking the implementation of `conversion_coefficient`.

Thanks for checking!

PS. Along the way I did find another issue #1660. It was not straightforward to fix directly, so I put it as a separate one.

benchmark/model-mps-folder/Power-flow.mps
datejada, line 1: Changes in this file are unexpected. Any reason why?

test/test-case-studies.jl
```diff
@@ -53,11 +53,11 @@ end
     connection = DBInterface.connect(DuckDB.DB)
     _read_csv_folder(connection, dir)
     energy_problem = TulipaEnergyModel.run_scenario(connection; show_log = false)
-    @test energy_problem.objective_value ≈ 2582.214722 atol = 1e-5
+    @test energy_problem.objective_value ≈ 4608.730116 atol = 1e-5
```
datejada, line 56: I assume this change is due to the changes in the test case. is that correct?

So, the concepts section uses this test case for the explanation of storage modelling, here: https://tulipaenergy.github.io/TulipaEnergyModel.jl/dev/30-concepts/#storage-modeling

This section was written before we started creating the plots in the documentation, so please update it as well and add code to generate the plots from the code, not from the PNG. Thanks!

src/sql/create-merged-tables.sql
```diff
@@ -99,19 +143,27 @@ from
...
+where
+    asset.type in ('storage')
+union
+from
+    merged_in_flows_storage_balance
 union
 from
-    merged_all_flows
+    merged_out_flows_storage_balance
 ;
```
datejada, line 167: This is the suggestion instead of the merge tables. It is longer code, but we avoid creating intermediate tables that increase memory usage.

```suggestion
select
    distinct ftrrp.to_asset as asset,
    ftrrp.milestone_year,
    ftrrp.rep_period,
    ftrrp.time_block_start,
    ftrrp.time_block_end
from
    flow_time_resolution_rep_period as ftrrp
left join
    flow_commission as fc on
        ftrrp.from_asset = fc.from_asset and
        ftrrp.to_asset = fc.to_asset and
        ftrrp.milestone_year = fc.commission_year
left join
    asset on ftrrp.to_asset = asset.asset
where
    fc.storage_coefficient > 0 and
    asset.type in ('storage')
union
select
    distinct ftrrp.from_asset as asset,
    ftrrp.milestone_year,
    ftrrp.rep_period,
    ftrrp.time_block_start,
    ftrrp.time_block_end
from
    flow_time_resolution_rep_period as ftrrp
left join
    flow_commission as fc on
        ftrrp.from_asset = fc.from_asset and
        ftrrp.to_asset = fc.to_asset and
        ftrrp.milestone_year = fc.commission_year
left join
    asset on ftrrp.from_asset = asset.asset
where
    fc.storage_coefficient > 0 and
    asset.type in ('storage')
;
```

src/sql/create-merged-tables.sql
```diff
@@ -66,6 +88,28 @@ where
...
+    flow_commission as fc on
+        ftrrp.from_asset = fc.from_asset and
+        ftrrp.to_asset = fc.to_asset and
+        ftrrp.milestone_year = fc.commission_year
+left join
+    asset on ftrrp.from_asset = asset.asset
+where
+    fc.storage_coefficient > 0 and
+    asset.type in ('storage')
+;
```
datejada: This one as well...

```suggestion
```

src/sql/create-merged-tables.sql
```diff
@@ -32,6 +32,28 @@ where
...
+        ftrrp.from_asset = fc.from_asset and
+        ftrrp.to_asset = fc.to_asset and
+        ftrrp.milestone_year = fc.commission_year
+left join
+    asset on ftrrp.to_asset = asset.asset
+where
+    fc.storage_coefficient > 0 and
+    asset.type in ('storage')
+;
+
```
datejada: I would avoid creating intermediate tables, this table and the next one can be used directly in the storage table below, see my comment there:

https://github.com/TulipaEnergy/TulipaEnergyModel.jl/pull/1661/changes#r3673752145

The main thing is that the intermediate cost memory usage is high, and in large cases we are using a lot of memory, so we need to avoid creating unnecessary intermediate tables. We might need to sacrifice some readability, but in this case I think it is okay; it is not that bad. Maybe adding a comment why it is like that...

```suggestion
```

datejada (commented):
Thanks @gnawin here are some comments, please check them and comment in the PR if you have questions. Main thing is to try to avoid intermetiate tables, we are seeing a lot of memory usage in large cases and I want to avoid creating more tables if they are not estricly necesary or reused in another constraints/tables...

benchmark/model-mps-folder/Power-flow.mps
gnawin: I checked the changes before pushing -- only the orders changed so no real impacts, most probably due to some randomness every time the MPS files are generated. But good point, in the latest commits, I removed the changes in this file.

src/sql/create-merged-tables.sql
gnawin: Implemented in the latest commits.

gnawin:
@datejada thanks for the comments, they are all addressed in the latest version:
- mps file change is now limited to only storage.mps,
- avoid intermediate tables by using CTEs for storage balance,
- update docs on storage modeling by using live plots,
- **avoid intermediate tables by using CTEs for conversion balance**.

I'd like to see the effects on benchmarks of the last one, but I cannot because of the schema change. But it should help with the memory usage as you commented, let me know otherwise.

benchmark/model-mps-folder/Power-flow.mps
datejada: great! thanks

datejada:
> @datejada thanks for the comments, they are all addressed in the latest version:
> 
> * mps file change is now limited to only storage.mps,
> * avoid intermediate tables by using CTEs for storage balance,
> * update docs on storage modeling by using live plots,
> * **avoid intermediate tables by using CTEs for conversion balance**.
> 
> I'd like to see the effects on benchmarks of the last one, but I cannot because of the schema change. But it should help with the memory usage as you commented, let me know otherwise.

Thanks for the changes. The benchmarks can always be run locally (if needed), it just that it needs to be done manually, meaning that you run the benchmark in main, save the results, then run the benchmark in the branch, save the results, and then compare (basically what the workflow does automatically in the CI). No need to do it, but just in case you want to do it later for another PR.

datejada (approved):
Thank you! LGTM after the suggested changes 🤓

PR #1662, opened 2026-07-30
Add simple demand response shifting with recovery

IsaiMaganTNO:
Implementation demand response v1, this time on upstream

## Related issues

Closes #

## Checklist

- [ ] Tests are passing


datejada:
Great! Let's continue here. The preview of the docs here is really useful. 
Just to keep track of my previous comments, here is the link to them: https://github.com/TulipaEnergy/TulipaEnergyModel.jl/pull/1652#pullrequestreview-4809349581
Remember the suggestion of splitting this PR into two. Maybe you can keep this one as the main one, create a second PR with only the changes in the files and the schema, and then you will need to rebase this one to only have the changes on top of those to avoid conflicts when the other PR is merged. Sorry, a bit cumbersome, but it is needed if we want to benchmark this feature, which I think we should do.
Thanks!

Issue #1588, opened 2026-04-17
Rename storage variables


datejada:
According to the code review in #1583 consider the following renaming:

- https://github.com/TulipaEnergy/TulipaEnergyModel.jl/pull/1583#discussion_r3099177501
- https://github.com/TulipaEnergy/TulipaEnergyModel.jl/pull/1583#discussion_r3099196543
- https://github.com/TulipaEnergy/TulipaEnergyModel.jl/pull/1583#discussion_r3099237696

gnawin:
I'd suggest rename `is_seasonal` to: `use_inter_period_constraints` or `use_inter_period`. Then implicitly the default is none -> rep_periods.

Or even in the direction of a method, e.g., `time_method`: "representative periods", "periods". This way it is explicit for the users what time horizon the model considers for what asset. 

@datejada ideas?

EDIT: now that I think of the method, it reads as mutually exclusive, but actually not, for an asset it can have both. So maybe method is not a good suggestion...

datejada:
@gnawin Thanks! The one that I like the most so far is: `use_inter_period_constraints`, more verbose, but it is exactly what it does haha

This could be another apparently easy issue; it is only renaming... only....haha

PR #1666, opened 2026-08-04
[docs] Add sections on how to model emissions

gnawin:
Emission modeling seems not straightforward for users @IsaiMaganTNO, and hence this PR adds instructions on how to model emissions which cover
- how to model CCS
- how to model BECCS, i.e., negative emissions
- how to model a emission budget

## Related issues

There is no related issue.

## Checklist

- [ ] Tests are passing

### DIFF
--- a/docs/src/10-tutorials/17-multi-year-investments.md
+++ b/docs/src/10-tutorials/17-multi-year-investments.md
@@ -52,7 +52,7 @@ energy_problem = TEM.run_scenario(connection)
 
 There is a new file *model-parameters.csv*. It contains model-wide parameters, in this case:
 
-!!! tip "Pro tip: Creating the `model_parameters` table manually"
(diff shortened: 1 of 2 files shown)


gnawin:
@datejada I added some more detailed documentation on emission modeling, please have a look and see if you have anything to add/suggest. Thanks!

datejada (approved):
Wow! Way better and more comprehensive. Thanks! I like it, more documentation is always good, plus showing different options to go depending on what you want to model. Hopefully this will help more users.

BTW: THe issue #1588 is for changing from `is_seasonal` to `is_inter_period` (or a better name); because it is that uses the inter-period constraints, not necessarily that is for seasonal behavior or storage only. What do you think? please add your comment on the #1588 and maybe it could be a good next step to make the inter-period more naturally useful in more applications apart from the storage, like this CO2 emmissions (or yearly productions/consumptions...)

PR #1670, opened 2026-08-07
Change initial storage level from energy units (e.g., MWh) to p.u.

### LINKED ISSUE #1104: Validate `initial_storage_level` levels
Check that if there is a defined `initial_storage_level`, it must be between the asset's max/min limits. Also, we need to check that if there are max/min profiles, then you must also check that the initial value is between those values.

datejada:
## Summary

Change `asset_milestone.initial_storage_level` from MWh to an inclusive `[0, 1]` p.u. value. Convert it to MWh inside the model by multiplying it by the existing available storage-energy-capacity expression, including investments and decommissions. Preserve `NULL` cycling behavior and all existing case-study/tutorial objective values.

## Implementation Changes

- Update the input schema and scientific formulation to document p.u. units, `[0,1]` bounds, and the multiplication by available energy capacity.
- In both seasonal and non-seasonal storage balances, replace the raw initial value with `initial_storage_level × available_energy_capacity`.
- Remove the constant final-variable lower-bound logic. Add named JuMP constraints `cycling_condition_intra_rep_period` and `cycling_condition_inter_period` so the final level remains at least the capacity-scaled initial level; omit the redundant constraint when the value is zero.
- Keep `NULL` as the existing cyclic initial condition. Treat `0` and `1` as valid inclusive boundaries.
- Keep rolling-horizon parameters in p.u.; after each window, normalize the solved MWh level by solved available capacity before updating the parameter. Use zero when both capacity and level are zero, and make missing rolling-horizon initial values fail early with an actionable message instead of a type assertion.
- Add validation rejecting out-of-range values according to the schema (general implementation, but with a test using the initial storage level)

## Migration and Repository Data

- Add `utils/scripts/migrate-initial-storage-level.py`, following the existing Python, DuckDB, and `--input-folder`/`--include-default-datasets` conventions without new dependencies.
- For standard datasets, divide legacy nonzero MWh values by `capacity_storage_energy × initial_storage_units`; preserve `NULL` and zero.
- Fail with guidance rather than clamp when capacity is missing/nonpositive, the resulting ratio is outside `[0,1]`, or investment/decommission makes exact result-preserving conversion impossible.
- Run the script over `test/inputs/`, `benchmark/EU/`, every tutorial dataset, and the OBZ documentation source data. The script is a documented one-time migration because an already-converted p.u. value cannot be distinguished reliably from a small legacy MWh value.
- Leave objective constants in `test-case-studies.jl` and `test-tutorial-files.jl` unchanged.
- Regenerate the MPS regression files because final-level bounds become structural constraints.

## Related issues

Closes #1104

## Checklist

- [x] Tests are passing


datejada:
@gnawin here is the PR we chatted about. I tried to keep the changes to a minimum; still, it was a lot. I made an effort to make the commits atomic so you can read them one by one and understand the changes and why more easily. Please let me know your comments, and thanks!

docs/src/20-user-guide/30-configuring-model-features.md
```diff
@@ -61,6 +61,34 @@ In addition, the parameter `capacity_storage_energy` defines the energy per unit
...
+
+When capacity can be invested in or decommissioned, `initial_storage_level` represents a fill fraction, not a fixed quantity of MWh. To impose a fixed initial quantity, fix the available capacity and convert the quantity to p.u.
+
+If `initial_storage_level` is missing, the model uses a cyclic boundary condition: the last storage level supplies the initial state. If it is defined, including as `0`, the first balance starts from the specified fraction and the final level must be at least that fraction of available capacity. Storage-level profiles apply after each dispatch block, so charging and discharging in the first block can move the level away from its initial value.
+
+In a rolling-horizon run, the input value initializes the first window. Each later window uses the previous window's solved storage level, divided by its solved available capacity, as its new p.u. value. Non-seasonal storage therefore requires an explicit initial level when rolling horizon is enabled.
```
datejada:
```suggestion
In a rolling-horizon run, the input value initializes the first window. Each later window uses the previous window's solved storage level in p.u. as the initial value for the next window. Storage assets therefore require an explicit initial level when the rolling horizon is enabled.
```

src/constraints/storage.jl
```diff
@@ -16,6 +16,8 @@ function add_storage_constraints!(
     var_storage_level_inter_period = variables[:storage_level_inter_period]
     var_accumulated_storage_level_intra_rep_period =
         variables[:accumulated_storage_level_intra_rep_period]
+    available_energy_capacity =
```
gnawin, line 19: I'm making notes for myself. Let me know if I get it wrong.

Rename because we only have only 1 vintage method.

src/constraints/storage.jl
```diff
@@ -56,11 +56,13 @@ function add_storage_constraints!(
                     storage_discharging_efficiency = row.storage_discharging_efficiency::Float64
 
                     if row.time_block_start == 1 && !ismissing(initial_storage_level)
-                        # Initial storage is a Float64
+                        initial_storage_level_in_energy_units =
```
gnawin: Maybe rename to `initial_storage_level_in_energy`? `*_units` is a bit misleading, I think.

src/input-schemas.json
```diff
@@ -319,12 +319,13 @@
     },
     "initial_storage_level": {
```
gnawin, line 320: Would it be better to rename to `initial_storage_level_per_unit`? Or if that's aligned with INES?

gnawin (approved):
Looks good! Only a few minor questions on the names, but otherwise this PR is approved. Feel free to merge it, if you will make some changes or not, either is fine.

docs/src/20-user-guide/30-configuring-model-features.md
gnawin: Rolling horizon passes the p.u. per window, which means the absolute level can change?

src/constraints/storage.jl
datejada: Correct, if it becomes available in the future for other vintage methods, then we can rename it. For now, it is just one, so it is straightforward like this. Thanks!

src/constraints/storage.jl
datejada: Great! I agree, I will rename it. Thanks!

src/input-schemas.json
datejada: We are not using the units in the names; the descriptions indicate whether it is per unit or a physical unit (INES does it like that). If we go this way, we should be consistent and name other parameters with that suffix, "_per_unit"; for example, "minimum_operating_point" -> "minimum_operating_point_per_unit", since it could also have been defined in MW... but we decided to use it per unit.
I would keep it like this if you are okay with it, but let me know if you have a strong opinion on this.

datejada:
@gnawin thanks for the review, I pushed the changes suggested here https://github.com/TulipaEnergy/TulipaEnergyModel.jl/pull/1670#discussion_r3757947317, and will merge after the tests pass.

PR #1672, opened 2026-08-12
Change objective discounting

gnawin:
**Summary**
Updates the economic representation to align the annualized investment cost with the standard capital recovery factor (CRF) convention used by OPERA and other capacity-expansion models. 

Currently we discount the first payment, so a single-year run would charge _annualized cost / (1 + discount rate)_, instead of _annualized cost_. After this change, the CRF convention makes the single-year result consistent with OPERA and comparable models, and the same formula applies uniformly to multi-year horizons - no single-year special case is needed as in #1495.

**Changes**

- Remove (1+ discount rate) in the denominator from annualized cost
- Add (1+ discount rate) to the investment discounting factor
- Apply both to assets, transport flows, and vintage flows
- Add `test-objective-discounting-factors.jl` 
- Update affected objective values in `test-case-studies.jl` and `test-tutorial-files.jl`
- Regenerate affected MPS files 

## Related issues

Refactor #1495 

## Checklist

- [x] Tests are passing

### DIFF
--- a/src/objectives/create.jl
+++ b/src/objectives/create.jl
@@ -110,8 +110,7 @@ function _investment_discount_sql(;
                 WHEN $discount_rate = 0
                     THEN $cost / $economic_lifetime
                 ELSE $discount_rate / (
-                    (1 + $discount_rate) *
-                    (1 - 1 / ((1 + $discount_rate) ** $economic_lifetime))
+                    1 - 1 / ((1 + $discount_rate) ** $economic_lifetime)
                     ) * $cost
             END AS $annualized,
             CASE
@@ -131,7 +130,7 @@ function _investment_discount_sql(;
                 -- the weight does not accept a zero cost in the denominator
                 WHEN $cost = 0
                     THEN 0.0 -- zero investment cost, so the weight does not matter
-                ELSE investment_year_discount * (1 - $salvage / $cost)
+                ELSE investment_year_discount * (1 + $discount_rate - $salvage / $cost)
             END AS $weight"""
 end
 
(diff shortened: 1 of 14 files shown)


gnawin:
@datejada this is an inconsistency we found with OPERA, currently in #1495. Before continuing, I only changed the formulas for you to have a look. All the objective values of all the current running cases (all single year I believe) would be influenced.

gnawin:
@datejada this PR is ready for review, but view it as **the next PR after #1673 (so merge that first please)**. Two main comments you had in #1673: duplication of code, type stability are also here, but will be gone after #1673. 

Let me know if you have any suggestions. Thx.

src/objectives/create.jl
```diff
@@ -131,7 +130,7 @@ function _investment_discount_sql(;
                 -- the weight does not accept a zero cost in the denominator
                 WHEN $cost = 0
                     THEN 0.0 -- zero investment cost, so the weight does not matter
-                ELSE investment_year_discount * (1 - $salvage / $cost)
+                ELSE investment_year_discount * (1 + $discount_rate - $salvage / $cost)
```
gnawin, line 133: The only real change is here. Since now this logic is written only ONCE, it reads easier and helps with the maintenance burden. I like your suggestions in the previous PR, now the value starts to appear. Thx!

gnawin:
@datejada I rebased and now the changes seem small (see my comment). The rest is docs and tests. Thx!

datejada (approved):
LGTM

PR #1673, opened 2026-08-12
Update discounting factor for storage energy

gnawin:
**Summary**
Fixes the discounting of storage-energy investment costs. Previously the storage-energy investment term in the objective reused the asset (power) investment discounting factor. As a result, a storage asset with zero power investment cost but a positive energy investment cost had its energy investment discounted by zero. This is a use case in OPERA.

This PR introduces a dedicated discounting factor for the storage-energy component, built from the energy investment cost, so the two components are discounted independently and correctly.

**Changes**
- New discounting factor for storage energy and related terms
- New test in `test-objective-breakdown.jl`

## Related issues

There is no related issue.

## Checklist

- [x] Tests are passing

### DIFF
--- a/src/objectives/create.jl
+++ b/src/objectives/create.jl
@@ -11,6 +11,31 @@ function _add_to_objective!(connection, objective_expr, name::String, expr)
 end
 
 """
+    _query_costs(connection, query) -> Vector{Float64}
+
+Run `query`, which must expose a `cost` column, and return that column as a
+`Vector{Float64}`.
+"""
+function _query_costs(connection, query::String)
+    return Float64[row.cost for row in DuckDB.query(connection, query)]
+end
+
+"""
+    _cost_weighted_sum(costs, terms) -> JuMP.AffExpr
+
+Return the affine expression `sum(cost * term)` over paired `costs` and `terms`,
+where `terms` is any iterable of JuMP variables or affine expressions. Returns a
+zero `AffExpr` when the inputs are empty.
+"""
+function _cost_weighted_sum(costs::Vector{Float64}, terms)
+    result = JuMP.AffExpr(0.0)
+    for (cost, term) in zip(costs, terms)
+        JuMP.add_to_expression!(result, cost, term)
+    end
+    return result
+end
+
+"""
     add_objective!(connection, model, variables, expressions, model_parameters)
 
 Build all objective components, register them in `obj_breakdown`, and set the
--- a/src/objectives/create.jl
+++ b/src/objectives/create.jl
@@ -86,6 +86,46 @@ function add_objective!(connection, model, variables, expressions, model_paramet
 end
 
 """
+    _asset_investment_discount_sql(cost, annualized, salvage, weight, end_of_horizon)
+
+Return the `SELECT` columns that compute the annualized cost, salvage value, and
+investment discount weight for the cost column `cost`, aliased as `annualized`,
+`salvage`, and `weight`. Shared by the power and storage-energy investment terms
+so both use a single formula. Requires `investment_year_discount` to be defined
+earlier in the same `SELECT`.
+"""
+function _asset_investment_discount_sql(cost, annualized, salvage, weight, end_of_horizon)
+    return """CASE
+                -- the closed-form annuity does not accept asset.discount_rate = 0 in the denominator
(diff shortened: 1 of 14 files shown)


gnawin:
@datejada another one found from the OPERA use case. If you could have a look first, that would be great. Thx!

src/objectives/storage_assets_energy_investment_cost.jl
datejada, line 27: There is an issue with this pattern (which is general in our code, not only in this part, but we should start fixing it step by step...maybe a new issue?). Here is the problem: 

The code is iterating over a DuckDB query result and then doing row.cost on a generic row object. The row type is not concretely known to Julia, so the field access is effectively dynamic, and the final JuMP accumulation becomes Any-typed.

This translates to slower JuMP expression construction and more dynamic dispatch than necessary.

This happens because:

1. DuckDB.query(...) yields a DuckDB.QueryResult
2. iterating that result gives generic table row objects
3. field access like row.cost is not statically typed as Float64
4. the sum(...) then accumulates into a generic expression container

So, here is one proposed solution that can also be consider as a pattern for other parts of the code:


```suggestion
assets_investment_energy = variables[:assets_investment_energy]

indices = DuckDB.query(
connection,
"SELECT
           var.id,
            obj.weight_for_asset_investment_energy_discount
               * obj.investment_cost_storage_energy
               * obj.capacity_storage_energy
               AS cost,
       FROM var_assets_investment_energy AS var
       LEFT JOIN t_objective_assets as obj
           ON var.asset = obj.asset
           AND var.milestone_year = obj.milestone_year
       ORDER BY var.id
       ",
)

rows = [(id = Int64(row.id), cost = Float64(row.cost)) for row in indices]

@expression(
    model,
    storage_assets_energy_investment_cost,
    sum(
        row.cost * assets_investment_energy for
        (row, assets_investment_energy) in zip(rows, assets_investment_energy.container)
    )
)
```

This is better because:
1. cost is known to be Float64
2. the iterator is over concrete tuples or vectors, not generic Tables.ColumnsRow
3. the JuMP accumulation path is more type-stable
4. it reduces compile-time overhead and dynamic dispatch

src/objectives/create.jl
datejada, line 1: The power and energy branches duplicate almost the same annualize/salvage logic in the same large CTE. That is not bad by itself, but it raises maintenance risk.

I’d recommend considering factoring the repeated annualized/salvage formulas into a shared common CTE or helper view, then reusing them for both investment_cost and investment_cost_storage_energy. This reduces the chance of divergence between the two formulas.

As I said, this is for maintainability improvement, I think the SQL query is correct. So, if it is too much of a hassle to make it like that, I am open to discuss it.

test/test-objective-breakdown.jl
```diff
@@ -17,3 +17,38 @@
...
+        )
+    )
+
+    # Power investment cost is 0, so its discounting weight is guarded to 0.
+    @test rows[2030].weight_for_asset_investment_discount == 0.0
+    @test rows[2050].weight_for_asset_investment_discount == 0.0
+
+    # Energy investment cost is positive, so its dedicated weight is not zeroed out.
+    @test rows[2030].weight_for_asset_investment_energy_discount > 0.0
+    @test rows[2050].weight_for_asset_investment_energy_discount > 0.0
```
datejada, line 53: Wouldn't it be better to test the actual expected number? not only that is positive? this will make the test more robust.

datejada (commented):
@gnawin thanks for the PR, I left some comments, please let me know what you think of them, and if you have any questions, please don't hesitate to comment on them.

src/objectives/storage_assets_energy_investment_cost.jl
gnawin: Point taken. I created two helper functions, one for type stability and another one for the repeated pattern sum (cost*var).

src/objectives/create.jl
gnawin: Good point. I refactored the whole function to avoid repetitions.

gnawin:
@datejada thanks for the review. I addressed your comments, and took the opportunity to refactor a bit to avoid as much as possible repetition, both in the normal code and in sql. This reduces maintenance burden, so better do it when we see something.

datejada (approved):

PR #1680, opened 2026-08-21
[Bug] Fix Dates.jl version on LTS

gnawin:
There is a bug for the compatibility of Dates.jl and Julia LTS version, see 
https://github.com/TulipaEnergy/TulipaEnergyModel.jl/actions/runs/32507175407/job/96849844401

This change should fix it.

## Checklist

- [x] Tests are passing


datejada (approved):
We don't test on lts in the PRs, but it should be fine, it is a conscious decision, we just need quick fixes like this afterwards.

Thanks!

PR #1683, opened 2026-08-24
Add all test cases to benchmarks

### LINKED ISSUE #1638: Add test cases to the benchmark
Thanks @datejada, these look great, especially the profile changes. I've made non-blocking comments, so it's approved.
One thing, though, these extra benchmarks look really suspicious. Maybe they include precompilation time? Otherwise, it's over 10 times faster?!
If they are indeed correct, them an "easy" win would be to add Tiny, Storage, and Norse to benchmarks, and let the our normal benchmark script capture that. Should be more or less straightforward for the agent to create a new "level", like `higher_level/Tiny/create_model`. Either way, not a requirement for this PR because the improvements are clear

_Originally posted by @abelsiqueira in https://github.com/TulipaEnergy/TulipaEnergyModel.jl/pull/1635#pullrequestreview-4593944499_

**Action in this issue**: Add all the test case studies to the current benckmark

gnawin:
## Related issues

Closes #1638 

## Checklist

- [x] Tests are passing

### DIFF
--- a/benchmark/benchmarks.jl
+++ b/benchmark/benchmarks.jl
@@ -5,15 +5,9 @@ using DuckDB
 
 const SUITE = BenchmarkGroup()
 
-function higher_level_input_setup()
-    # The following lines are checking whether this is being called from the `test`.
-    # If it is not, then we use a larger dataset
-    input_folder = if isdefined(Main, :Test)
-        joinpath(@__DIR__, "../test/inputs/Norse")
-    else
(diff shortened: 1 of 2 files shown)


datejada (approved):
Great! this is helpful to detect changes in more features, even if there are small cases, I think they should show if something is really wrong. The next level will be scale up x times a case study a check the times. But, I don't know how to do that, yet. Maybe for a next project with eScience 🤓

Issue #1688, opened 2026-08-25
[Bug] Tests are failing at Julia lts


datejada:
### Description

The objective function of the tutorials is changing between lts and 1.x. The reason is that the new version of TulipaClustering 0.6.0 can lead to different clusters depending on the julia version compared to version 0.5.0. See images below.

<img width="398" height="93" alt="Image" src="https://github.com/user-attachments/assets/9ba1138b-3b37-42e4-9feb-693d9f75c316" />

<img width="442" height="440" alt="Image" src="https://github.com/user-attachments/assets/79468e08-6929-4e82-b56a-a5ba317d06cd" />

possible solution is to add parameters to the cluster function such that we obtain the same cluster for both julia versions, for instance the  weight_fitting_kwargs= Dict(:niters => 2000, :learning_rate => 0.001) could help to iterate more and obtain the same or closer clusters such that the objective function doesn't change between Julia versions (and possible update to a value the expected objective function with these new parameters)

### Package Version

0.23.0

### Julia Version

lts

### Reproduction steps

run the tutorials 4, 5, and 9 in latest julia version and lts using TulipaClustering v 0.6.0 and compared the objective functions. They are different. See image in the relevant log.

### Relevant log output

```shell
check log here: https://github.com/TulipaEnergy/TulipaEnergyModel.jl/actions/runs/32825219549/job/97731598173
```

datejada:
@gnawin sorry, this one slipped through because we don't test on LTS the PRs 😔  It should be easy to fix, but it is not a major problem, since TulipaEnergyModel runs in LTS. It is just that the objective functions are different, and it is not a straightforward update of the value as I tried to describe here. Not urgent, but to fix it at some point in the near future. I might even do it next week. I will keep you posted here if that is the case. 

EDIT: BTW, if you think it is beneficial, maybe we should also add LTS testing in Ubuntu to the PRs. It will take longer to test, since LTS is slower than 1.x, but it will avoid this type of bug. If so, we can create another issue to update the CI.

gnawin:
No worries :) does this mean we have to to add parameters to the cluster function **for every TC use** to make sure the clusters are correct?

This version can still be used without TC. 

Since this is the second time recently with problems on LTS, I'd say let's add LTS testing in Ubuntu #1689.

gnawin:
Closed by #1692.

PR #1692, opened 2026-08-28
Add lts Ubuntu on PR + fix bug on main

### LINKED ISSUE #1689: Add LTS testing in Ubuntu to the PRs
> EDIT: BTW, if you think it is beneficial, maybe we should also add LTS testing in Ubuntu to the PRs. It will take longer to test, since LTS is slower than 1.x, but it will avoid this type of bug. If so, we can create another issue to update the CI. 

 _Originally posted by @datejada in [#1688](https://github.com/TulipaEnergy/TulipaEnergyModel.jl/issues/1688#issuecomment-5408898617)_

gnawin:
- Add LTS Ubuntu testing on PR. It is much slower but it prevents cases where LTS might break.
- Change niters to 2000 for TC, and then the objective values for tutorial 4, 5, and 9 do not differ anymore between LTS and 1.
- Align all workspace environment TC versions to v0.6 in test, docs, and benchmark.
- Update invalid references in documentation and code.

## Related issues

Closes #1689, #1688

## Checklist

- [x] Tests are passing


gnawin:
@IsaiMaganTNO @datejada this PR fixes the bug on main. Maybe we should also release a minor version after this fix.

datejada (approved):
@gnawin thanks! and I agree, please release a minor version after this one, and maybe the pre-commit update

PR #1700, opened 2026-09-09
Wrap orcid in quotes hopefully fixing Zenodo integration

abelsiqueira:
@gnawin, @datejada, the Zenodo integration did not trigger with release v0.23.2 (ref: https://zenodo.org/account/settings/github/repository/TulipaEnergy/TulipaEnergyModel.jl).
I'm not sure what happened, but I have a guess, which is this small change in here. It's not documented anywhere, but the other ORCIDs follow this.
This is a valid CITATION.CFF, which is why the pre-commit CFF validation passed, but I guess Zenodo's internal conversion does its own thing.

After this is merged, I'll manually create a `v0.23.2-zenodo` release just to trigger it and avoid creating a whole new Julia release unnecessarily, if you're ok with it.

## Related issues

Closes #

## Checklist

- [ ] Tests are passing

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -42,7 +42,7 @@ authors:
       family-names: Elgersma
       email: m.b.elgersma@tudelft.nl
       affiliation: TU Delft
-      orcid: https://orcid.org/0009-0007-1417-8757
+      orcid: "https://orcid.org/0009-0007-1417-8757"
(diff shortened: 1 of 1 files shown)


datejada (approved):
Good catch! Thanks :)

PR #1702, opened 2026-09-12
Fix missing quotes in affiliation in paper.md

abelsiqueira:
@datejada @gnawin, this is a small fix of missing quotes in the affiliation in the paper.md.
It only triggers an error on the JOSS generate PDF pipeline, so it was uncaught. Hopefully this will be the last of these issues :)

PS. The last PR indeed fixed the Zenodo issue. After this is merged I'll do the same and create another manual release.

## Related issues

Closes #

## Checklist

- [x] Tests are passing

### DIFF
--- a/paper/paper.md
+++ b/paper/paper.md
@@ -18,7 +18,7 @@ authors:
     - given-names: Diego A.
       surname: Tejada-Arango
       email: dtejada@epri.com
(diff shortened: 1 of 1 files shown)


datejada (approved):
LGTM

